PR #5, opened 2025-09-08
Infraredbaron's Court Outfits Port

Jamdrawers:
## About The Pull Request

Brings back the cool keep outfits from ratwood; the Hand's swaggy coat, the Duke's swish epaulets, the counsellor's comfy duds. 

Also changes the Hand, Counsellor, Lord, Steward and Magos to start with it in their armour slot. In future mapping efforts I'd also want to put these in the appropriate role's wardrobe.

Oh, also, since these are -licensed- sprites I'm putting this up in a draft so the appropriate person can do the appropriate code wizardry to make it secret and sneaky.

image

Since we don't have Watchmen, the Watchman armour isn't going to be in any role's starting equipment, though I'd be tempted to tuck it away in the watch garrison's armoury somewhere. Otherwise, we might want these outfits in the appropriate role's spawn equipment, or alternatively we can just leave it in the role's wardrobe. Currently I have the more popular outfits in the starting armour slot, but either way I'd want the alternative to be in a wardrobe and available for sake of variety.

## Why It's Good For The Game

These were cool and good.


Pidgey-chan:
These (court outfits and the watch armor both) didn't fit the game even when there was about a month of effort towards completely redoing the aesthetic unasked for. Let's not make this mistake again.

Jamdrawers:
To do list:
1) Fix sleeves
2) Fix sound effects on the town watch armour (maybe just move them to another file to inherit traits from something)
then
3) Apply to select roles as desired
4) place into the map in wardrobes

Alanii:
I would change this PR to just adding them to the code and existing. Perhaps even craftable. These items have a place, and I have some ideas where they may be.

Jamdrawers:
> why would we do this again

Because these outfits are so good and iconic!

>only for it no go where

It went into a cool and unique and memorable hightown, and some nice outfits, and I assume the cool early flintlocks as well?
Nice to not have all of the roguetown servers be exactly identical to eachother.

PR #8, opened 2025-09-09
ROCKHILL!!

Jamdrawers:
## About The Pull Request

ports ROCKHILL!

Reintroduces the map with innumberable tweaks for the new codebase.
Ports back objects like street lights and desk bells as well.

## Testing Evidence

Rockhiiiill!
image
image
image
image
image

Bandit camp up in the mountains!
image
image

Improved Vampire Manor
image

New Inquisitorial Manor!
image
image
83c8c484eeed7fc5d521d5d435dd17f8

## Why It's Good For The Game

I love Rockhill. More map variety is good, and it makes the server meaningfully not-identical to all the other servers.

Notable Changes:

1) Hightown Barracks and Manor Barracks share the same keys, so MAA can use the buildings interchangeably
2) Mayor's office can be opened with a Sherriff's key, by the handful of leadership roles that have a sheriff's key (until we decide what to do with all that)
3) ~the room-key seller in the tavern might have a few more keys to sell than rooms. Don't buy a key for room 7~ fixed!
4) Counsellors no longer get a unique counsellor key but I added some new ones that you can claim by taking the key on the table in the room
5) There's an electro-frankenstein revival chair in both the town physician and the court physician's areas - both of which have been expanded to accommodate.
7) There's a Jester's room on the 3rd floor of the manor
8) New Seneschal room and re-re-re-redesigned servant's quarters
9) Nightmaster now has their swagface. Remember to find it behind the reeds in their bedroom
10) New inquisitorial manor in the southwest of town!
11) Probably more things I forgot to mention

ISSUES:

1) ~Lighting. Lights in AP are often not as powerful as in ratwood so rooms that would have been fully plainly illuminated by, say, blue candles are instead baarely not pitch-black.~
Pretty much every room in rockhill has been manually re-lit!

2) ~No metal lights - these are pretty iconic for Hightown and a lot of places are dark without them. I've replaced them with lanterns and torches and candles but it'd be nice to see them reintegrated~
I've ported the iconic hightown lights and reintegrated them - though I haven't replaced ALL of them - outside of the main highstreet, merchant and keep they should mostly be torches so that thieves can snuff them.

3) ~Area Generation: Need to fix Bog Generation to generate foliage at a rate similar to OG rockhill, rather than the AP bog generation which is set to utter maximum turbo-mode~
I have fixed area generation! I made an alternate area and generator called /bograt which generates nicely now.

4) Wretch travel tiles: Currently I have them set to the old Goblin travel tiles - For sake of balance we proooobably want to move them into the bog like bandits, but for now it'll be fun to try out.

5) ~Since the court physician in rockhill lives in the keep, they should have a different set of keys than regular physicians~
Added a Court Physician's key in the Court Physician's keyring. Opens doors in the court physician's upstairs quarters. Will add a key to the court physician's private quarters in dunworld when aren't  big mapping PRs being TMed there. Until then it shouldn't be a problem.

Most of the issues there should be fixable in the coming weeks, some are more of a bigger design decision or project to do in the mid-to-long term

1) Due to popular support, somewhere down the line we ought to be bringing back the Bog Gate. Not sure exactly what form this will take but it'll likely result in a bit of an overall lowtown restructuring (just when I'd got it just the way I like it! Ah well) since the position of the vanguard fort is so central to the town layout.
2) ~~Probably ought to kick the inquisition out of the church's basement with how much more unique stuff they have going on as a faction now, more of their own thing. I sorta partly did this with the secret hideout safehouse I added but that was designed for one inquisitor rather than the full team faction we get now.~~ done!
3) Thinking I might merge the Lord's Office and the counsellor's Meeting Chamber (both big rooms with a big table and chairs and paper), and replace the area with more lavish Guest rooms.
4) A big bog/wilderness overhaul with dungeons full of traps and monsters and loot and more varied terrain and landmarks
5) ~A dedicated Martyr's room~ done
6) Probably more!


Jamdrawers:
image
image
image
This also has Centcomm changes, partly to make it compatible with rockhill (just a tiny bit), partly to make it distinct, and partly to just make it better.

I removed the totally superfluous kitchen from the lair of people-who-never-get-hungry, and replaced it with a cozy RP zone.

Jamdrawers:
a3e6d41257ad7013939441af89e18fbf
0c1934e3e8d07a5eca49771678abe4c3
83c8c484eeed7fc5d521d5d435dd17f8
Inquisitorial Manor! I realised they don't even have church keys so out of the basement they go. They used all that otavan money to buy an old manor in the edge of town

RadiantFlash7:
I've ran a glance through some of the updated buildings in town. Everything is looking good; though my main comment is that we may want to consider reducing wall lighting and replace it with silver and gold candle sticks in places where feasible. 
It provides reasons for players to break into buildings with inherent value, while also providing light generally.

CVSthePharmacy:
NO DON'T MERGE IT NOOOOOO

PR #11, opened 2025-09-09
First anti goblin pr

Heheniy:
## About The Pull Request
Kicks out goblins from roles they shouldnt have access to

## Testing Evidence
you will get none

## Why It's Good For The Game
Goblin duke/priest/martyr/inquisitor/etc? Ah hell nah

### DIFF
--- a/code/datums/migrants/migrant_waves/heartfelt roles.dm
+++ b/code/datums/migrants/migrant_waves/heartfelt roles.dm
@@ -2,7 +2,7 @@
 	name = "Lord of Heartfelt"
 	greet_text = "You are the Lord of Heartfelt, ruler of a once-prosperous barony now in ruin. Guided by your Magos, you journey to the vale, seeking aid to restore your domain to its former glory, or perhaps claim a new throne."
 	outfit = /datum/outfit/job/roguetown/heartfelt/lord
-	allowed_races = RACES_NO_CONSTRUCT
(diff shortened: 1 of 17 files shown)


DeltaDav:
Since they didn't bother saying what role in description, here is the following :

Heartfelt roles
Inquisition
Matyr
Bishop
Councilor
court Magician
Veteran
Knight Captain
Hand
Duke
Knight
Steward
Suitor
Prince / Princess

In my opinion, I don't like the PR, its funny to play a little green shit in those position and isn't that bothering. It make sense for Heartfelt, Inquisition, Matyr however.

RadiantFlash7:
I'm hesitant to start racelocking roles at present.

One of the main issues with goblins on RW1 was that goblins were always little shitheads, always had a license to cause trouble, and were always valid, but as it was their behaviour had to be balanced against the other rules regarding being interesting and good faith and such, which, with no clear guidance, meant it was basically entirely subject to whichever admin happened to take a ticket on.
While we could lock them back down to limited job selections; AP race stats are much less influential in determining that jobs they can play. And I'm not really wanting to restrict players from jobs given the state of flux in our server population.

That said, they /are/ a goblin. If you see a goblin noble, you're welcome to react as you would to them existing in character.

In the end, it's something we can come back to later and have a discussion on in the future.

PR #29, opened 2025-09-14
Initial Unarmed Pass

ProbablyCarl:
## About The Pull Request
Modifies some of the most polarizing weapons, by far.

~~All unarmed given 0 WDefense. Bracers serve this purpose.
My concern with that change was brass knuckles, but they've higher integrity and heavy weight, so they should be fine. I guess? It's not like anyone used them as is. We really need to add caestus, for a proper unarmed defensive option that doesn't use bracers.~~

In general, the changes below should make unarmed specific weapons not as absurd while still being viable. We can number tweak them further, especially if the damage is too low, but they were fine in testing. The DPS decrease, alongside the base damage, serves to make them reasonable, and not just swift intent nose spam knockout clubs. In regards to the katar, that is. Knuckles having defence values was odd, given bracers serve for that with unarmed.

Finally, this wraps them all into their own weapon file, so it's not as difficult to navigate, and fixes berserker punch dagger spawns.

Anyhow, here's the raw stuff below.
For katars:
 - Katar cut lost damage increase.
 - Katar thrust and cut now has a cooldown of 8, as opposed to 6. Same as a dagger.
 - Katar's damage dropped from 24 to 18. Punch dagger's was 15 by comparison.
 - Katar integrity raised to 120, from 80. Identical to punch dagger.
 - Barotrauma from 27 to 21.

For knuckles:
 - ~~Knuckles dropped 1 point of damage.~~
 - ~~Close Caress dropped 1 point of damage.~~
 - ~~Given a unique daze intent as tradeoff for the loss of defence.~~ Same functionality as maces. -2INT, -1FOR. 10 seconds. 20% less damage, same strike time as smash. Used for opening a parry window. Should give them more use.

## Testing Evidence
Lotta number changes. Spawned in and it worked for the Berserker fix. Weapons tested fine locally. Not really sure how you'd want me to document that.

## Why It's Good For The Game
Katars are, by far, the best weapon in the game. Hands down. No competition. You can take on an entire crowd and win without contest just by aiming nose. You don't even need swift intent. Templar Monks are an incredible example of a meta class that makes use of this. Berserker Wretches are even worse, spawning with a katar as is. There's no variety for unarmed. Knuckles are pointless by comparison.
Not to mention that katars had their own unique cooldown of SIX(6!!!!), compared to contemporaries (daggers) having EIGHT(8!!!!). So with them unified in cooldown for thrust, it shouldn't be as nutty.

~~The only weapons unarmed wise to retain their defence are Psydonite unarmed weapons, and that's a basic +1. Nothing great, but it still makes them stand out by being the only weapons in the lineup with defensive value. For however little.~~

### DIFF
--- a/code/datums/status_effects/rogue/debuff.dm
+++ b/code/datums/status_effects/rogue/debuff.dm
@@ -391,6 +391,10 @@
 	effectedstats = list(STATKEY_PER = -3, STATKEY_LCK = -1)
 	duration = 8 SECONDS
 
+/datum/status_effect/debuff/dazed/unarmed
+	effectedstats = list(STATKEY_INT = -2, STATKEY_LCK = -1)
+	duration = 10 SECONDS
+
 /atom/movable/screen/alert/status_effect/debuff/dazed
 	name = "Dazed"
 	desc = "You've been smacked on the head very hard. Which way is left, again?"
--- a/code/game/objects/items/rogueweapons/melee/special.dm
+++ b/code/game/objects/items/rogueweapons/melee/special.dm
@@ -1,29 +1,4 @@
 /// INTENT DATUMS	v
-/datum/intent/katar/cut
-	name = "cut"
-	icon_state = "incut"
-	attack_verb = list("cuts", "slashes")
-	animname = "cut"
-	blade_class = BCLASS_CUT
-	hitsound = list('sound/combat/hits/bladed/smallslash (1).ogg', 'sound/combat/hits/bladed/smallslash (2).ogg', 'sound/combat/hits/bladed/smallslash (3).ogg')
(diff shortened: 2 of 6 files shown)


MarioTheShiba:
this is good. i like this.

CVSthePharmacy:
Don't fuckin' nerf knuckles they're already ass as is, in fact they kinda need a BUFF. Keep the defense and give them an intent modifier, it'll make it easier to use in general cause right now it has NO place in combat. It USED to be a 185% integrity damage weapon, making it better for taking down armor, but that got removed, keeping it as a overall shitty weapon. Just don't nerf it even harder man, it's like kicking a dead horse, for fucks sake.

PR #38, opened 2025-09-15
Fixes Varangian Class (Statblock, Miracle Tier, Traits)

snowysquirrel:
## About The Pull Request

Varangian got ported strangely and had some inconsistencies between comments and reality. Placed them down to tier 1 miracles. Ported traits & stats from Azure equivalent.

## Testing Evidence

**Traits**
image

**Statblock ((as wildkin) (spawned in thirsty, hence reduced WILL))**
image

## Why It's Good For The Game

Big norse women + class functional again

### DIFF
--- a/code/modules/jobs/job_types/roguetown/mercenaries/classes/atgervi.dm
+++ b/code/modules/jobs/job_types/roguetown/mercenaries/classes/atgervi.dm
@@ -50,7 +50,7 @@
 	beltl = /obj/item/storage/belt/rogue/pouch/coins/poor
 
 	var/datum/devotion/C = new /datum/devotion(H, H.patron)
-	C.grant_miracles(H, cleric_tier = CLERIC_T2, passive_gain = CLERIC_REGEN_WEAK, devotion_limit = CLERIC_REQ_2)	//Capped to T1 miracles.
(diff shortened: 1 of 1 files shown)


CVSthePharmacy:
Can you give them 2 hurlbats or smth, the hurlbat description says they'd have 'em but for some reason they don't

CVSthePharmacy:
Wait eyo why do these guys get perception, that doesn't seem right vro

snowysquirrel:
> Actually yeah hol' up. 9 stat spread, compared to the other's 7? Atop being a miracle caster.

what statspread is recommended? i pulled the current one straight from azure
i'm wondering if their miracle casting should be removed outright as well? i don't see why they have access to it when their counterpart shamans already do
main goal was to just get varangian working since they were turbobroke earlier

CVSthePharmacy:
> > Actually yeah hol' up. 9 stat spread, compared to the other's 7? Atop being a miracle caster.
> 
> what statspread is recommended? i pulled the current one straight from azure i'm wondering if their miracle casting should be removed outright as well? i don't see why they have access to it when their counterpart shamans already do main goal was to just get varangian working since they were turbobroke earlier

Azure stats are inflated, if anything mayb' jus' give 'em -1 per instead of +1 and it'd be sick I think, unless mon con an' mon wil are too much for the other classes. Though tbh grudgebearer exists so I think the miracles are a good tradeoff for the -1 per and -1 spd


I may be stupid with this comment but I might as well

snowysquirrel:
adjusted statline again, is this more fitting?

image

PR #41, opened 2025-09-15
More ratwood music

Jamdrawers:
## About The Pull Request

There were a few tracks I missed last time (The bandit startround theme and the roundend credits music from ratwood 1 that I so adored)
I'm also putting forward two additional tracks, this one to replace the bandit theme, which I thought was waaaay too excited and jaunty to be background ambience - https://www.youtube.com/watch?v=jO0iUaWrmL8&ab_channel=DmitryV.Silantyev-Topic
It's suitably ambient to be something droning constantly in the background, and it's got a little bit of a sinister edge to it to fit thematically.

And this one for the underdark - https://www.youtube.com/watch?v=SVx_ZD8eoYE&list=RDSVx_ZD8eoYE&start_radio=1&ab_channel=Khalbrae
I just think it has the perfect underdark-y sense of mystery. This one is less necessary of a change since the original one wasn't bad either.

Lastly, found some very appropriate music for a tavern that I added to the list of tavern music - https://www.youtube.com/watch?v=NEAdgacuwjk&list=PLYMVwIDsNc9FXGVGS1XVK2JmfCdekvlDf&ab_channel=Dryante-Topic

I'd like to find a replacement for the dwarf/smithy district which on the flip side is way *too* ambient but it's weirdly difficult to find.


RadiantFlash7:
I'm personally not the biggest fan of the change for the bandit ambience here.
That comes largely down to personal preference though. I think I'll throw a poll on it in discord and see what everyone else thinks between the two.
Othert then that this is a fairly successful tm

Jamdrawers:
I don't hate the idea of the original bandit music but it just isn't *ambient* enough to constantly playing in the background. I could imagine us finding a better replacement track in the future some time.

PR #42, opened 2025-09-15
Ratwood renames

Jamdrawers:
## About The Pull Request

Renames a bunch of things back to Ratwood versions instead of the AP version that was copied over, tweaks a few other names, and adjusts recipe descriptions.

1) Atms renamed back to Nervelocks. Also adjusted the filenames to be more consistent
2) ~Skills renamed back to Beginner/Competent/Proficient. ~ Woe is me
3) Raneshi cake is now Zybantine cake, as it was before.

4) 'Crude Selfbow' is renamed to 'Simple Shortbow', with the description changed to be less of an insult 

5) Changed a series of recipe names so instead of listing "wooden table" it simply lists "table"

## Testing Evidence

image
image
image
image
image

## Why It's Good For The Game

1) Nervelock is a cool name
2) I think it was a much better naming scheme that described it much better rather than describing your relationship in a guild system. Not many people are Apprenticing under anyone to learn swimming, etc.
3) It's fun to say
4) Feel like the previous description discouraged you from using it too much
5) You already know the table will be wooden the moment you opened the Carpentry table.  Makes the list much easier to look through.

It's also nice to be less of a copy-paste of other servers.

I would be tempted to include the skill requirement for all of the crafting recipes but I suspect there must be a more efficient way of doing that.

### DIFF
--- a/code/controllers/subsystem/skills.dm
+++ b/code/controllers/subsystem/skills.dm
@@ -10,18 +10,18 @@ SUBSYSTEM_DEF(skills)
 	var/list/all_skills = list()
 	///Static assoc list of levels (ints) - strings
 	var/static/list/level_names = list(
-		SKILL_LEVEL_NOVICE = span_info("<span class='small'>Beginner</span>"),
-		SKILL_LEVEL_APPRENTICE = span_info("Competent"),
-		SKILL_LEVEL_JOURNEYMAN = span_biginfo("Proficient"),
+		SKILL_LEVEL_NOVICE = span_info("<span class='small'>Novice</span>"),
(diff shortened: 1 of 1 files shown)


RadiantFlash7:
Taking a look at this, most of this seems relatively fine.
That said- I'm not a fan of the name change to skills, and to my knowledge nor is the other maintainers particularly a fan of it.

I'll say remove the change to skill names and this can be test merged

Jamdrawers:
> Taking a look at this, most of this seems relatively fine. That said- I'm not a fan of the name change to skills, and to my knowledge nor is the other maintainers particularly a fan of it.
> 
> I'll say remove the change to skill names and this can be test merged

Alas, it is done.

PR #45, opened 2025-09-16
Balancing Zad Form

HelennaC:
## About The Pull Request
Removing completely due to same problem from testing-alpha

## Testing Evidence
It just works

## Why It's Good For The Game
- Increased cooldown in case of antag/prisoner using it.
- Removed from Witch due to abuse being repeated over and over
- Witch loadout already got enough goodies to sustain itself, like the +2 speed. So it won't make much difference if they need to run.
- Gib and shout added for visual and audible reference of transformation


Jamdrawers:
This sounds very sad. Would be so much nicer to try and implement some counterplay rather than removing a fun and flavourful and cool feature from the game.
Maybe if zadform was vulnerable, and you could force them out of zad form by hitting them with an arrow or a rock... Or if it had a longer cast time, or something. like that

PR #49, opened 2025-09-17
Buffs for Witch towner subclass

MarioTheShiba:
## About The Pull Request

After Witch - understandably - had their zad form removed pending a rework, currently, in my opinion, witch is in a terrible and deplorable state, outclassed by other mages and pretty much has nothing going for it except for magic missile and a few things. This aims to change it.

Gives witch Message spell.

Gives witch 9 spells points ( was 6 )

## Testing Evidence

It's just number changes.

## Why It's Good For The Game

Witch, understandably, was hard nerfed due to Zad form being arguably incredibly overpowered and unfun to play against. Now, though, it's got very little going on for it. I wouldn't go as far as to say it's potter level of useless, but it definitely feels gutted. This should bring some love back to it.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/pilgrim/witch.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/pilgrim/witch.dm
@@ -5,7 +5,7 @@
 	allowed_races = RACES_ALL_KINDS
 	outfit = /datum/outfit/job/roguetown/adventurer/witch
 	category_tags = list(CTAG_PILGRIM, CTAG_TOWNER)
-	traits_applied = list(TRAIT_DEATHSIGHT, TRAIT_WITCH, TRAIT_ARCYNE_T2)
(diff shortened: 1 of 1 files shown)


Rutobia:
I think it's important to remember that they are still a towner class and shouldn't really be compared to casters. To be honest even magic missile feels like it doesn't belong on them. Their purpose as a role is to make potions and perform rituals, the spells they have access to are just to give them magic flavor without giving them anything powerful, even Zad form was just because it's a common depiction of witches to be able to turn into crows/ravens. If they lost their Zad form I think the best way to offset that would simply be to put haste, longstrider and/or teleportation in their spell list so they can use magic to get away from danger but not to be a danger themselves. If they would get anything offensive it should be like a buff/debuff deal so they can offer people "boons" or "curses" to fit more in line with the witch theme but nothing to deal direct damage.

Elli-Skala:
Witches arent doing advanced summoning to my knowledge,  I think it would be nice neich for them if they had the ritual trait back to be the single nom church/antag to do rituals

PR #54, opened 2025-09-18
Wretch PQ + Respawns

ProbablyCarl:
## About The Pull Request
Simply removes slots opening on Wretch death, and jumps the PQ requirement to a staggering 60. Up from 20. Nearest highest is Lunatic at 100, and Martyr at 10 below both, IIRC.

We don't want new folks playing this. At all. Spawning armour roundstart and rushing into town is cringe and you should stop. Play other roles, such as bandit, to get a feel for antag. You're intended to progress the round and contribute in a meaningful way. You get the tools to do serious damage when you're trusted and gain enough PQ.

It's not as if PQ is hard to get either. This is just a gate so we don't have new friends stumbling into it and being dumb.
## Testing Evidence
It just works.
## Why It's Good For The Game
Read above.
New players should not have access to tools such as heretical major regen casters who get rites and such. They complain too much about dying immediately in the armour after being valid, or don't actually understand the role.
Wretch drives the tempo and happenings of events within round, and is something we want experienced players to be doing.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/wretch.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/wretch.dm
@@ -12,7 +12,7 @@
 	outfit_female = null
 	display_order = JDO_WRETCH
 	show_in_credits = FALSE
-	min_pq = 20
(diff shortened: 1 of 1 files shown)


CVSthePharmacy:
Can you lower it to fifty

PR #59, opened 2025-09-19
More small QoL map fixes

HelennaC:
## About The Pull Request
More QoL for the map

## Testing Evidence
It just works

## Why It's Good For The Game
- Updated the south gate with more details.
- Fixed Apothecary with some small details and fixed area issue
- Named the wolfs of the keep (Reference to Rockhill but they are new pups, so different names)
- Bath-house works rooms are now considered proper rooms and not using the 'claim house' system
- Some small touches to a few kitchens and items that should be on display for more fluff and usage
- Now you can walk over the small rocks on the beach. (What was the issue before, too scared?)
- Goldface spawns locked, nice try kiddo. (I was thinking of doing the same to the never master buuuuut nah.)
- Nerfed the Apothecary building, why every room has a SCOM while they all spawn with rings?


HelennaC:
Closing this until map on main stop causing conflict on big map changes

PR #65, opened 2025-09-20
Falsewall It.1

Lezerchil:
## About The Pull Request

A simple addition of structure which can be opened and closed with a codeword.
A word for opening and a word for closing are actually 2 different things here.

## Testing Evidence

Here's a short video. I'm a bum so it is compressed in clideo.

https://github.com/user-attachments/assets/dcab6518-d02c-49b0-b391-d367ee25832a

## Why It's Good For The Game

1) It was requested.
2) It is a good way to make secret escape or entry routes for some secretive groups.


GravyMaker:
A good PR. I don't see any immediately obvious issues with the code, and since there's no actual use of the walls in this PR that I can see - either in terms of such a wall being placed in the map, or a recipe to allow players to make them - a test merge wouldn't really achieve anything, so I'm going to full merge this so that our mappers can get to work with it sooner rather than later.

PR #70, opened 2025-09-21
Conjure primordial

RadiantFlash7:
## About The Pull Request
This PR started as a PR to fix order minion spell because it was only half functional. However, it rapidly balooned in scope to add conjurable creatures for t3 arcane mages to wield.
It adds 3 types of conjurable primordials, Air, Water, and Fire.
To conjure them consumes an essence of the type held in hand.
Upon conjuring your first primordial, you're given 2 other spells, Order primordial, and primordial mark.
Order primordial functions much like order minion, Clicking the ground orders your primordial towards that spot. Clicking an enemy sets them at the target, Clicking yourself as your primordial follow you, and finally clicking the primordial will toggle their aggressiveness towards all neutral creatures (Mostly players)

Primordial mark does two things. Clicking any creature with it will add it to your faction, thus keeping them from being attacked by your primordials. Clicking a tile with it will activate the primordials special ability if they're within 3 tiles of the target.
The flame primordial will throw out a cone of fire
The water primordial creates a temporary 3x3 vortex of river water pointing towards a deep water center
The Air Primordial outputs a wave of air that travels towards the target, flinging anyone hit by it away.
There is a limit of 2 active primordials per conjurer at one time.
Presently, these spells can not be refunded once picked.

## Testing Evidence
image
image
image
image

## Why It's Good For The game
Conjuration mages are pretty sweet. It seems like a neat feature idea. Probably the last feature I throw to mages for a while though.


CVSthePharmacy:
> > Hi ummmm the fire primordial is preeeeetty op
> 
> Thanks for the feedback but er. Could you be a little more specific? Saying their OP isn't quite helpful so much as why you believe so.

the fire primordial is kinda like a greater fireball that also walks up to people and attacks. In my opinion just remove the cone of fire and replace it with spitfire and keep the melee attacks, otherwise yeah the cone of fire is FUCKED

PR #71, opened 2025-09-22
Ports Miacraft

RadiantFlash7:
## About The Pull Request
Ports Miacraft from https://github.com/GeneralPantsuIsBadAtCoding/Azure-Peak/pull/3790 and https://github.com/GeneralPantsuIsBadAtCoding/Azure-Peak/pull/3845

## Why It's Good For The Game
Better crafting menus and system for crafting.


RadiantFlash7:
Only noted issue is crafting recipe labled the same can cause issues when theres multiple recipes I believe. I'll take a look into fixing it before this is full merged

PR #75, opened 2025-09-24
Inherent Traits, Continued

ProbablyCarl:
## About The Pull Request
Species traits are something we'd done to set them apart. Something unique and flavourful. This continues that trend, or so I'd hope, and provides them something to play into.

So, here we go. The last to receive traits, before eventual reworks or refluffing to better shape them.
Half-Kin, Wild-Kin and Half-Elves are not slated to receive traits. I'd considered giving them a list of unique stuff to choose from to build their character's identity around, but you're already a fence sitter by playing them. RAAAA!!!! In the future they'll get something, I guess. Not now, though.

New Traits:
 - Humen: Humen Ingenuity. Jack of All Trades refluffed. 25% cheaper to increase sleep level, with a new sleep message, as opposed to the 50% of Jack of All trades before. Also caps the cheapest a level can be to 2.

 - Dwarf: Drunken Healing & Dwarvish Knowledge. The former allows you to slowly (very slowly) heal with anything of a boozepower at 25 or over in your system. The latter allows for free repairs in the exact same way that Failed Squire works, though doesn't give you the tools or polishing knowledge, of course.

 - Tiefling: An inherent 15% to not catching fire. At all. They roll each time they're to be ignited, and it's a 5% chance they just don't catch flame. I wanted to make it impossible for them to catch flame below a certain fire stack level, but that wasn't playing nice at all and causing problems. So, they get something much less useful, but still flavourful I hope.

 - Drakian have lost their darksight. It was weird and didn't fit. Instead, they get the 'Wing Bound' trait. What this does is act like OldAcrobat, to a degree. They can fall one Z-Level, but are off balance for twelve seconds(kick to floor), and get stunned for two seconds on top of that. I'd considered just making them take less damage, but this is a bit more soulful IMO. If it's too powerful, we can tweak it.

Virtues 2 Species:
 - ~~Elves get Keen Ears. That's it. I couldn't think of anything flavourful to give them beyond that. Someone else can do them later.~~ Elves get Woodwalker. Drow get Webwalker. The latter isn't a virtue but I'll shove it here anyway.

 - Half-Orcs are a similar situation, but instead they get Big Guy, which is for soft-grab throwing and smashing through doors. They also all get Bed Breaker. Because it's funny.

Other:
 - Homesteader removed entirely. As with the Jack of All Trades feat/trait.
 - Zadform removed from Prisoner. Unrelated to this PR but they really, really shouldn't have it.

## Testing Evidence
image

image

## Why It's Good For The Game
Missed species/races getting a trait pass. They missed out in the initial PRs due to no ideas rattling around in the other contributor's head, or my own for that matter.


anconfuzedrock:
Woodwalker is one of the strongest virtues, lets baddies be basically impossible to pursue in forests and still helps with climbing in town.

PR #76, opened 2025-09-24
Dreamwalker Disabling

ProbablyCarl:
## About The Pull Request
This was dropping in super lowpop rounds and curbstomping folks. It really has no reason to exist with our current lore, either, and could do with a rework. I guess. Staff can still inject it, but the lowpop tide of these was not great.

## Testing Evidence
Carl slop. No testing.

## Why It's Good For The Game
I really don't think I need to explain this. We started refunding / canceling it in round with how common it was.

### DIFF
--- a/code/modules/events/antagonist/solo/dreamwalker.dm
+++ b/code/modules/events/antagonist/solo/dreamwalker.dm
@@ -16,7 +16,7 @@
 	maximum_antags = 2
 
 	weight = 18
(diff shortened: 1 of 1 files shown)


RadiantFlash7:
I will say, dreamwalkers were included in our new lore. The question is more so whether or not we can add a poplock to it so it isn't rolling at lowpop constantly.

PR #84, opened 2025-09-27
Brings Back Lobby Chat

yupup123:
## About The Pull Request

Pretty much brings back lobby chat just like how it was in Ratwood with the anonymize button and an extra button for admins to be able to see it at all times if they so wish. No need to use LOOC anymore, so it's disabled in the lobby.

## Testing Evidence

image

Here's me schizoing out talking to myself. It works locally at least.

## Why It's Good For The Game

Lobby chat good. Lets players differentiate each other in the lobby and such.


RadiantFlash7:
I'm just about ready to full merge this, But I have a potential request if its not too difficult. Would you be able to change the color of lobby ooc, so its different then standard in game chatter?

PR #86, opened 2025-09-28
Armored Templar Buff

Skyglazer:
## About The Pull Request

As title, buffs the heavy armor Templar stats+skills. The changes are as follows:

Added: +2 int, +2 per (EDIT: Removed this.)
Adjusted: +1 wrestling, +1 athletics, up to Expert from Journeyman for both
Removed: 1 job slot, 4 -> 3, to account for its increased strength there shouldn't be as many of these guys running around-- and this is how it was on old-RW.

## Testing Evidence

Builds and runs. Just a numbers change.

## Why It's Good For The Game

Templars used to be a very strong force for the church back on old Ratwood. Azure Peak nerfed them quite severely, presumably to make room for roles like the Martyr as well as the 'Monk' Templar subclass, which is so much stronger than heavy armor Templar it's not even funny.

That subclass and the possible adjusting/removal thereof is a whole different subject entirely, so I've limited the scope of this PR to purely change the heavy armor Templar-- to put their wrestling and athleticism on-par with other martial fighters (such as men at arms, the other templar subclass, knights, etc.) and granting them some of their mental faculties that holy warriors ought to be revered for.

The role should be strong. You shouldn't have to run an unarmed build to make it good.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/church/templar.dm
+++ b/code/modules/jobs/job_types/roguetown/church/templar.dm
@@ -220,8 +220,6 @@
 		STATKEY_WIL = 3,
 		STATKEY_STR = 2,
 		STATKEY_CON = 2,
-		STATKEY_INT = 2,
-		STATKEY_PER = 2,
 	)
 
 /datum/outfit/job/roguetown/templar/crusader/pre_equip(mob/living/carbon/human/H)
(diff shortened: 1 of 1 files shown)


Pidgey-chan:
This new statline is superior to knights, which Templars already have superior equipment (special weapons and half-plate) over in addition to the obvious miracles.

CloverIsLucky:
Don't be fooled guys, this is a garrisonmain and they want to nerf the manpower of the clergy.

Taxing a job slot from the templar to give them expert athletics and wrestling doesn't seem worthy.

PR #94, opened 2025-09-29
Aavanic Murder Corp

ProbablyCarl:
## About The Pull Request
Adds in the TRACK SAPPER to the Steppesman loadouts. You're objectively worse in every regard to a standard sapper. Hands down. No contest.
One less axe/shield skill. No CON increase. -2SPD on top of -2SPD. No armour skill. Light armour setup. No dagger or whip.
What's the counter balance to that? You get three sticks of TNT, alongside an impact grenade of your choice. One impact. Any of them. Additionally, should they not take a grenade? They're instead given +2 levels in Alchemy/Engineering. I'm sure you can imagine why.
This entire role is basically the 'mad bomber'. That's the entire gimmick. We've no roles utilizing this stuff as is, beyond a change listed below to provide it (one stick) to Underdwellers.

Beyond that, light Steppesman infantry get access to the boar spear as an option for weaponry with their banner hanging off of it. Just a tiny resprite of the Frei spear, which was of Aavar make anyhow far as I'm aware.

I've also given the Underdwellers a single stick of TNT, since it fits them thematically and I don't know why they didn't already get something like this.

## Testing Evidence
image
image

image
image
image

## Why It's Good For The Game
It's FUN!!!!!!
Really though, just a way to put these items into use, and this is one of the more played mercenary roles. I wanted to give them something since they're not nearly as absurd as the other choices. Not that this is much better, beyond the spear addition, of course, but it's _something_.


CVSthePharmacy:
Also, will we finally get the aavnic weapons on the merchant silverface/goldface?

CVSthePharmacy:
Ping us on discord when this is tmed

PR #99, opened 2025-09-30
Reverts wood smelt result to coal

wow50000:
## About The Pull Request
See title

## Testing Evidence
one line change, I will give you my home adress so you can put a shotgun in my mouth if this actually breaks code wise

image


RadiantFlash7:
I'm going to say we're keeping charcoal- Though we really ought to add the recipe so that furnaces and such can be built with charcoal as well as coal.

CVSthePharmacy:
I love charcoal, so tasty, yum yum yum yum yum

PR #100, opened 2025-09-30
Lowers 2 Tier2 Spells to Tier1

Mryourface13:
## About The Pull Request

Lowers some utility spells to tier 1, so novice casters can use them.
Leap, Message

## Testing Evidence

image

## Why It's Good For The Game
Tier 1 is supposed to be for utility spells that aren't designed for combat specifically. These 3 spells will allow tier 1 casters some actual utility over the current lackluster tier 1 spell list we have. Our Tier2 spell list is massive...

Message is 100% a utility spell that everyone should get.
Leap is pretty utilitarian overall, and not necessarily required for good combat. Also allows for sneaky people to get more options.


Mryourface13:
P.S Balance tag probably fits here

PR #103, opened 2025-10-01
Removes male outfit requirement for horcs

snowysquirrel:
## About The Pull Request

There's some outfits that don't have a male sprite variant (i.e. armor bikinis), so feminine horcs would be doomed to nudity. This fixes that. It doesn't fix any weird spritework that may occur.

## Testing Evidence

**Leatherkini**
image

**Platekini**
image

## Why It's Good For The Game

This goes out to my favorite horc main In'ha. I told her I'd fix it if it required a single line change and, what do you know: it did! Some of the sprites will look odd (platekini as an example), but that's inherently better than them being invisible and thus unusable. I didn't test every single outfit that was broken, just a few.


RadiantFlash7:
Odd. We may want to instead simply block certain body types from wearing clothing types that don't have sprites for it, rather then making the illfitting sprite visible overtop

snowysquirrel:
i agree that a better method ought to be pursued, this is more of a bandaid solution until someone does it themselves
since it blocks outfits that do fit, i.e. leatherkini

if a horc main or coder went through and found the sprites that do fit horc female bodies but don't have a male variant in code, they could just duplicate the affected sprites and remove the part of the name that makes it for women only

PR #110, opened 2025-10-03
Attempt to make Shaman more unique (Hopefully)

Guidesu:
## About The Pull Request
This PR is oddly inspired on old Ratwood when we had apostate, which gave you specific statlines for specific inhumen gods. I tried to do same thing here. But hopefully balanced... i merely make the skill and perks to make the more distinct from each gods aside of miracles. And stats that hopefully won't be too OP? 

I am open for suggestion please don't flay me alive... still testing.

I essentially make them better at climbing (for some god forsaken reason a monk class don't have monk climbing skills) and swimming naturally and gave them 1 sneaking baseline.

## Testing Evidence
image

## Why It's Good For The Game
More diversity good!

### DIFF
--- a/code/modules/jobs/job_types/roguetown/mercenaries/classes/atgervi.dm
+++ b/code/modules/jobs/job_types/roguetown/mercenaries/classes/atgervi.dm
@@ -128,8 +128,8 @@
 			ADD_TRAIT(H, TRAIT_GRAVEROBBER, TRAIT_GENERIC)
 			ADD_TRAIT(H, TRAIT_NOSTINK, TRAIT_GENERIC)
 			ADD_TRAIT(H, TRAIT_INTELLECTUAL, TRAIT_GENERIC)
-			H.stats[STATKEY_INT] += 2
-			H.stats[STATKEY_STR] -= 1
+			H.change_stat[STATKEY_INT] += 2
+			H.change_stat[STATKEY_STR] -= 1
 		if(/datum/patron/inhumen/graggar)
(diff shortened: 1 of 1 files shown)


Guidesu:
Ssoooomething is wrong with the stats ugghr. Thoguht it was simple i gotta find the bug.

Ruediger4:
Okay if I may point out a few things:
- This makes Shaman the only merc class to be awfully inconsistent with whichever patron you go with - and not for the sake of 'being unique'. There's not a common ground on stat points, some skillspreads don't make any sense or are simply BETTER (Zizo getting Seedknow yet no farming? Mage armor but no spells? Graggar getting Sleuth AND perfect tracker AND outdoorsman AND battle ready AND master athletics AND master swimming AND expert climbing)
- The lines and code are a bit all over the place, they could use some better ordering for the sake of clarity.
- While the idea is cool, Apostates had different gear per class, different drip, different skill focus. They were balanced on their own way. Baotha and Zizo look like wimps next to Matthios and **especially** Graggar.

If you want to make something like Apostate, then maybe **making** Apostate is better. Otherwise, keep in mind that Shamans already get QUITE DIFFERENT depending on which patron they go for _because of Patron spells_. 
At most, maybe limit it to 1 statpoint or 1 trait per patron (Zizo Shamans could get +1 or +2 INT to make up for the INT malus, Graggar Shamans could get Nasty Eater and Sleuth, Matthy's can get +1 SPD, Baothans can get Good Lover and Deceiving Meekness for example)

PR #116, opened 2025-10-04
Return of Species Stress Events | Starsugar Nerf

ProbablyCarl:
## About The Pull Request
Two very simple changes.
- - - 
Firstly, pulls a change from SR([#672](https://github.com/Scarlet-Reach/Scarlet-Reach/pull/672)) and chops it apart.

Makes use of the vestigial Tolerant trait, by making folks who have it immune to negative species stress events on examine. Additionally introduces it as a virtue, if you want that, for some reason.

Xenophobic classes (just slavers for now) are hit by the old Xenophobic stress, unlike those without tolerant, who'll just get the old standard stress event.

- - -
Now, for the other, which should be atomised, but I'm lazy and angry. A flat nerf to Starsugar. All this does is tank the multiplicative slowdown. From a MASSIVE -2, to -0.5, which means you're still objectively faster than others. Just not teleporting wall-to-wall without sprint. A very simple but effective nerf. You still get DE/DS from it, too, alongside the other benefits. So it's not a big deal.

## Testing Evidence
image
image

## Why It's Good For The Game
Species interactions are fairly fun for both RP and otherwise. To have it locked as SR did behind a stress threshold, or paranoia? Well, it's certainly better than it used to be. We'll have to see if this is something we want.

Otherwise, starsugar BLOWS to fight. Everyone knows this. If you defend it, you're probably a starsugar user. I say this as someone who carries, at minimum, three at a time depending on role if I can get access to it. No more teleporting wall-to-wall, at least unless you're sprinting, which, y'know, I suppose live your life king.

### DIFF
--- a/code/modules/mob/living/carbon/human/examine.dm
+++ b/code/modules/mob/living/carbon/human/examine.dm
@@ -778,6 +778,19 @@
 				if (stuck_thing.w_class >= WEIGHT_CLASS_SMALL)
 					. += span_bloody("<b>[m3] \a [stuck_thing] stuck in [m2] [part.name]!</b>")
 
+	if(ishuman(user))
+		var/mob/living/carbon/human/H = user
+		var/stress = H.get_stress_amount()//stress check for racism
+		if(H.has_flaw(/datum/charflaw/paranoid) || stress >= 4)//Paranoid or stressed, for basic examine.
+			if(H.dna.species.name != dna.species.name)
+				if(dna.species.stress_examine)//some species don't have a stress desc
+					. += dna.species.stress_desc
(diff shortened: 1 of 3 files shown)


Hmeister-real:
> starsugar needs a buff it's too slow even before this pr, I should be able to vaporize people by running into them but also killing myself in the process, same for walls, lemme turn into a meaty satchel bomb

Let us force starsugar into people and watch them run into a wall and become alternative Japanese art

Pidgey-chan:
Should slavers become stressed from looking at their own slaves?

Hmeister-real:
> > Should slavers become stressed from looking at their own slaves?
> 
> yes

"Thieving Kobold... thank god it's on a leash..."

PR #121, opened 2025-10-05
Adds Cross Training of Skills

josephpatrickgreen:
## About The Pull Request
This PR adds the ability to train secondary skills from primary skills, for example, if you train how to use an axe, your lumberjacking gets better. If you swim, you become more generally athletic (at a low rate), if you train in armorsmithing, your general blacksmithing gets better, etc etc.

The skills are laid out and their secondary skills are assigned along with a multiplier rate.

## Testing Evidence
There is no testing done or necessarily required. It can not be perfect and I fully assume the crosstraining table MUST be tweaked and altered to be higher in the future. An example being, training from 0 to level 3 with swords wouldn't even get you a knives level 1

## Why It's Good For The Game
If you learn how to cut a tree, you at least know how to use an axe in a basic manner. If you learn how to smith armor, you can at least understand how a sword COULD be made. I feel this is immersive.

### DIFF
--- a/code/datums/sleep_adv/sleep_adv.dm
+++ b/code/datums/sleep_adv/sleep_adv.dm
@@ -180,12 +180,12 @@ var/global/list/CROSS_TRAINING_MAP = list(
 	//"slings" = (),
 
 	//"farming" = (),
-	"mining" = list("axes" = 0.1, "maces" = 0.1),
+	"mining" = list("axes" = 0.1, "maces" = 0.1, "athletics" = 0.1),
 	//"fishing" = (),
-	"butchering" = list("tanning" = 0.25, "medicine" = 0.25),
-	"lumberjacking" = list("axes" = 0.25, "unarmed" = 0.05),
+	"butchering" = list("tanning" = 0.25, "medicine" = 0.25, "knives" = 0.25),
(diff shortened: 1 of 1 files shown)


RadiantFlash7:
I'm not sure how to feel about this honestly. We don't want Combat roles to FURTHER creep into noncombat roles shtick.
However Useroth thinks this would be good to have;.

At a minimum We'd need to cap the skill gain from crosstraining to level 2 before I even think about tming this.

josephpatrickgreen:
> I'm not sure how to feel about this honestly. We don't want Combat roles to FURTHER creep into noncombat roles shtick. However Useroth thinks this would be good to have;.
> 
> At a minimum We'd need to cap the skill gain from crosstraining to level 2 before I even think about tming this.

I dont thinks so, most skills are at a .25 rate. so going from level 0 to level 4 would get you level 1 in that skill, I think that's actually okay.

josephpatrickgreen:
> Alright. With that much limitation- we can try test merging this. I still have some reservations, but we'll see where it gets us.

I train skills pretty regularly and havent seen this trigger once, but it DOES seem to make training the crosstrained skill get to novice faster.

PR #115, opened 2025-10-04
Ports AP's Handclaws with a twist! Disarm enemies!

Ruediger4:
## About The Pull Request
Ports part of https://github.com/GeneralPantsuIsBadAtCoding/Azure-Peak/pull/3917 - Mostly the **claws**! New unarmed weapons. They've a twist compared to AP!
- Adds all 3 claws (iron, steel, and Atgervi-Shaman unique). 
- Gives Iron and Steel claws smithing recipes.
- All 3 claws are available to be purchased as exotic weapons.
- ~~Buff most (if not all?) unarmed weapons - mostly damage. Katar gets peel to get a bone against the new unarmed weapons.~~
- ~~If the DMG values are too high, we can just shoot that shi' down.~~
- CLAWS ARE NOW UNIQUE! NO MORE STABBING, IMPALING OR RENDING! THEY CAN **DISARM** OPPONENTS!

The disarm checks your Unarmed Skill vs their Wrestling, and it is affected by your current stamina, your STR, the enemy's STR, and you get a bonus if you're dual-wielding.

## Testing Evidence
Works On My Machine I forgor to take screenshots

image

## Why It's Good For The Game
I mean more content is good. It may be bad. Can always just shoot this dead.

Edit: Carl didn't shoot me dead but demanded I became an ideasguy. I became, in fact, an ideasguy.

PR #123, opened 2025-10-06
PROPERLY ports AP's Handclaws with a disarming twist!

Ruediger4:
## About The Pull Request
This is #115 but properly properized. This will work. I hope.

## Testing Evidence
Look at #115 

## Why It's Good For The Game
More Weapons with a Niche


CVSthePharmacy:
Can we have a video demonstration of the weapon and how it works? Please?

Ruediger4:
> Can we have a video demonstration of the weapon and how it works? Please?

I'll get one later today after work, sure!

Pidgey-chan:
Needs major adjustment, the disarms are far too reliable for something that comes out only marginally slower than a lucerne smash, if having disarming possible through a single click on an intent is a good idea at all.

Ruediger4:
> Needs major adjustment, the disarms are far too reliable for something that comes out only marginally slower than a lucerne smash, if having disarming possible through a single click on an intent is a good idea at all.

I've been hearing feedback that they're "too reliable" at disarming people, so I may tweak the check to not be THAT good.

The Hooksword was THIS good, having used it in the past. I'll probably nerf the numbers a bit to make ABSOLUTE DISARMS (aka the claw user STEALING the weapon) harder to get by tweaking numbers.

As for the video thing, whenever i'm tryna record it, the fuckin file ends up weighing so much I can't reduce it so I may record it & upload it to some site for people to check. Probably after a few tweaks/nerfs

PR #127, opened 2025-10-07
Astratan Miracles & Resonance Introduction

ProbablyCarl:
## About The Pull Request
Replaces sacred flame at T1 for a miracle called 'Path of Order'. This new miracle allows Astratans to select between their two offensive miracles at will, on a ten minute cooldown, for a hundred devotion.

Sacred Flame:
 - Renamed to Fire Lance.
 - Damage from 10 to 25.
 - Modifier changed from 150% to 60% of an increase for damage against undead biotypes.
- ~~Firestacks for non-undead targets raised from 4 to 6.~~
 - Devotion cost lowered from 100 to 75.
 - It retains the cooldown of 15 seconds.

Path of Order:  image
 - 100 devotion cost.
 - 10 minute cooldown.
 - Allows selection between the lance or the spear. That's about it.

Summon Spear: image image
 - 30 second cooldown.
 - 100 devotion cost.
 - Puts a unique throwing weapon into your hand.
    - When thrown, spawns a mini thunderstrike at the target location, after a delay.
    - Initial middle area of strike is ~~40~~ 25 damage.
    - Surrounding tiles are ~~30~~ 19 damage.
    - Slowdown applied to all caught within the AoE.
    - +15 damage to undead.
    - You cannot attack with this object outside of throwing it. THROW IT!!!!

Fortify:
 - Now provides a visible overlay and message when active on a mob.

Resonance:
 - A unique trait intended for Astratans, recycled to the Priest exclusively.
    - Applies fortified to all around the trait holder in a two tile AoE, when a healing miracle is used. It was intended to buff miraclists nearby, but I scrapped the idea and couldn't find the want to complete it. Someone else can do it, if this is even something we want.

## Testing Evidence
Temp sprites. The spear, for example, is just the Martyr's sword in the ult animation but without the hilt. Because I'm a great, creative soul.
image
image
image
image
image
image

## Why It's Good For The Game
Sacred Flame used to be a meta miracle. Stupid powerful. Obscenely useful in near every situation. Instantly igniting up and down Zs. It was awful, to be clear, but it was _something_. Someone absolutely gutted it to TEN DAMAGE, with a miniscule amount of firestacks, and made it cost a hundred devotion to cast, alongside having a windup. Which is rather not great, to be very kind about it.
This change doesn't make it on-meta again, but it'll hopefully make it actually ok-ish. The rename is to reflect that it's not what it was, even remotely.

As for the spear miracle, it was intended to be higher tier, but I folded it into the ability to swap between. Taking the old devotion cost of sacred flame and shoving it into a slower cooldown miracle with AoE effects. The damage might be a bit high, really, and that's an easy fix. Just need to see it in play. Considering it takes a slot in hand, however? Alongside having no use outside of _throwing it_? I think it'll be fine.

Finally, in regards to Resonance, I'm sure you can imagine what the original intent was behind it. Instead, I'd scrapped the idea until I come back to it, and just gave what was able to be done in like five seconds to the Priest. Something for them to get, given otherwise they're in an odd spot.


Pidgey-chan:
Aren't Templars good enough? This is a lot of damage through armor and a lot of pain from the burns, too.

Skyglazer:
As we were talking about, it would probably be nice to incentivize actually hitting people with it- the delay on the thunderbolt means that you'd be wanting to aim it like ahead of people if you're chasing them down, which is a bit counter-intuitive. I think that with some tweaking, this would be a very worthy addition to an Astratan miracle-user's arsenal, to compete with other patron T2s- unlike the current Sacred Flame, which is a step or two above useless.

PR #129, opened 2025-10-08
Adventure reward

Rayforx:
## About The Pull Request
Changed the reward from
Low and high payment
Easy 15 - 25 to 30 - 60
Medium 30 - 50 to 75 - 125
Hard 100 - 140 to 250 - 300

Deposit from
5 easy
10 Medium
20 Hard
to
10 Easy
15 Medium
30 Hard

## Testing Evidence
N/A It's a define change.

## Why It's Good For The Game
A Hide costs 8, and fur costs 10
By hunting one wolf, it was possible to earn a greater reward with a less dangerous job than it would be to undertake a mission that carries a lot more risk.
Or you could cut 6 trees and get the same reward as a hard mission.
This made nobody consider doing the missions; it was simply not a good investment, and even more so when hard missions usually require a party, with some good gear. And then you had the repairs, potion cost, etc. That made the entire process a losing battle of finances for any adventure party. Ripped off would be the right word people use for the system, which is awful, since the quest system is awesome for the antags, as they can target adventures, and the people who wish to use the system.


Rayforx:
> More than doubling it is pretty extreme. I mean, especially when you will often encounter wolves to slay and skin along the way!

You can also go and try and get ambushed by going into a bush, by going back and forth
Or use bait, for that matter
2 meat, 8
2 hide, 16
1 fur, 10
1 volf head, 10
This is 44 for one volf.
This is a lot easier than doing a bounty, and honestly, a lot more rewarding.
I could honestly go on with a lot of the safer jobs around, but this feels like a good example of what an adventure usually does to earn quick cash.
I do not see why this PR is extreme. 

Encountering something along the way should not impact the balance; a woodcutter, smith, servant, or miner can all find a volf on the road, kill it, and get the fur and hide for it. This should not impact how much the miner gets from the mines, or the woodcutter gets for wood.

PR #142, opened 2025-10-11
More bounties

Rayforx:
## About The Pull Request
Adds four more beast-related contracts
Everything works, drafted because of another PR that will conflict with this one, and for balance reasons

## Why It's Good For The Game
More variety, and a framework to add monsters to the various contracts


code/modules/roguetown/roguemachine/questing/monsters/questbehemoth.dm
```diff
@@ -0,0 +1,163 @@
+/mob/living/simple_animal/hostile/retaliate/rogue/elemental/questbehemoth
```
ProbablyCarl: Could save the headache and make it a subtype, if it's identical for most of what's here.

PR #145, opened 2025-10-11
Trades a Warden's Expert axes for Journeyman in a wider variety of weapons (except not really now)

Jamdrawers:
## About The Pull Request

-Trades away the Warden Forester's Expert axe skill for Journeyman in axes, swords and maces, making them journeyman in all of them instead of expert in one, journeyman in another, and having no skill at all in the rest
-Warden Ranger gets Apprentice in a variety of melee weapons to meet their apprentice skill in axes.
-Adjusts descriptions to match
-Changes the text on the helmet-choice to more accurately inform you what you're doing (choosing a helmet)

## Testing Evidence

image

## Why It's Good For The Game

Means you can have a modestly wider variety of playstyles as a Warden Forester and can use more of the weapons you might scavenge while out doing warden stuff. Also means you don't feel so bad about using the other weapons because you're not significantly worse with them than with your axe.
And I don't think wardens needed expert weapon skill - proficient works serviceably enough, especially as wardens have so much else going for them beyond being an expert axe-duellist.

It makes them wind up mechanically a little closer to the old bog/vanguard; I don't know whether we'll be eventually rebranding them more towards one of those but in the meantime I think this would be a good change.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/garrison/bogguard.dm
+++ b/code/modules/jobs/job_types/roguetown/garrison/bogguard.dm
@@ -85,7 +85,6 @@
 	H.adjust_skillrank(/datum/skill/combat/knives, 3, TRUE)
 	H.adjust_skillrank(/datum/skill/combat/axes, 2, TRUE)
 	H.adjust_skillrank(/datum/skill/combat/swords, 2, TRUE)
-	H.adjust_skillrank(/datum/skill/combat/maces, 2, TRUE)
 	H.adjust_skillrank(/datum/skill/combat/polearms, 2, TRUE)
 	H.adjust_skillrank(/datum/skill/misc/athletics, 4, TRUE)
 	H.adjust_skillrank(/datum/skill/misc/climbing, 4, TRUE)
@@ -151,11 +150,9 @@
 		/obj/item/rogueweapon/scabbard/sheath = 1,
 		/obj/item/signal_horn = 1
 		)
-	H.adjust_skillrank(/datum/skill/combat/axes, 3, TRUE)
+	H.adjust_skillrank(/datum/skill/combat/axes, 4, TRUE)
 	H.adjust_skillrank(/datum/skill/combat/polearms, 3, TRUE)
 	H.adjust_skillrank(/datum/skill/combat/swords, 3, TRUE)
-	H.adjust_skillrank(/datum/skill/combat/maces, 3, TRUE)
-	H.adjust_skillrank(/datum/skill/combat/whipsflails, 2, TRUE)
 	H.adjust_skillrank(/datum/skill/combat/shields, 2, TRUE)
 	H.adjust_skillrank(/datum/skill/combat/slings, 2, TRUE)
 	H.adjust_skillrank(/datum/skill/combat/bows, 1, TRUE)


Dedwolf:
this would be much better implemented if you gave them a choice of primary weapon upon spawning in which gives them +1 skill to whatever weapon they pick, meaning they'd be journeymen (or apprentice if you want to lower it) with other weapons while retaining the axe as a sidearm and utility weapon while swapping the spear out for, say, a sword or mace or whatever

other roles have this implemented already, check out the MAA, the knights, and some of the adventurers and mercs

Jamdrawers:
> I'm skeptical of this honestly. Part of the melee warden's identity is being good with their axe. Lessening it, to give other skills feels very meh in my opinion. I could see instead slotting the melee warden with a few journeyman/apprentice ones, like the ranger? But it should primarily be an axe combatant- it's their unique weapon after all.

also as a unique weapon I think it works great as it is as a *utility*, with the choppy bit for chopping trees and removing the heads of bounties and the picky bit being good for stone. Makes it feel weird that they start with a spear that you're disincentivised to use in favour of expert axe-fu.

code/modules/jobs/job_types/roguetown/garrison/bogguard.dm
```diff
@@ -106,7 +109,7 @@
 		"Path of the Bear"		= /obj/item/clothing/head/roguetown/helmet/sallet/warden/bear,
 		"None"
 	)
-	var/helmchoice = input("Choose your Path.", "HELMET SELECTION") as anything in helmets
+	var/helmchoice = input("Choose your Helmet.", "HELMET SELECTION") as anything in helmets
```
RadiantFlash7: No need to remove path flavor

code/modules/jobs/job_types/roguetown/garrison/bogguard.dm
```diff
@@ -179,7 +185,7 @@
...
```
RadiantFlash7: Keep path flavor

RadiantFlash7:
I'm not opposed to journeyman swords on forrester.
However, I don't think maces, nor whips/flails are particularly fitting for the role's flavor.
Expert axes should also definitely be kept.

Make those changes and I'll see about having this TM'd.

code/modules/jobs/job_types/roguetown/garrison/bogguard.dm
RadiantFlash7: Hm. On second thought, thats fair.

Jamdrawers:
Kinda - though for me the focus wasn't reaaaally on nerfing as a primary goal of just wanting Wardens to be less powerful overall (though I still wanted to do that as a secondary goal, as someone who has been mostly playing Warden lately), the main intent was to make more weapon options viable options compared to axes by not making axes better for them.
Since like, a melee warden starts with a big spear and a little axe sidearm, but there's little reason to use the spear because you have a lower skill with it and it's bigger and more unwieldy and you have easy access to long axes anyway.

But I guess with the way AP balance is set up there's a fear of letting any primary non-adventurer combat class ever dip above or below exactly expert in their highest melee skill (perhaps because the gap between each skill level is too high), aaaaand I guess the Woodsman flavour (which i feel could be achieved with a smidgen of lumberjacking or carpentry more than expert axe-fu. Compared to men-at-arms they could be less of an expert fighter but more rounded out with skills for surviving out in the wilderness)

But Alas. Maybe another day...

PR #150, opened 2025-10-12
Necromancer changes

RadiantFlash7:
## About The Pull Request
This PR ports a changes to raise undead by making it raise lesser undead formation, Summoning 3 skeletons to fight alongside the caster. Simple undead skeletons are now temporary, and start taking damage after x amount of time being present (To prevent simple NPCs from overwhelming our AI controller long term)

It also adds raise undead as a form of raising a single NPC Skeleton at the cost of 4 bones.

A new command spell has been written up and given to necromancers called 'command undead'.

This spell will give orders to both Simple skeletons and NPC skeletons.  It's a bit sloppily done to hybridize that, but for now its what we'll go with- subject to change.
Finally another spell has been added called tame undead- allowing necromancers to raise dead animals as undead ones,
as well as taming rogue skeletons to not be hostile to you.
To account for this bit of buff in power, Necromancers have lost 6 spellpoints from their starting 18, bringing them to 12.

## Why It's Good For The Game
Current interation of necromancer is woefully inadequate. This should help, I think.

### DIFF
--- a/code/game/objects/items/rogueitems/necro_relics.dm
+++ b/code/game/objects/items/rogueitems/necro_relics.dm
@@ -14,7 +14,7 @@
 /obj/item/necro_relics/necro_crystal/attack_self(mob/living/user)
 	..()
 
-	if(!user) 
+	if(!user)
 		return FALSE
 	if(world.time - src.last_use_time < src.use_cooldown)
 		to_chat(user, span_warning("The crystal thrums under your touch, but remains inert."))
@@ -30,7 +30,7 @@
 		user.flash_fullscreen("redflash1")
 		new /obj/item/natural/glass_shard(get_turf(src))
 		playsound(src, "glassbreak", 70, TRUE)
-		qdel(src) 
+		qdel(src)
 		return FALSE
 
 	var/turf/T = get_step(user, user.dir)
@@ -58,12 +58,13 @@
 	addtimer(CALLBACK(target, TYPE_PROC_REF(/mob/living/carbon/human, choose_name_popup), "FORTIFIED SKELETON"), 3 SECONDS)
 	addtimer(CALLBACK(target, TYPE_PROC_REF(/mob/living/carbon/human, choose_pronouns_and_body)), 7 SECONDS)
 	target.mind.AddSpell(new /obj/effect/proc_holder/spell/self/suicidebomb/lesser)
+	target.faction |= list("[user.mind.current.real_name]_faction")
 
 	to_chat(user, span_notice("The crystal dissipates into dust."))
 	user.flash_fullscreen("redflash1")
 	new /obj/item/natural/glass_shard(get_turf(src))
 	playsound(src, "shatter", 70, TRUE)
-	qdel(src) 
+	qdel(src)
 
 	return TRUE
 


RadiantFlash7:
> Actual gem necromancer content update. The skeletons they spawned were woefully weak that they'd get like two shot anyhow so good to see that people can actually be scared of necromancer now.
> 
> Ill take any spellpoint drop to keep this merged, a suggestion might be being able to select the different variations of NPC skeletons (ex. the pirate ones) and adding something to the code so you dont have to use the tame spell on the nec armoured ones so that the other skeletons dont aggro on them and that you don't need to gravemark your own crystal spawned skeletons so they dont get aggro'd on either.
> 
> Otherwise good PR thank you for the gem 💎

I'll probably be bumping up simple mob skeleton health as well- But. I have 3 concerns predominantly with my necromancer changes right now. 

Sometimes, for reasons I have not yet parsed; Summoned Simple skeletons turn hostile towards Summoned NPC skeletons- seemingly randomly. Sometimes they play nice, other times they don't. 

Secondly- Getting bones was mainly intended to be from grave robbing- but I think the simple skeleton mobs end up dropping 3 bones per, which means you can kill your own simple skeletons to summon an NPC one.

Finally, Bones are EVERYWHERE. Animals leaving them behind means players often don't bother to pick them up as they have little use- meaning necromancers have free material just- sitting around. Leading to unintended consequences of necromancers able to field quite a few NPC skeletons- Though it hasn't proven a problem yet.

RadiantFlash7:
> I do agree with all your points but I think that limiting bones to graverobbing or natural genned skeletons would be a bit too harsh. If there could be a compromise with perhaps spawned skeletons only dropped one bone considering that 3 skeles spawn. That'll mean it'll take 2 spell casts to get the 1st and 3 for the 2nd. It just slows down the growth while not overly stopping it entirely.

That could be a pretty good option. I might also see about potentially adding a use for necrans to 'put bones to rest' as a way of encouraging them to be collected, and hopefully reducing the animal bone flood.

So far the NPC skeleton seems to be in a pretty good place. having level 3 skills across the board, all wielding a shield, with either a flail a sword or a spear. And they have a helmet, so people need to actually break it to bust through to one tap the skeleton.

atomicplan666:
image

When you tame skeletons you cannot control them

atomicplan666:
Casting "Command Undead" on greater skeletons should be toggling idle/aggressive, but does not - 
image

atomicplan666:
@RadiantFlash7 you're a legend

RadiantFlash7:
> @RadiantFlash7 you're a legend

If you find any other over issues let me know.
At present, tame undead isn't meant to make them commandable, just make them passive to you. Partly because orderable NPC skeletons are a subtype.

RadiantFlash7:
> yo idk if someone's said this before but in my experience my lesser undead formations kept attacking my summoned (armored, nonsentient) undead pls fix (gravemark does nothing when used on summoned skeletons that aren't the sentient ones)

I thought I fixed this not long ago. Ugh. I'll run some more tests see if I can figure out whats up.

Jamdrawers:
Also I think there should probably be a max range put on the summon-skeleton spells, feels weird that you can cast it anywhere you can see, at super far range and from different z-levels

PR #154, opened 2025-10-13
Reimplements Merar and Canilunzt

Dedwolf:
## About The Pull Request

Direct counter of https://github.com/GeneralPantsuIsBadAtCoding/Azure-Peak/pull/2262

Reintroduces canilunzt and felid languages and sets them as standard for lupians, venardines and tabaxi.

## Testing Evidence

image

## Why It's Good For The Game

These languages were originally removed under the idea that there were no existing states or regions associated with them. We do in fact have Lupian nations and lore for them, the language is spoken throughout their culture. Not sure about the tabaxi language though, it could use a nicer name but i'm adding it back anyway for now. 

I believe it is due to Azure Peak using entirely different lore that they weren't able to keep it in.

I'll be writing additional lore for the lupians in time but for now I believe reintroducing the language meets the standards for reimplementation.

### DIFF
--- a/code/modules/language/roguetown/grenzelhoftian.dm
+++ b/code/modules/language/roguetown/grenzelhoftian.dm
@@ -1,6 +1,6 @@
 /datum/language/grenzelhoftian
 	name = "Grenzelhoftian Imperial"
-	desc = "The primary language of the Grenzelhoft Emperiate, a robust language filled with hearty words and enthusiastic enunciation. Often said to be similar to the Lupians Canilunzt."
+	desc = "The primary language of the Grenzelhoft Emperiate, a robust language filled with hearty words and enthusiastic enunciation."
 	speech_verb = "remarks"
 	ask_verb = "inquires"
 	exclaim_verb = "asserts"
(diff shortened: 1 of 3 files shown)


code/modules/language/roguetown/grenzelhoftian.dm
```diff
@@ -1,6 +1,6 @@
 /datum/language/grenzelhoftian
 	name = "Grenzelhoftian Imperial"
-	desc = "The primary language of the Grenzelhoft Emperiate, a robust language filled with hearty words and enthusiastic enunciation."
+	desc = "The primary language of the Grenzelhoft Emperiate, a robust language filled with hearty words and enthusiastic enunciation. Often said to be similar to the Lupians Canilunzt."
```
ProbablyCarl: Said, but not functionally similar, nor does it share syllables. Given the span being the only similarity, which is meta info? It should be dropped. Not that players will typically see this, but still.

code/modules/language/roguetown/merar.dm
```diff
@@ -0,0 +1,18 @@
+/datum/language/merar
+	name = "Merar"
+	desc = "The desert sands the ancestors used to walk echoes through this raspy tongue that's mostly composed of murrs and mrrps with the ocassional meow or hiss if one is getting quite vocal. Hailing from the desert tribes and kingdoms ruled by the swift tabaxi this language is sharp and direct. With a keen ear you can notice the vague gizan influence on this language."
```
ProbablyCarl: gizan > Gizan

code/modules/language/roguetown/grenzelhoftian.dm
Dedwolf: yep, will do

PR #156, opened 2025-10-14
The Return of Racism

Jamdrawers:
## About The Pull Request

Restricts the following classes from Shunned or lower races:
Duke
Prince
Knight+Knight Captain
Marshall
Councillor

I left open other nobility roles like court physician, steward and magos because they could feasibly be attained on specialised merit by just being a good banker, as well the Hand who is personally elected. Suitors are also welcome to try their hand, even if their odds might be slim.

I also moved Lamias up a notch from Shunned since I don't *think* it's that incongruous to see them in leadership roles, though I can go either way on that depending on how people feel about lamias.

For reminder, Shunned or Lower means Goblins, Verminfolk, Kobolds, Halforcs and Dullahan.
I could also go either way on Dullahan, depending on how we want to go about Dullahan lore

If we're set on wanting to keep Verminfolk and Kobold and Half-Orc etc dukes, we could alternatively lessen it to merely the *despised* race (goblins) not being allowed to be knights or dukes; it's definitely possible to do this kind of thing in a way that'd too heavy-handed, such as, like, banning non-humans from basically everything, but also I think the current situation is too light-handed. Good to have a middleground there.

## Testing Evidence

image

## Why It's Good For The Game

Goblin dukes are silly and I think we've had our fun playing with that silly gimmick, time to put the sensible trousers back on. Removing racelocks was never a conscious decision, just something that was inherited with the rebase, designed for a different server with a different set of priorities and preferences. We were all pretty happy with it in RW1 so lets keep it going.

### DIFF
--- a/code/__DEFINES/roguetown.dm
+++ b/code/__DEFINES/roguetown.dm
@@ -221,13 +221,13 @@ Balloon Alert / Floating Text defines
 	/datum/species/akula,\
 	/datum/species/anthromorph,\
 	/datum/species/demihuman,\
+	/datum/species/lamia,\
 
 #define RACES_SHUNNED \
 	/datum/species/halforc,\
 	/datum/species/anthromorphsmall,\
 	/datum/species/kobold,\
 	/datum/species/dullahan,\
-	/datum/species/lamia,\
 
 #define RACES_DESPISED \
 	/datum/species/goblinp,\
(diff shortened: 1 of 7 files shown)


Pidgey-chan:
Good idea, shame it got speedclosed lol. Lore should be represented mechanically and this would be a good way of doing it, just like it was in oldwood. Revolting is NOT a solution when a goblin never should have risen to Duke anyway.

PR #164, opened 2025-10-16
Pioneer

DelineFortune:
## About The Pull Request

https://github.com/Scarlet-Reach/Scarlet-Reach/pull/1192
1 vendor in hamlet
2 vendor around bog vendor

## Testing Evidence

image
image

### DIFF
--- a/_maps/map_files/rockhill/rockhill.dmm
+++ b/_maps/map_files/rockhill/rockhill.dmm
@@ -39179,6 +39179,7 @@
 /area/rogue/outdoors/bograt/above)
 "nFK" = (
 /obj/structure/mineral_door/wood/towner/miner,
+/turf/open/floor/rogue/dirt/road,
 /area/rogue/indoors)
 "nFR" = (
 /turf/open/floor/rogue/blocks/stonered,


snowysquirrel:
perhaps add pickaxes and a backpack to pioneer's hoardmaster? would match what they are + they spawn with these two things

code/modules/jobs/job_types/roguetown/adventurer/types/antag/pioneer.dm
```diff
@@ -0,0 +1,187 @@
...
+	releasedrain = 0
+	recharge_time = 90 SECONDS
+	max_targets = 0
+	cast_without_targets = TRUE
+	antimagic_allowed = TRUE
+	associated_skill = /datum/skill/craft/traps
+	invocations = list("Measure twice… set once.")
+	invocation_type = "whisper"
+	miracle = FALSE
+	req_items = list(/obj/item/rogueweapon/shovel/)
```
ProbablyCarl: Path doesn't need to have /

anconfuzedrock:
yeahh this seems a bit much. I like the tradeoff of being able to do house stuff as bandit but being worse in combat, but- they don't seem THAT bad in combat, and they do have the mines too. maybe lower the ranged skills, con, end or per.

On the other hand, they have a lot of skills, but it's much harder to abuse carpentry and masonry than it is 4 str and expert weapons. do they even have carpentry? idk

CVSthePharmacy:
> I ditto Carl's thing. I get it that its a Bandit subclass but they basically just... make a lot of things redundant. And bandits are already pretty strong _FROM THE START_.
> 
> * They already often go to dungeons to loot. This will ALSO make them just get gear to smelt to get MORE gear that's not in their store. Eliminates any idea of taking a hostage that may be able to smith for them, too.
> * Master xbow/bow when Knave gets Expert tops.
> * Demo modifier of **7** is insane. Steppesmen's axe gets a 2 and it is already REALLY QUICK. Lower it to 2 as well (or 1.75 maybe)
> 
> This needs a lot of work imo. Just because it flies/"works" in Scarlet doesn't mean it'll work here.
> 
> ALSO another thing: They start with, arguably, a genuinely really good gear loadout from the get-go. No other class (except Hedgeknight and, arguably, Iconoclast) gets this.
> 
> * Brigands start with shit hide armor and some bodyparts are flat out exposed, on top of iron weapons.
> * Knaves start w/a basic leather armor
> * Sellswords start w/a gambeson and maille shirt, as well as a gorget. No more.
> * Rogue Mage and Sawbones may as well have no armor.
> 
> Meanwhile this one starts with... A metal helm, studded leather over a gambeson, OK gloves, and something that'll stop a crit on the neck and pants.
> 
> Is it great gear? No. But when 5 out of 7 roles **need** to use their starting favor to buy shit, it's kinda... noticeable.

Okay a few things, from my opinions. 

I think pioneer should start with a gambeson and a steel chain coif, with a padded cap, BOOM the armor problem is solved

Shovel should have a demo mod of 4 since, their whole thing is just demolition and building. Along with their crossbows skill being lowered to journeyman, same with bow skills. 

Also nobody does smelt scumming here as a bandit... I play bandit.... we don't tolerate that shit.... we're not SR....

snowysquirrel:
more suggestions;
- give masonry skill (already has pretty much every construction skill, carpentry included, may as well allow them the flavor? i could see one fortifying mount decap and turning it into an external bandit extension if the bandits were particularly strong that round. would also allow them to fill in keep walls where broken, hiding signs of forced entry via any walls they break)

- reduce ranged skill (could remove it entirely? what do they need it for, realistically? their niche lies elsewhere. i could maybe see them retaining crossbow access, that fits a pioneer/engineer type)

- maybe remove lockpicking skill (why do they need it? knave specializes in it already, w/ none of the other bandit classes having lockpick skill. this is the one bandit class that probably doesn't need lockpicking at all, given they can break in wherever with their i-hate-society-and-its-structures shovel) 

- seems like a class that should be allowed some shield skill (no armor training, no dodge expert, journeyman weapon skills)

ProbablyCarl:
Superseded by another PR. Closing.

PR #172, opened 2025-10-18
Balance tweaks

RadiantFlash7:
## About The Pull Request
Martyr sword peels,
Blunt strikes do more damage to armor,
Smash does more damage in general.
Dendorite zapshroom now have chargeup and an icon when activating.

## Why It's Good For The Game
Blunt weapon ap changes were good, But unfortunately strike still tends to be somewhat pitiful. Decided mace strike can differentiate from other weapons by dealing more integrity damage to armor. Smash on the other hand, does more damage overall.
The martyr sword has also been given peel as every other longsword has it, and it didn't make much sense for the martyr to have to buy a second sword to do any decent AP when not using their 'ult'.
Druid kneestinger instant hardstun memeing is unfortunately a bad meme. This will hopefully curtail it to a degree, subject to change.

### DIFF
--- a/code/game/objects/items/rogueweapons/melee/blunt.dm
+++ b/code/game/objects/items/rogueweapons/melee/blunt.dm
@@ -8,7 +8,7 @@
 	chargetime = 0
 	penfactor = BLUNT_DEFAULT_PENFACTOR
 	damfactor = 1.1
-	intent_intdamage_factor = 1.35
+	intent_intdamage_factor = 1.15
 	swingdelay = 0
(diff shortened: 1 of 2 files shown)


ProbablyCarl:
PR looks good, aside from the mace change, because it doesn't account for flails.
Also, keep in mind that this change will stack with a weapon's inherent intdamage_factor. It applies both intent and weapon factor in tandem.
Maces, aside from godens, will be doing 70% more integrity damage per strike. Warhammers at 55%.

PR #173, opened 2025-10-18
Ports training sword and blunt arrows from AP

RadiantFlash7:
## About The Pull Request
Ports https://github.com/GeneralPantsuIsBadAtCoding/Azure-Peak/pull/4005
Training swords are swords that have combat intents while being dull. Good to practice differant intents in sparring.

## Why It's Good For The Game
Sparring my beloved. This was a pretty cool PR from Azure Peak.

### DIFF
--- a/code/game/objects/items/rogueweapons/melee/swords.dm
+++ b/code/game/objects/items/rogueweapons/melee/swords.dm
@@ -101,6 +101,7 @@
 
 /datum/intent/sword/peel/weak
 	name = "weak armor peel"
+	attack_verb = list("<font color ='#e7e7e7'>lightly peels</font>")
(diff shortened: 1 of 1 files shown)


code/game/objects/items/rogueweapons/melee/swords.dm
```diff
@@ -92,6 +99,10 @@
 	reach = 2
 	peel_divisor = 5
 
+/datum/intent/sword/peel/weak
```
ProbablyCarl: Add the unique peel attack verb to this. Specifically, we've already got a 'weakly peels' and 'strongly peels', on top of standard peeling. So perhaps better wording for this would do.

PR #180, opened 2025-10-19
Removes Silly Weapon Skills

Jamdrawers:
-Tools such as the frying pan, pick and scythe use maces, maces and polearms instead of cooking, mining and farming for combat calculations like parry chance. 
-The sickle and thresher now are effected by knife and whip skill!
-Militia weapons use normal combat skills too, instead of using Mining to determine parry chance with a milita war-pick
-Militia weapons are repaired with weaponsmithing too instead of carpentry, because they're made with as much metal as a spear (A majowski is just an ugly sword! It isn't even made of wood!) (I could go either way on this one) 
-Miner has been compensated by getting maces instead of polearms like in RW1 (no idea why they got polearms to begin with)

## Testing Evidence

I wanted to make sure that changing the skill used in the code wouldn't negatively impact mining ability, so I made a testing area and a testing pick, so that I had one pick that used mining and one that used maces, and both of them behaved exactly identically while mining, but of course had different parry chances against skeletons (though against spiders is was still 90% at journeyman
image
image

## Why It's Good For The Game

Overall the cooking-to-parry-chance mechanics were silly. Towner roles already -get- combat skills (soilers get 3 in polearms, miners get 3 in axes and (now) polearms), and I think it's better to make use of those skills than to ignore them in favour of legendary pick-fu. It's more organic and natural. This also coincides with cross-training.
We wouldn't expect Woodcutters to use their lumberjacking skill to determine how well they can parry a sword with their woodcutting axe, and we wouldn't expect tailors to use their sewing skill to determine how well they can shank someone to death with their scissors.

It's a mechanic we've inherited from AP and not one intentionally put in ourselves.

### DIFF
--- a/code/game/objects/items/rogueweapons/melee/flail.dm
+++ b/code/game/objects/items/rogueweapons/melee/flail.dm
@@ -167,7 +167,7 @@
 	wbalance = WBALANCE_HEAVY
 	smeltresult = /obj/item/ingot/iron
 	associated_skill = /datum/skill/combat/polearms
-	anvilrepair = /datum/skill/craft/carpentry
+	anvilrepair = /datum/skill/craft/weaponsmithing
 	dropshrink = 0.9
 	wdefense = 4
 	resistance_flags = FLAMMABLE
--- a/code/game/objects/items/rogueweapons/melee/special.dm
+++ b/code/game/objects/items/rogueweapons/melee/special.dm
@@ -275,7 +275,7 @@
 	force_wielded = 25
 	minstr = 10
 	max_blade_int = 130
-	anvilrepair = /datum/skill/craft/carpentry
+	anvilrepair = /datum/skill/craft/weaponsmithing
 	smeltresult = /obj/item/rogueore/coal
 	wdefense = 4
 	wbalance = WBALANCE_HEAVY
@@ -292,7 +292,7 @@
 	minstr = 8
 	max_blade_int = 120
 	max_integrity = 200
-	anvilrepair = /datum/skill/craft/carpentry
+	anvilrepair = /datum/skill/craft/weaponsmithing
 	smeltresult = /obj/item/rogueore/coal
 	resistance_flags = FIRE_PROOF
 	light_system = MOVABLE_LIGHT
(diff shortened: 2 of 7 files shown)


Jamdrawers:
> while probably a good change, i will miss being a dwarffortressesque pickaxe warrior

Picks will use maces, and miners will get mace skill, so it'll still be viable, you just won't be doing Legendary unblockable ultra-pick-fu

snowysquirrel:
i would like it if these classes received some other quirk or buff for using their tools in combat because it's funny and cute
these classes should be incentivized to still prefer their tools in combat in some way, otherwise the flavor of the tool being able to be used as a weapon is gone

Jamdrawers:
> By removing flavor skills from tools, we require roles that frequently use them to become actual combatants outside of that one very specific tool, and they frequently don't have the athletics, wrestling or statspread for it to matter outside of initial encounters.
> I get the intent behind this PR, but I'm wary. AP gave non-combat classes a way to participate and engage in mechanics if they're caught out, and I appreciate the idea. Not that they should be, of course, but still. Hesitant to approve this.

AP kinda already gave non-combat classes that way to participate by already giving them moderate combat skill! 
Just doesn't give them a way to participate with _Legendary_ skill, because obviously they're not supposed to.

Jamdrawers:
Whoops misclick

Vizoplez:
This PR isn't needed.

Due to stat inflation and skill inflation this isn't like ratwood 1.0 where a farmer was the best combat class in the game with a scythe.

It's a concern, yes, but statistically and skill-wise the farmer will get outstammed and beheaded immediately. Not a needed PR.

Pidgey-chan:
Weapon skills aren't meant to be grindable alone or raised above jman with virtues, I don't see why picks and pans should be any different. Especially picks, which are more than viable weapons; I might even call them meta.

Not to mention that cooking and mining have nothing to do with fighting and parrying. Rocks don't hit back.

ProbablyCarl:
This has been sitting for some time. To clean up the Github, I'll be closing it. This isn't an outright denial of it, given there's merit for it. Please put up a new PR if you wish to see this in again, or I'll probably do it myself at some point in the future.

PR #192, opened 2025-10-23
Matthios' Theft of the Flame [Draft]

josephpatrickgreen:
## About The Pull Request

Adds Astrata's Flame, an objective for bandits.

-1 lit brazier (astrata's flame) is placed on the top floor of the church
-1 unlit brazier is placed at the bandit camp
-a player may steal the lit flame with a torch/lamptern, snuffing the brazier's flame in the process, the special brazier can not be put out by any other means
-when lit by this holy source the torch/lamptern becomes "astrata's flame" and notifies all astratans what has happened
-astrata's flame can never be extinguished only placed onto a torch or one of the two braziers
-when lit at the church, church areas have a passive devotion increase and a flat astrata influence increase
-when lit at the bandit camp it provides bandits a passive point increase and a flat matthios influence increase, further it informs all church staff of the bandit camp location
-when in torch/lamptern form, astrata's flame provides +4 perception when held in your hand

## Testing Evidence

none yet

## Why It's Good For The Game

A fun capture the flag mini objective

### DIFF
--- a/code/game/objects/items/devices/flashlight.dm
+++ b/code/game/objects/items/devices/flashlight.dm
@@ -16,7 +16,9 @@
 	var/weather_resistant = FALSE
 	possible_item_intents = list(INTENT_GENERIC)
 	var/on = FALSE
-
+	var/divine = FALSE
+	var/original_name = ""
+	var/original_color = "#f5a885"
 /obj/item/flashlight/Initialize()
 	. = ..()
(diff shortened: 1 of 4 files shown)


josephpatrickgreen:
> very much vibe with it! will it only be matthites (and or bandits) who can steal it?

hypothetically if there was another major faction, like the wretch camp? vampires? maybe it would give them specific buffs too.

PR #194, opened 2025-10-24
Unarmed Destruction

ProbablyCarl:
## About The Pull Request
~~One line adjustment.~~
Now applies diminishing returns after 14STR.
```
Normal scaling: +20% per point over 10

Diminishing returns after 14
Start with the full +0.8 from 14 STR

Each point beyond 14 gives a smaller bonus than the last:
e.g., +0.1, +0.075, +0.05625, etc.
```
As opposed to 80 per hit at 20STR without gauntlets...

20STR = 42.1 UNARMED | 50.5 PLATE GAUNTLETS
14STR(BEFORE LOSS OF RETURNS) = 22.4 UNARMED | 26.9 PLATE GAUNTLETS
9STR(BEFORE BONUS) = 7.2 UNARMED | 8.6 PLATE GAUNTLETS

~~20STR = 72 BRUTE with PLATE GAUNTLETS~~
~~14STR = 22.4 BRUTE | 25.8 with CHAIN GAUNTLETS | 30.2 with PLATE GAUNTLETS~~

~~20STR = 25.4 BRUTE~~
~~14STR = 15.3 BRUTE~~
~~1STR = 1 BRUTE~~

Is this good? I 'unno. CivBarb will still help for actual crits. But this'll stop a very particular playstyle that is very gross and wildly blowing everything else out of the water. Thanks, unarmed. Continuing to be the pain that you are.

## Testing Evidence
Very simple change. Tested numbers are above.

## Why It's Good For The Game
I didn't think I'd need to argue this. But I do.
Extreme edge cases had brawlers outcompeting anything else by far. It felt like the katar all over again.

Unarmed cannot be disarmed. You cannot take it. You cannot prevent it from being used, unless you outright break someone's arms. It was the equivalent of bringing a hand grenade to a knife fight in a lot of cases. Classes like berserker were better off discarding knuckles / katars as a result.

Further adjustments to this have retained it as a viable strategy, to be clear, if you've the strength. It's a simple 0.3 to 0.2.

### DIFF
--- a/code/modules/mob/living/carbon/human/human_helpers.dm
+++ b/code/modules/mob/living/carbon/human/human_helpers.dm
@@ -131,7 +131,24 @@
 		used_str = get_str_arms(used_hand)
 
 	if(used_str >= 11)
-		damage = max(damage + (damage * ((used_str - 10) * 0.2)), 1)
+		var/bonus = 0
+
+		if(used_str <= 14)
+			// Normal scaling: +20% per point over 10
+			bonus = (used_str - 10) * 0.2
+		else
+			// Diminishing returns after 14
+			// Start with the full +0.8 from 14 STR
+			bonus = 0.8
+			var/extra = used_str - 14
+			// Each point beyond 14 gives a smaller bonus than the last:
+			// e.g., +0.1, +0.075, +0.05625, etc.
+			var/next_bonus = 0.1
+			for(var/i = 1, i <= extra, i++)
+				bonus += next_bonus
+				next_bonus *= 0.75 // reduces 25% each time
+		damage = max(damage + (damage * bonus), 1)
+
 	if(used_str <= 9)
 		damage = max(damage - (damage * ((10 - used_str) * 0.1)), 1)
 


Lyroy:
> I don't need to argue this. Why would I have to argue why roundstart 80 damage per punch, also amplified by gloves, is a bad thing?

I don't think there's any class starting a round with 20 STR to begin with.
If we want to make the calculations based on what those classes will usually have on roundstart (14 to 16 STR), then we get the numbers of:

30.8 damage at 14 STR, 44.8 damage at 16 STR
Or, in the best case scenario of them using plate gauntlets, 37 damage and 53.8 damage respectively.

This is in comparison to, for example, a steel arming sword, wielded and using the cut intent at 14 STR, dealing 42 damage. Or a longsword under the same conditions dealing 46.2 damage. Both of these have the added ability of being versatile weapons that can stab, **peel, chop,** and so on, things fists cannot do. And let's not bring up rend weapons!

Fists cannot delimb, they need to bruteforce their way to destroy armor instead of being able to just remove it with peeling, they cannot bleed enemies out (which makes them almost useless against anyone with crit resist), can't Guard, etc.

Do the numbers look scary when you somehow get 20 STR? Absolutely. How do you get 20 STR reliably? No idea. I just think that essentially killing monk classes like this is absurd.

josephpatrickgreen:
```
/mob/living/carbon/human/get_punch_dmg()

    var/damage
    if(STASTR > UNARMED_DAMAGE_DEFAULT || STASTR < 10)
        damage = STASTR
    else
        damage = UNARMED_DAMAGE_DEFAULT

    var/used_str = STASTR

    var/obj/G = get_item_by_slot(SLOT_GLOVES)
    if(domhand)
        used_str = get_str_arms(used_hand)
    if(used_str >= 17) //edge case for 17+ gamer STR
        damage = (used_str == 17 ? max(damage + (damage * ((used_str - 10) * 0.25)), 1) : max(damage + (damage * ((used_str - 10) * 0.225)), 1))
    else if(used_str >= 11)
        damage = max(damage + (damage * ((used_str - 10) * 0.3)), 1)
    else if(used_str <= 9)
        damage = max(damage - (damage * ((10 - used_str) * 0.1)), 1)

    if(istype(G, /obj/item/clothing/gloves/roguetown))
        var/obj/item/clothing/gloves/roguetown/GL = G
        damage = (damage * GL.unarmed_bonus)

    if(mind)
        if(mind.has_antag_datum(/datum/antagonist/werewolf))
            return 50

    return damage
```
    
image

nerfs edge cases, not everyone else, whoops, pic is wrong, nerf starts at 17

Dedwolf:
This doesn't go far enough, players can spam unarmed attacks at a rate of about 3 every second, even while on defend intent you cannot parry them, it makes no sense whatsoever for unarmed attacks to be that fast to the point people with regular weapons cannot compete

josephpatrickgreen:
No no no, do not put a for loop on each damage calc, are you nuts?

josephpatrickgreen:
Once again, the objective was to nerf the edge case of extremely high str. Not lower everyone's unarmed by 40%. This pr is shit.

CVSthePharmacy:
> > homeboy you take MUCH more stamloss if you miss than a sword................ or even TRY to attack, you lose 10 on hit and 8 on miss on grabs. And with punches? 4 if you miss. Swords? 1 if you miss. You lose moar, ur argument is _ahem_ DESTROYED!!!!
> > On another point, rapier meta absolutely shreds against fists. just play differently and aim at other spots, rn this is mega lolz. I know you have your opinions and such but yeah this is like saying moondust is op because it removes your click cd with maces. Ogey try it out <3
> 
> You're flat out wrong, you don't lose 10 when you hit, you lose 1 every time you swing and an additional 4 if you miss. Swords are always 1. Your argument sucks and you sound like a fucking idiot. I'm telling you that swift intent doesn't affect the stamina cost of PUNCHING. See EXTRA_STAMDRAIN_SWIFSTRONG 10 which means every attack, no matter the weapon, on swift intent is 10.
> 
> But for punches? image image
> 
> Swift intent has no effect on punches. Sorry your meta is getting nerfed, you're so fuuuuucking upset i can see it through your writing.

Erm achsually yroue wrong bc attacking also means grabbing. Which is what punch combat is centered around. 10 on hit and 8 on miss. Pwned!!!!

josephpatrickgreen:
image

you could have achieved the same effect by lowering someone's strength by 5 points.

PR #195, opened 2025-10-24
makes mimi the fox less annoying

Jamdrawers:
## About The Pull Request

Sets Mimi the fox's (the one in the guild smithy)'s density to 0 meaning they don't block squares

## Testing Evidence

One line variable change that works for the Merchant's cat

## Why It's Good For The Game

Heard complaints about the fox getting in the way, and they seem small enough that it'd make sense for you to be able to step over them


Elli-Skala:
Biggest issue is more they'll try to kill you if you grab them ngl

PR #197, opened 2025-10-24
I HATE WARDEN AXE PICK!!!!!!!!!!!!!!!!!!!!!

CVSthePharmacy:
## About The Pull Request

Changes the pick intent to be a stab

## Testing Evidence

text change

## Why It's Good For The Game

Currently it's kind of like.... the best weapon for axe users in the game. This should diversify stuff without being the meta pick. As of rn the pulaski axe sux for picking into people so I see this as a good change.

### DIFF
--- a/code/game/objects/items/rogueweapons/melee/axes.dm
+++ b/code/game/objects/items/rogueweapons/melee/axes.dm
@@ -274,8 +274,8 @@
 	icon_state = "wardenpax"
 	force = 22
 	force_wielded = 28
-	possible_item_intents = list(/datum/intent/axe/cut,/datum/intent/axe/chop, /datum/intent/mace/warhammer/pick, /datum/intent/axe/bash)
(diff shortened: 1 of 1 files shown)


Jamdrawers:
I wouldn't necessarily be against just giving the special pickaxe-axe to another class and redesignating wardens as not being The Axe Specialists but IDK if that'd fly. I've been playing Warden a lot and if anything I'd kinda sorta want to nerf it in some ways - but buff in others; it's not about absolute power overall, it's about where it's placed. The design, how it comes together.

I do feel like being a Pick is kinda integral to the design of the axe.  Steel warhammers have stab - but that's because they have a long thin pointy bit at the end you can stab with! Not a big chunky hook on the back you have to swing! Would make it feel very unintuitive.

RadiantFlash7:
I think we can keep the axe as is presently. As pointed out, it's effectively a warhammer version of an axe, having a pick option to it. 'pologies for getting to this late, was hoping another maint would.

PR #207, opened 2025-10-26
Peeling Inaccuracy

ProbablyCarl:
## About The Pull Request
Peeling has a +25 to hit, right now.
This swaps that to a -10. Simple as.

We were discussing behind the scenes a way to make weapons more viable, for those that don't have access to peeling. Given right now, a sabre outperforms most weapons because of peeling being a wild thing when combined with certain intents. I will not sling shade at my beloved partizan. Leave it alone.

This may well be a way to correct the issue. Is it good? I 'unno. Peeling is weird and awful though, but we're stuck with it. May as well make the best of a bad thing.

Doubly so without gutting existing weaponry. Ironic as it is to say, given recent PRs.

## Testing Evidence
Simple number tweak. Another coal Carl PR. Slop. Not tested. Objectively awful. Whatever.

## Why It's Good For The Game
Find man who doesn't have peel weapon, while wearing full armour. Watch as you peel and delimb him / poke out his eyes in seconds, while he struggles to do much of anything. Assuming he's not smart enough to grapple.
Does this fix that? No. It helps with it being so obscene, though. I'd hope.

### DIFF
--- a/code/modules/mob/living/combat/accuracy_checks.dm
+++ b/code/modules/mob/living/combat/accuracy_checks.dm
@@ -1,5 +1,5 @@
-#define ULTRA_PRECISE_ZONE 1 
-#define PRECISE_ZONE 2 
+#define ULTRA_PRECISE_ZONE 1
+#define PRECISE_ZONE 2
 #define NO_PENALTY_ZONE 3
 #define RANGED_MAX_ULTRA_PRECISE_HIT_CHANCE 50 // No matter what max 50% chance to hit
 #define RANGED_ULTRA_PRECISE_HIT_PENALTY -25 // -25 for you - THEN we clamp.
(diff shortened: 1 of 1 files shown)


kepesk1:
Gem. I've been peeling legs and eyes for days and winning nonstop. Peels needs a nerf, I really don't care what avenue it comes in under. If I had it my way I'd remove peeling altogether for any weapons that aren't axe-bladed or picks. 

Peeling is too meta, and of course the meta will change if this gets accepted but it's just no bueno for any armored individuals (like knights). Armor peeling just isn't a typically normal concept in fighting anyhow, so I'm happy if it gets a boot. Back to beating the shit out of each other and using armor pen weapons instead of peeling the eyes off to gouge out both.
~Ala

PR #209, opened 2025-10-26
Peel Annihilation

ProbablyCarl:
## About The Pull Request
Removes all current instances of peel, outright, as used on weaponry. This doesn't remove the mechanic code wise, or the intents themselves, as they still exist if we wish to use it in the future. Should someone wish to justify the return. As messy as it is, for it be be slinking about.

For not-so-great compensation, it provides striking with a blade 10% more integrity damage, and mordhau bashing 30% more.
The Martyr's Sword, and, by extension, the See Blade, have been given a two-handed strike as well.
## Testing Evidence
Slop PR. Untested. Whatever.
It's simple line changes and should function without issue. Despite this being a massive wide-sweeping change.
## Why It's Good For The Game
Peeling makes entire weapon classes pointless. Outright.
Not only does it destroy certain weapons as an option, but it also completely invalidates armour.

Is this a good idea? Probably not. This PR goes up to argue implementation and balance, or if we even want it altogether.

Something considered was just reintroducing rondels and the like, to allow grapple pen, but also allowing them to peel armour. Which would mean knights and brigands have a reason to engage with their sidearm.

### DIFF
--- a/code/game/objects/items/rogueweapons/melee/axes.dm
+++ b/code/game/objects/items/rogueweapons/melee/axes.dm
@@ -133,7 +133,7 @@
 /obj/item/rogueweapon/stoneaxe/battle
 	force = 25
 	force_wielded = 30
-	possible_item_intents = list(/datum/intent/axe/cut/battle, /datum/intent/axe/chop/battle, /datum/intent/axe/bash, /datum/intent/sword/peel)
+	possible_item_intents = list(/datum/intent/axe/cut/battle, /datum/intent/axe/chop/battle, /datum/intent/axe/bash)
 	wlength = WLENGTH_LONG		//It's a big battle-axe.
 	name = "battle axe"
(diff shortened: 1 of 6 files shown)


kepesk1:
Huge. Ultra meta annihilation dub. RIP L Bozo for all peel fiends myself included. Gem please merge.

josephpatrickgreen:
@Useroth I recommend this PR.

CVSthePharmacy:
> I am very skeptical of a flat removal of peel. It would put most swords back into the realm of being less useful then other weapons again. While peel is certainly too good at present, We'll try tweaking things like peel's accuracy first.

Nobody wants peel. Sabres and rapiers are swords. They're not underpowered, their DPS is insane. I am VERY sketchy when seeing all the other weapons and their DPSs and you're saying they're going to the least useful weapon class. Hmmmmm

PR #211, opened 2025-10-26
Gives the Inn Cook a cleaver on spawn

Mryourface13:
## About The Pull Request
Gives the innkeep cook a cleaver on spawn, just like every other role. Just in case you latespawn and someone stole it.

## Testing Evidence

image

## Why It's Good For The Game

Fixes a possible annoyance of having to "acquire" a cleaver as a latejoiner. Every other role has this on spawn.


wow50000:
what about the cleaver on the table in the inn?

PR #219, opened 2025-10-28
Dualwielding Zeybek

RadiantFlash7:
## About The Pull Request
Adds the dualwielding trait to Zeybeck mercenaries

## Why It's Good For The Game
It's fitting given they're given two identical swords, only no way to use both at once.
If any other roles should have the dual wielding trait, feel free to let me know.

It's entirely possible I just like the role and want it buffed, so I'd prefer others vibe check me on this. It seemed fitting though.


RadiantFlash7:
> Could you also fix the dagger + whip zeybek loadout? The lhand gets replaced further down so they never actually get a dagger.

Can do.

PR #226, opened 2025-10-29
Buffs Orison water healing a bit.

Guidesu:
## About The Pull Request
Buffs holy and unholy water to heals a bit more than 0.1 per tick. Not better than normal health potions at all but it SHOULD heal now roughly 2~ per two ticks.

## Testing Evidence
I tested but eh... too lazy. Its just line edit.

## Why It's Good For The Game
Makes holy water less useful.

### DIFF
--- a/modular_azurepeak/code/modules/spells/orison.dm
+++ b/modular_azurepeak/code/modules/spells/orison.dm
@@ -208,21 +208,21 @@
 
 /datum/reagent/water/blessed
 	name = "blessed water"
-	description = "A gift of Devotion. Very slightly heals wounds."
+	description = "A gift of Devotion. Heals the body from within, but not physical wounds."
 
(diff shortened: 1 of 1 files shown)


DelineFortune:
You can get unlimited healing from a fireplace, you know it, right?

AisuDracov:
If you want an idea to buff orison, could always have Holy water just restore blood/remove oxy damage faster compared to regular water. Not sure how that part is coded, but an idea at least.

That way, there'll be reason to use holy water over regular water as a stabilizer rather than a healing method.

Also- as for why most servers dont just have the church do alchemy off the bat with the means to do so, its to help try and give Church and the Apothecaries separate identities. If the church is capable of pumping out potions like the Apothecary, why bother being an Apothecary anymore? People already prioritize taking the injured and dead to the church, since they can act faster on stabilizing the dying or reviving them.

PR #231, opened 2025-10-30
Fencer cuirass more integrity

Heheniy:
## About The Pull Request
Fencer cuirass integrity 180 -> 300 (in line with other cuirasses)

## Testing Evidence
varedit

## Why It's Good For The Game
This is literally the most expensive light chest armor, yet it has only slightly more then half integrity of its analogues which is unfair

### DIFF
--- a/code/__DEFINES/roguetown.dm
+++ b/code/__DEFINES/roguetown.dm
@@ -48,7 +48,7 @@
 #define ARMOR_INT_CHEST_LIGHT_MASTER 300 // High tier cloth / leather armor
 #define ARMOR_INT_CHEST_LIGHT_MEDIUM 250 // Medium tier cloth / leather armor
 #define ARMOR_INT_CHEST_LIGHT_BASE 200
-#define ARMOR_INT_CHEST_LIGHT_STEEL 180
+#define ARMOR_INT_CHEST_LIGHT_STEEL 300
 #define ARMOR_INT_CHEST_CIVILIAN 100
 
 // LEG PIECES - Leg Armor


CloverIsLucky (changes requested):
Leave the code define in peace. Use the high-tier light armor's integrity instead, or the medium armor integrity's if what you look is to make the fencer's cuirass to be on par with medium armors' integrity.

Go to the Fencer's Cuirass code and write another replacement armor integrity define: 
`max_integrity = ARMOR_INT_CHEST_LIGHT_MASTER`

CloverIsLucky:
Huh, how strange. That armor define( ARMOR_INT_CHEST_LIGHT_STEEL) looks so redundant it only exists for the Fencer's Cuirass.

Jamdrawers:
> Huh, how strange. That armor define( ARMOR_INT_CHEST_LIGHT_STEEL) looks so redundant it only exists for the Fencer's Cuirass.

Guess the Fencer's Cuirass is the only piece of armour like that (so far)

Ruediger4:
> I would suspect that it's intentional that the only LIGHT piece of metal chest armour has less integrity than heavier chest pieces as a balancing point - what you pay for in exchange for making it light.
> 
> IDK, maybe 180 is too harsh a difference, but it shouldn't be _in line_ with a heavy chest piece.

Exactly this

The fencer's cuirass is the ONLY light armor piece with plate-tier defense. It is arguably the best chest piece for dodgejaks because it is plate defense and not brigandine or leather.

This **SHOULDNT** have high integrity. This is a piece that's expensive because it is plate protection for a class that is, likely, either A) a Freifetcher so they rely on weapon skill and parry, or B) a dodge build class which does NOT need a high integrity chest piece with plate defense to make up for the fact they'll eventually get hit.

This is already balanced, dont give this higher integrity
> This is literally the most expensive light chest armor, yet it has only slightly more then half integrity of its analogues which is unfair

It's the Most Expensive because it is Plate Protection For Light Builds. It's VERY fair.

CVSthePharmacy:
Just make it 202, chainmaille integrity.

PR #232, opened 2025-10-30
Gives metal skirts leg protection

Heheniy:
## About The Pull Request
Chain and plate skirts protects not only groin but legs too

## Testing Evidence
varedit

## Why It's Good For The Game
Skirts are long enough and even have in their descrtiption that they are covering thighs, making it logical that they should protect legs
This will make them the only metal variant of light leg armor


Jamdrawers:
Surely we have the chain kilt for that?

Heheniy:
it is medium armor

PR #233, opened 2025-10-30
Smithing items quality affects max integrity

Heheniy:
## About The Pull Request
Same as for sellprice, quality of item made by smith affects its max integrity too (only for items that are worse then "standart")

Kinda experimental, lets see how it will go

## Testing Evidence
varedit

## Why It's Good For The Game
Item quality has more impact

### DIFF
--- a/code/modules/roguetown/roguejobs/blacksmith/smith_components.dm
+++ b/code/modules/roguetown/roguejobs/blacksmith/smith_components.dm
@@ -255,6 +255,9 @@
 					I.polished = 4
 					I.AddComponent(/datum/component/metal_glint)
 
+		if(modifier < 1)
+			I.max_integrity *= modifier
+
 		I.sellprice *= modifier
 		if(istype(I, /obj/item/lockpick))
 			var/obj/item/lockpick/L = I


Purpleturtle3:
It only should go downwards, armor doesn't need to have higher integrity. It would just make peel better.

Ruediger4:
This will fuck over:
- Bandits
- Any role with roundstart gear if they cannot self-smith or if there's no smith
- Wretches if there's no wretch smith dedicating all round to leveling up smithing
- Migrant waves
- Literally everyone who may not have access to a smith

Make it so that **SHITTY QUALITY ARMOR** has **SHIT INTEGRITY**, but anything of regular quality or above has the expected quality. This is bad otherwise. Awful.

RadiantFlash7 (changes requested):
I'm not going to close this off the bat, But as is, we don't want to increase max integrity from good smithing that opens a horrendous can of worms we had to deal with on RW1. I can, however, see bad smiths having poor quality armor as a potential shift if thats something you want to shift this PR to cover only.

Otherwise, the PR can be closed if you'd rather not do that.

PR #235, opened 2025-10-30
Sets the black market navigator's tax to 50% instead of 70%

Jamdrawers:
## About The Pull Request

As per title

## Testing Evidence

image
Text displays.
image
place an item worth 42
image
Turns to... 21 coins!
Tried again, worked.

## Why It's Good For The Game

The cut is more than enough to encourage merchant use as it is I think. We had it somewhere around this in RW1 and I think that was fine. At 70% it feels overly punishing for something you have to really go out of your way to find.


Jamdrawers:
Works!

PR #236, opened 2025-10-30
Peel tweak attempt 2 and peel related code fixes from upstream.

RadiantFlash7:
## About The Pull Request
This PR ports some fixes and changes from upstream on azure in relation to peel: https://github.com/GeneralPantsuIsBadAtCoding/Azure-Peak/pull/4010
https://github.com/GeneralPantsuIsBadAtCoding/Azure-Peak/pull/3701/files
Predominantly, 

- Peel messages are now consistent and wont show the white text when no peeling can be done
- Added a new sound for striking bad peel areas and a message

- fixes runtimes for impacts to Ears & Skull specifically. This also makes Ears be a part of "FACE" coverage. 

-----

Finally, we've given peel intent  a larger click cooldown.

## Testing Evidence
image

## Why It's Good For The Game
Peel still has issues with making swords the best weapon in the game. This PR is one of three options I see potentially as a solution that I'd like to have tested.

Current option 1 implemented in this PR: Giving Peel has a slower click cooldown to allow people a better chance at baiting someone peeling as a counterplay, potentially even removing peel from swift balanced weapons so people can't be peeled at highspeed.

Future option 2 not included in this PR: Peeling  reduces an armor value by 50-75% on success instead of removing coverage, allowing swords to do chip damage through armor, but not outright letting hard crits  like arteries go through.

Future option 3 not included in this PR: Full peel removal. Figuring out another method that keeps swords still usable.


RadiantFlash7:
> * Giving pointed swords a thrust intent that acts exactly as spear stab does (Bleeding through armor) but carries the same kind of commitment (hit delay) that pick on a dagger has

This has potential, but at that point feels a bit too much like it's polearm infringing to a degree.

> * Replacing the concept of peel altogether with something that scales durability loss on armor instead of allowing you instant perfect access. I.E. New (replacement) intent with low durability damage, but with the ability to apply a "stack" on every hit, making that intent do increasingly more damage to the armor piece until otherwise interrupted at which point it resets or hits the maximum durability damage scaling (x2? x3? Increments of .3 per hit starting from .5? Increments based on sword type?)

This is similar to my second idea I was planning on trying after this. Where my idea was to have successful peeling reduce the armor protection of the armor, degrading it, rather then elimating it complete. I'm keen to keep combat mechanics that have give and push with interaction in other forms, such as being able to counter it by doing one thing or another.

> * Make swords generally the sturdiest weapons durability wise in exchange for their poor armored combat use?

That is an option if all else fails, but not a particularly interesting one.

Edit: It might be worth adjusting it so certain armors aren't peelable. Such as chain maille.

PR #237, opened 2025-10-30
Unarmed Diminishing Returns Fix

josephpatrickgreen:
## About The Pull Request
-Fixes the near global unarmed nerf so that it only effects higher STR
-Fixes the diminishing returns so that they are actually diminishing returns instead of a small flat increase

See the below graph where old damage was exponential and new damage is weird and lumpy ending in a 1.5 flat damage increase per STR, the proposed is a diminishing damage curve ending at the damage of an iron sword at 20 STR.

All numbers in the graph are plate gloves against armor so 20% higher than no gloves at all.
## Testing Evidence
Number tweak
image

## Why It's Good For The Game
-People with 11 STR didn't need a nerf.
-Makes the numerous unarmed subclasses playable again

### DIFF
--- a/code/modules/mob/living/carbon/human/human_helpers.dm
+++ b/code/modules/mob/living/carbon/human/human_helpers.dm
@@ -130,27 +130,18 @@
 	if(domhand)
 		used_str = get_str_arms(used_hand)
 
-	if(used_str >= 11)
-		var/bonus = 0
-
-		if(used_str <= 14)
-			// Normal scaling: +20% per point over 10
-			bonus = (used_str - 10) * 0.2
-		else
-			// Diminishing returns after 14
-			// Start with the full +0.8 from 14 STR
-			bonus = 0.8
-			var/extra = used_str - 14
-			// Each point beyond 14 gives a smaller bonus than the last:
-			// e.g., +0.1, +0.075, +0.05625, etc.
-			var/next_bonus = 0.1
-			for(var/i = 1, i <= extra, i++)
-				bonus += next_bonus
-				next_bonus *= 0.75 // reduces 25% each time
-		damage = max(damage + (damage * bonus), 1)
-
 	if(used_str <= 9)
 		damage = max(damage - (damage * ((10 - used_str) * 0.1)), 1)
+	else if(used_str >= 11)
+		if(used_str >= 15) //15 and up
+			var/dim_returns = 0.3
+			var/extra = used_str - 14
+			dim_returns = dim_returns - (extra * 0.025)
+			if (dim_returns < 0.125)
+				dim_returns = 0.125
+			damage = max(damage + (damage * ((used_str - 10) * dim_returns)), 1)
+		else // 11 to 14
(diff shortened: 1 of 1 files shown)


RadiantFlash7:
### I wrote up an entire thing on my opinion on this last night, and promptly forgot to hit the review button; But that was probably too long to begin with. I'll try and simply my thoughts on this change:

----

The issue with the way you're looking at the graph is one of perspective.  We're not just considering damage unarmed does to armor, but also to unarmored injuries. The higher the damage against armor goes, the more it compounds when it's against unarmored areas, because as you've noted prior, it's halved.

This means that while with your proposal of 20 damage at str 14, You're actually doing 40 to unarmored areas. That of course keeps extending where 30 damage to armor at str 20, is doing 60 damage a punch. This is something you understand without a doubt- but I need to emphasize that is where the problem comes in.

The ideal placement of balance in our codebase for punches to be actively worse then knuckles, katars, punch daggers, and any future unarmed weapons when it comes to dealing with specifically armor, but it's better against unarmored targets- something  both the present calculation and your proposed one do fairly well. 

Specifically, we want to strive to have it so punching an unarmored area is 'on par' with other general (non-unarmed skilled) weapon damage, because punching is extremely stamina efficient, and is practically free to do on swift intent. This is something that becomes extremely overt and even opressive when grappling comes into play, as anyone thats built for unarmed is also built for wrestling. All it takes is a grab and then rapid swift punching to create some of the highest damage output per stamina usage in the game.


To that end, I was specifically wanting Carl to limit the amount of damage that can be done with punches in general, with diminishing returns after str 14 so both so unarmed weapons have a place against armor, and so that we're not winding up with damage much over 50 per punch to unarmored areas no matter the strength.

----

All that said, I can certainly say the math calculation we have now is awkward looking on a graph., and certainly we could wind up with a proper diminishing curve like your proposed calculation is, but the goal remains to keep the unarmored damage on par with weapons as opposed to the armored one.

josephpatrickgreen:
> Specifically, we want to strive to have it so punching an unarmored area is 'on par' with other general (non-unarmed skilled) weapon damage, because punching is extremely stamina efficient, and is practically free to do on swift intent. This is something that becomes extremely overt and even opressive when grappling comes into play, as anyone thats built for unarmed is also built for wrestling. All it takes is a grab and then rapid swift punching to create some of the highest damage output per stamina usage in the game.
> 
> To that end, I was specifically wanting Carl to limit the amount of damage that can be done with punches in general, with diminishing returns after str 14 so both so unarmed weapons have a place against armor, and so that we're not winding up with damage much over 50 per punch to unarmored areas no matter the strength.
> 
> All that said, I can certainly say the math calculation we have now is awkward looking on a graph., and certainly we could wind up with a proper diminishing curve like your proposed calculation is, but the goal remains to keep the unarmored damage on par with weapons as opposed to the armored one.

image

I've added katar and knuckles to the chart along with the minimum and maximum damage of unarmed before, after and proposed for informational purposes, I'm looking at the graph and just determining a response with the info, I'll append my thoughts in this message...

PR #238, opened 2025-10-31
Non-Tossblades in Boots

ProbablyCarl:
## About The Pull Request
Allows players to shove both knives and lockpicks into their boot. You can have one of each.
You could do this with tossblades, but given how fragile they were, it wasn't ideal.
Nor do most folks know about that, seemingly. This should open it up to more use.

You may MMB to remove the lockpicks. RMB is still used for knives.
You cannot store lockpick rings, to be clear. Just a singular lockpick.

Drawing either knife or lockpick takes two seconds, while also presenting a message to yourself and anyone nearby.
## Testing Evidence
image

## Why It's Good For The Game
It probably isn't.
Tossblades were the item you'd stuff into a boot.
This makes sheathes pointless in almost every case, for knives.

However, it's _cool_, and gives another use for boots. I've had times elsewhere, where I've gutted a captor because they forgot to check my boots. Not that you can't do that already with tossblades. Perhaps now players will have another method of escape / fighting. If we humor this.

### DIFF
--- a/code/modules/clothing/rogueclothes/feet.dm
+++ b/code/modules/clothing/rogueclothes/feet.dm
@@ -58,19 +58,23 @@
 
 /obj/item/clothing/shoes/roguetown/boots/attack_right(mob/user)
 	if(holdingknife != null)
-		if(!user.get_active_held_item())
-			user.put_in_active_hand(holdingknife, user.active_hand_index)
-			holdingknife = null
-			playsound(loc, 'sound/foley/equip/swordsmall1.ogg')
-			return TRUE
+		user.visible_message(span_warning("[user] is drawing something from [src]!"), span_warning("I begin drawing a knife from [src]!"))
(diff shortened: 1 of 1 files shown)


Dedwolf:
i had no idea you could do this in the first place damn

Jamdrawers:
If you draw a knife from a boot does it at least have the delay from taking a knife out of a bag without a sheathe? Because I think that'd go a long way to making sheathes not worthless and not making dagger-boots something that EEEEVERY one will be doing ALL the time.

ProbablyCarl:
> If you draw a knife from a boot does it at least have the delay from taking a knife out of a bag without a sheathe? Because I think that'd go a long way to making sheathes not worthless and not making dagger-boots something that EEEEVERY one will be doing ALL the time.

It does not.
I'll add a delay.

PR #239, opened 2025-10-31
Re-adds RW1 scruffing, unshitcoded

Savotta:
## About The Pull Request

As per the title, adds scruffing. You can only scruff **ADULT** Tabaxi and Lupians. Middle-aged and Old aren't affected by this.

Their neck has to be exposed, and you have to be directly behind them to do so. This is for balancing reasons, since scruffing is meant to be more of a racial gimmick than an actually viable combat strat.

If you manage to scruff someone, they'll be paralysed for 2 seconds. The new SCRUFF icon will only show up when aggrograbbing the neck of a Tabaxi or Lupian.

## Testing Evidence

https://github.com/user-attachments/assets/c0bc6653-86a1-49f1-97a0-e6778d983634

## Why It's Good For The Game

This was present in RW1, but shitcoded since it was linked to the twist mechanic when aggrograbbing someone's neck. This would usually result in you snapping their neck alongside scruffing them. A simple improvement, and a funny little mechanic that probably won't see much use, same as it didn't on RW1.

Open to balancing considerations, but I think getting an aggrograb from behind on someone with an exposed neck is already difficult enough as is.

### DIFF
--- a/code/modules/mob/living/grabbing.dm
+++ b/code/modules/mob/living/grabbing.dm
@@ -268,7 +268,6 @@
 											span_userdanger("You go limp as your scruff is twisted!"), span_hear("I hear aggressive shuffling!"), COMBAT_MESSAGE_RANGE, user)
 									to_chat(user, span_warning("You twist [H]'s scruff, causing them to go limp!"))
 									log_combat(user, H, "scruffed")
-									user.changeNext_move(CLICK_CD_RESIST) // higher cooldown, 2 seconds
 								else
 									to_chat(user, span_warning("I need to be behind [H] to scruff them!"))
 							else
@@ -688,3 +687,4 @@
 	name = "scruff"
 	desc = ""
 	icon_state = "inscruff"
+	clickcd = CLICK_CD_RESIST


code/modules/mob/living/grabbing.dm
```diff
@@ -250,6 +250,31 @@
...
+									H.Paralyze(20) // 2 seconds of paralysis, might be a bit much...?
+									H.visible_message(span_danger("[user] grabs [H] by the scruff, causing them to go limp!"), \
+											span_userdanger("You go limp as your scruff is twisted!"), span_hear("I hear aggressive shuffling!"), COMBAT_MESSAGE_RANGE, user)
+									to_chat(user, span_warning("You twist [H]'s scruff, causing them to go limp!"))
+									log_combat(user, H, "scruffed")
+									user.changeNext_move(CLICK_CD_GRABBING) // cooldown
+								else
+									to_chat(user, span_warning("I need to be behind [H] to scruff them!"))
+							else
+								to_chat(user, span_warning("Scruffing doesn't work on non-adults!"))
```
RadiantFlash7: Perhaps have this be "Scuffing doesn't work on older adults."

RadiantFlash7 (changes requested):
Scruffing makes it's glorious return. Only a few minor tweaks needed before this can be TM'd.

PR #240, opened 2025-10-31
[AP Port] Critical Resistance Rework

ProbablyCarl:
## About The Pull Request
A near 1:1 of [4230](https://github.com/GeneralPantsuIsBadAtCoding/Azure-Peak/pull/4230), from AP.
It excludes the weak limb removal of constructs, since we never had easy dismember for them in the first place.
Additionally, it adds the intended alert type to the actual 'debuff'.

This is otherwise identical, excluding spacing.

See the PR above for more information.
## Testing Evidence
image
image
image
image

## Why It's Good For The Game
Crit resist is awful to fight.
You, quite literally, cannot deal with them in a matter that doesn't go to lethal.
Barb? Decap or bleed out.
Ruma? Decap or bleed out.
Berserker? Get your own skullcracked, and promptly get decapped or bleed out.
Using blunt in any capacity? They're invincible. Have fun.


josephpatrickgreen:
I like the stacks idea a lot.
I dunno about the damage increase though cause the reduction is normally survivability tool for classes that are slow, low int and lightly armored, meaning they are surely going to get hit. I would recommend getting rid of the damage increases but let's hear what a maint has to say.

Sirgir:
As much as I love running an Enduring/Muscular statpack Barbarian, it is understandably unfun to play against someone with critical resistance-- because if they truly desire it, they simply cannot be caught bar delimbing or bleeding them out. Considering most classes with crit resistance ALSO have pain immunity, your already limited number of options of taking someone with these traits down got even smaller due to certain criticals just being outright blocked and the fact they can just waltz out of combat because they don't keel over from pain.

So this is a good way to have it, limited critical blockage but not Truly Immune.

Jamdrawers:
Crit Resistance sounds much better than the Crit immunity it was before. Sounds very reasonable.

Elli-Skala:
Does this mean skeletons won't just instantly die anymore

Elli-Skala:
> > Does this mean skeletons won't just instantly die anymore
> 
> you might be thinking of Critical Weakness-- so no, skeletons are still quite vulnerable to skullcracks/rib smashing/nose destruction.

Saw "critical resistance stack npc" and misunderstood

PR #246, opened 2025-11-01
Mercenary respawn

Rayforx:
## About The Pull Request
Allows people to spawn in as a mercenary if they end up dying.

## Testing Evidence
N/A
Same code as the adventure.

## Why It's Good For The Game
It sucks that there are always 4 mercenaries busy in a private room, while 4 are dead in a corner of the map. Allowing the ability for people to respawn when it's high-pop would make a lot of people able to play their character.


Hmeister-real:
Absolutely not. Just play a new role, mercs are bloated af anyways.

PR #249, opened 2025-11-02
Buffs Mordhau.

CloverIsLucky:
## About The Pull Request

Adds Stab intent for the longsword family's alt-grip(mordhau)

## Testing Evidence

One code change(or two if counting the training sword so you don't disembowel with the longsword's sharp thrust.)

## Why It's Good For The Game

Stab, stab, stab~!

### DIFF
--- a/code/game/objects/items/rogueweapons/melee/swords.dm
+++ b/code/game/objects/items/rogueweapons/melee/swords.dm
@@ -257,7 +257,7 @@
 	force_wielded = 30
 	possible_item_intents = list(/datum/intent/sword/cut, /datum/intent/sword/thrust/long, /datum/intent/sword/strike, /datum/intent/sword/peel)
 	gripped_intents = list(/datum/intent/sword/cut, /datum/intent/sword/thrust/long, /datum/intent/sword/peel, /datum/intent/sword/chop)
-	alt_intents = list(/datum/intent/effect/daze, /datum/intent/sword/strike, /datum/intent/sword/bash)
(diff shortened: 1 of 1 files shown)


anconfuzedrock:
ok but actually why

Sirgir:
> ok but actually why

I suppose it's to represent half-swording?

Draganfrukts:
> > ok but actually why
> 
> I suppose it's to represent half-swording?

image
Which makes absolutely zero sense  because they're two completely different techniques....



Mordhau is meant to make it so you can give away all your blade-related intents for mace intents. This just makes it so nobody ever wields a great or longsword normally again unless they want to use peel.

anconfuzedrock:
unrelated to what the pr does in practice, but if your goal is to buff mordhau (which I'd approve of) you need to fix it being a 1 handed technique. it's really unintuitive and stops it from being very effective- it' s only good if you were using longsword with shield, otherwise you're better off using it two handed for more damage. and with greatswords it's entirely pointless since one handed does half damage.

RadiantFlash7:
Unfortunately, Balance Tweak PRs are being closed for the time being until the AP Parity PR is wrapped up and merged, due to the refractor done to each class and a great deal else. If you'd like, You can open a new PR after the merge goes through.

CloverIsLucky:
> > > ok but actually why
> > 
> > 
> > I suppose it's to represent half-swording?
> 
> image Which makes absolutely zero sense because they're two completely different techniques....
> 
> Mordhau is meant to make it so you can give away all your blade-related intents for mace intents. This just makes it so nobody ever wields a great or longsword normally again unless they want to use peel.

If you wanna act nerdy with me, how about if I add pick intent instead, huh? So it accurately looks like you can now pierce people's skulls with the handguard.

PR #258, opened 2025-11-08
Garrote Buff - Absolver Rites - Slurbow Rarity

ProbablyCarl:
## About The Pull Request
Does as it says on the title. A number of changes facing Inquisition content.

Firstly, garrote.
These required EIGHT resists to snap. You would never actually knock someone out with this, as it has 200 integrity, and takes 25 damage per resist. On escape, it immediately broke the wire. I've replaced this with removal of the integrity damage on resist, with escapes dealing a flat 100 integrity damage. Two escapes will snap it. I hope that it adds some actual use to an otherwise typically ignored item, given it's only used currently for stopping chases and muting mages, as opposed to the actual functionality intended for it. Especially since wrestling 'balance' as is make these even more painful to use.

Skipping ahead, slurbows.
I've removed them from the merchant vendor, fullstop. They had an identical price range to crossbows. For some reason. Wow. Wonder why. Additionally, knaves get it at an identical price setup to the longbow, making it a direct upgrade of the crossbow. Because it is.

Now, to actual content. Absolver rites.
This takes the rite circle stuff, the existing non-functional rite circle for Psydonites and actually turns it into something with content. Three rites exist. 'Conversion', 'Admonishment' and 'Freedom'. 'Admonishment' and 'Freedom' provide a five minute cooldown. 'Conversion' can be used at any time, so long as that cooldown isn't in effect. Only Absolver (Ritualists with the Inquisition trait) can actually use these rites.
 - 'Conversion': Allows any non-Psydonites to be converted to Psydon worship. It functions identically to 'Inhumen' conversion, with the exception being that it doesn't actually do anything beyond conversion. You get no skills or boons. It's just a matter of faith swapping, with your devotion taken into account as with others.
 - 'Admonishment': Quicksilver poultice via ritual. That's it. It's that simple. There's nothing more to this. Once every five minutes, to allow for mass-deconversion.
 - 'Freedom': Fullstrips anyone that this is used on. Can only be used on overt heretics, which are those in rite armour. It also removes the trait, which once more allows them to enter hallowed ground without issue. Pair with conversion for best results. As with 'Admonishment', this is on a five minute timer.

## Testing Evidence
image
image
image
image
image
image

## Why It's Good For The Game
Garrotes are boring. They don't work for their intended function. Is it a good change? I 'unno. Is it balanced? Probably not. But they're so rarely used as is, that I don't think I've seen anyone else use them, and the one time I had it on me, I'd immediately snapped it and sprinted away without issue.

Slurbows are a meta weapon. They just are. They shouldn't be so stupidly easy to get.

Absolver can do with more content, and ways to handle antags they otherwise can't deal with. This PR will allow them to ACTUALLY CONVERT PEOPLE and deconvert antags when poultice counterplay abuse has taken place. Also just otherwise removing rite armour from folks for whatever reason they'd need.

### DIFF
--- a/code/datums/status_effects/rogue/debuff.dm
+++ b/code/datums/status_effects/rogue/debuff.dm
@@ -290,7 +290,6 @@
 	desc = "I can barely feel my limbs!"
 	icon_state = "chilled"
 
-
 /datum/status_effect/debuff/ritesexpended
 	id = "ritesexpended"
 	alert_type = /atom/movable/screen/alert/status_effect/debuff/ritesexpended
@@ -309,6 +308,16 @@
 /atom/movable/screen/alert/status_effect/debuff/ritesexpended_heavy
 	name = "Rites Complete"
 	desc = "It will take a lot of time before I can perform a next rite. I am drained."
+	icon_state = "ritesexpended"
+
+/datum/status_effect/debuff/ritesexpended_lesser
+	id = "ritesexpended_lesser"
+	alert_type = /atom/movable/screen/alert/status_effect/debuff/ritesexpended_lesser
+	duration = 5 MINUTES
+
+/atom/movable/screen/alert/status_effect/debuff/ritesexpended_lesser
+	name = "Rites Complete"
+	desc = "It will be a short period before I can perform another rite."
 	icon_state = "ritesexpended"
 
 /datum/status_effect/debuff/call_to_arms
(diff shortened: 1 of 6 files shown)


kepesk1:
garroting you asap

CVSthePharmacy:
I fw this but what if we made the garrotes 300 integrity? Figure the instant breakout from Iconoclast against a confessor is FUNNY and all- but also kinda shifty if you don't have 4 extra cordages (not to capture but to use effectively in a way where your team can frag them b4 they break out)

Jamdrawers:
Getting these back soon I hope?

PR #260, opened 2025-11-10
Fixes bathmaster spawns

Jamdrawers:
## About The Pull Request

I think the issue was with the namechange not carrying over, and everyone just not noticing because they usually latejoin. Should stop bathmasters from spawning in the bandit camp.

## Testing Evidence

Loaded into a round without this change, spawned in a towner spawnpoint in the tavern. Loaded in with the change and spawned in the bathhouse.

## Why It's Good For The Game

Fixes a bug


ProbablyCarl:
Superseded by #330.

PR #261, opened 2025-11-10
Makes the SCOM sounds just a little bit quieter

Jamdrawers:
## About The Pull Request

Adjusts Garrisonscom.ogg and scom.ogg in audacity to be just a little bit quieter and lower pitched.

## Testing Evidence

Hard to screenshot a sound

## Why It's Good For The Game

It was kinda painful on the ears having a houndstone before, the notification is really harsh on the ear.


Jamdrawers:
Think I fixed the conflict

PR #274, opened 2025-11-11
Allows grabs to silence through masks again

Dedwolf:
## About The Pull Request

Direct revert of https://github.com/GeneralPantsuIsBadAtCoding/Azure-Peak/pull/3294 minus their coverage flag changes.

## Testing Evidence

One line change.

## Why It's Good For The Game

The same reasons used in the Azure Peak PR can be used for why this shouldn't be a thing. Players will often yell immediately when grabbed, without a way to silence those players, sneakily breaking into high security areas isn't feasible, nor is safely kidnapping combat roles without having to escalate to all out combat and either player getting skullcracked, and even then the player who isn't an antagonist can just scream for help and ruin the antagonists plans. 

This is not a good thing, players should not be punished for getting the drop on someone by making the dreaded combat roles even more resilient. You call it a 'grab meta' but the alternative is smashing your skull with a mace. 

As for spellcasters, they put out far too much DPS and have many ways to keep players at bay, they can continue to cast so long as they may speak, even when downed, even when stuttering in pain, covering their mouth as a means to shut them down is not often abused in the first place. I've seen plenty of frag clips from magic user's perspectives.

Besides, you can very reasonably slip your hand up under someone's visor to cover their mouth, same goes for steel masks. Strangling someone who is wearing dedicated neck armor, which is rigid, is far less believable so that has not been changed.

### DIFF
--- a/code/modules/mob/living/carbon/carbon.dm
+++ b/code/modules/mob/living/carbon/carbon.dm
@@ -1363,7 +1363,7 @@
 	if(mouth?.muteinmouth)
 		return FALSE
 	for(var/obj/item/grabbing/grab in grabbedby)
-		if((grab.sublimb_grabbed == BODY_ZONE_PRECISE_MOUTH) && (get_location_accessible(src, BODY_ZONE_PRECISE_MOUTH)))
+		if(grab.sublimb_grabbed == BODY_ZONE_PRECISE_MOUTH)
 			return FALSE
 	if(istype(loc, /turf/open/water) && !(mobility_flags & MOBILITY_STAND))
 		return FALSE


Sirgir:
Watched a bandit mage steamroll most of the server on lowpop using dragonhide, fireball, lightning bolt, spitfire, blink, and repel. No one could stop him because he wore a steel mask. Being totally immune to mouthgrabs is not a good thing I think, especially spellcasters who's one big counter is getting mouthgrabbed (this is why they have low wrestling!!!).

Murderfoxen:
No way, unless masks prevent this specifically. Ratwood code already makes casting when chained impossible, and thats silly as is. This doesnt make the game more interesting.

kepesk1:
> why are people so up in arms over reverting a change that came into effect 2-3 days ago without being explicitly documented?

I've been arguing that mouth grabs on mages should be somehow removed since RW started. We finally get it, and it's reverted 2 days later for an unrelated reason.

It's extremely frustrating. It's always been tied to kidnapping and restricting speaking rather than countering mages. This, since AP made spells weaker for mages to lower DPS output since mouthgrabbing was removed, makes this problem just a flat nerf for mages -- and sure we can fix it, but who? I don't really have time to sit down and code dive and neither do a lot of others.

anconfuzedrock:
If you don't know what I'm talking about read it again. Not everyone has weapon stats. seamsters have no wrestling skill.

of adventurer mage classes, 1 has novice wrestling and 2 have apprentice. 2 is what battlemaster adventurers get, it's what some full combat mercenaries (condottiero) get. It's what councillors get. 

Every combat mage class has at least apprentice wrestling, some like hedge mage get wrestling three. If it has low wrestling, maybe it's not a combat role? mage associates get novice, sure, but so do apothecaries. 

And not having knife skill is irrelevant if your issue is getting GRAPPLED. weapon skill only matters when dodging or parrying, and you can't do either while grabbing someone. you can stab your attacker even if they have full plate and you have no knife skill, because it won't miss and dagger pick intent  has 75 AP. 

also, people have two hands, yes. mages have two hands and a mouth. you can't grab both a mage's hands and their mouth. if you get disarmed use your other hand.

kepesk1:
> Edit: also noticed the personal attack, piss off. I play mages and I play non-combat roles, don't assume you know me.

Oh, I'm well aware you min-max and know the code inside and out for maximizing your stats, you don't need to brag about successfully beating people. I think your specific case is not reflective of the majority of players playing mages with a mixture of non-combat spells and combat spells. You play mostly mercenaries melee classes anyways.

It would be beneficial for the thread to stay on the topic of mouth-grabbing instead of armor penetration on knives, because it is not a realistic alternative for the average player being shut down from nearly all avenues of defense from a weak grab.

anconfuzedrock:
to clarify to others, stopped arguing since I realized this guy's an asshole. not even reading that. up for discussing this with anyone else... maybe those other combat classes need more wrestling.

CloverIsLucky:
> > Average Muggle realizes masks really protects mouth and is unpleasant about it. Dorohedoro mages let's go!!
> 
> I don't find using the correct channels to amend an issue to be 'unpleasant', I do however find calling people 'muggle' to be antithetical to constructive feedback. I don't know what 'dorohedoro' means either.

Jokes aside, here's my criticism: I think is best to fix some armor pieces that shouldn't give mouthgrab protection instead of directly touch mouth armor code. For example, the fullcoifs shouldn't give mouth protection when unhooded.

Pidgey-chan:
About time. Sick of mages blinking out of fucking grapples.

RadiantFlash7:
Riiight. I don't think we're going to be removing mouth coverings blocking silencing grabs frankly.
Mages are good but they're not glass cannons anymore. An instant shutdown to their entire class shouldn't effectively be a thing like it was on rw1.

Further more, this PR isn't even intentionally nerfing magea- thats kind of a side effect. Realistically, You probably shouldn't be mouth grabbing fully armored people to silence them tor abductions.

I get that makes banditing more difficult but we'll need another option. Someone suggested garrotes- That may be one. Iirc, the inquisition gets a form of one. Giving one to bandits may suffice in solving the same issue?

Pidgey-chan:
> Riiight. I don't think we're going to be removing mouth coverings blocking silencing grabs frankly. Mages are good but they're not glass cannons anymore. An instant shutdown to their entire class shouldn't effectively be a thing like it was on rw1.
> 
> Further more, this PR isn't even intentionally nerfing magea- thats kind of a side effect. Realistically, You probably shouldn't be mouth grabbing fully armored people to silence them tor abductions.
> 
> I get that makes banditing more difficult but we'll need another option. Someone suggested garrotes- That may be one. Iirc, the inquisition gets a form of one. Giving one to bandits may suffice in solving the same issue?

Are you kidding? You're seriously alright with letting mages teleport while grappled? Holy fuck, rogues have jack shit on wizards with regards to slipperyness. So mages are just immune to grappling with a mask on and can still teleport around. Another shitty Azure change we're stuck with, great.

PR #275, opened 2025-11-11
Fixes cooked cake base sprite conflict

RadiantFlash7:
## About The Pull Request
Cake sprite conflict fix- We have 2 types of cake sprites. this is troubling really.
This is actually kind of leaivng me uncertain which should be used. We have new iced cakes, but we also have the old cake that looks better with the old sprites. We'll be using the newer ones for now.

## Why It's Good For The Game
Conflict fix


Jamdrawers:
Still, think the old sprite ought to be the one used I think. Pretty is good. And other than the final sprite for the various new cakes all the previous steps are with the nicer artstyle. Would be jarring to make a cheesecake that switches artstyle when you add the egg then switches back when you add the cheese. More jarring than the final result being different at the end, I think.

PR #281, opened 2025-11-13
Add Silver Weakness as a flaw

Guidesu:
## About The Pull Request
This adds a new character flaw called "Silver Weakness. Gives you silver weakness. Thats it.

## Testing Evidence
So simple :(

## Why It's Good For The Game
RP


CVSthePharmacy:
> This is nice, but it feels like it's too non-impactful as a standalone flaw imo

I toss a silver goblet at you and you get the "MY BANE" message +10 second painstun like in rw1

PR #285, opened 2025-11-13
Wardrobe unbloat

Heheniy:
## About The Pull Request
Unbloats silverface unbloat 
ie 5 different positions of robes of different colors replaced by 1 white robe which you just paint yourself later

Also it now sells 1 item instead of 2 (why it was 2???)

Also removed white robe as re-paint of parent robe and made parent white
Also made default dress white

## Testing Evidence
varedit

## Why It's Good For The Game
unbloat


Jamdrawers:
I feel like it's best if the default colour is either like a beige or brown, stark white is kinda distinctive

PR #286, opened 2025-11-13
Lets Adventurer spellblade choose non-sword weapons

anconfuzedrock:
## About The Pull Request
In addition to sword options, spellblades can choose a mace, whip, spear, or axe, and gain proficiency in that weapon skill.

## Testing Evidence
image

## Why It's Good For The Game
Not really made for me, I don't play spellblade much, and prefer swords. But since swords are generally the best weapon type, I don't see harm in giving spellblade more options. Their summon weapon spell can create other weapons so it'd be nice to start with skill to use them.


CVSthePharmacy:
It's in the flipping name.... I don't choose spellblade and expect a mace, I expect a blade, a sword...

PR #292, opened 2025-11-14
Rewords the AP trait descriptors from 'Physical Feature: New Descriptions.'

Dedwolf:
## About The Pull Request

~~Revert of https://github.com/Azure-Peak/Azure-Peak/pull/4148 while retaining the additional descriptors for other values~~

Rewords the descriptors so they don't sound stupid.

## Testing Evidence

All code.

## Why It's Good For The Game

Because it is pretty stupid examining someone and seeing "He is very moderate" or "She is very tailed" 
Physical descriptors are great but this is just redundant and almost always seemingly out of place. 

image
image
image
image
image

Actual descriptors, they do not sit right at all. 

When players are masked they look kind of neat, for example, highlighting someone might have them appear as 'Fanged Rogue' or 'Furred Blaggard' and-

Actually as I'm writing this I think it'd be better to keep them invisible and retain it as a bit of flavor when highlighting over someone.

### DIFF
--- a/code/datums/mob_descriptors/descriptors/trait.dm
+++ b/code/datums/mob_descriptors/descriptors/trait.dm
@@ -2,152 +2,133 @@
 	abstract_type = /datum/mob_descriptor/trait
 	slot = MOB_DESCRIPTOR_SLOT_TRAIT
 
-/datum/mob_descriptor/trait/moderate
-	name = "Moderate"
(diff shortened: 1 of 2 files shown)


Guidesu:
There is a problem that causes heavy runtiming due to preference saving errors. Un-TMing it for now until a solution is found.

PR #253, opened 2025-11-04
AzurePeak-Parity Sync

RadiantFlash7:
## About The Pull Request
This PR is a direct sync up to AP as of 2 days ago, and will be featuring some content PRS from  Scarlet, which will be linked and posted once I've actually ported them in.
There was 551 conflicts, with 264 errors I had to correct during this port, so I may have made a mistake or two. This PR is almost certainly not perfect, but any major kinks can be worked out during a TM.
A giant shout out to the AP devs and contributors who've done all this work to which I can't try and credit individually- Their commits are the best I can do.
Notably, this PR Moves us to using AP's modern version of dunworld. Hopefully with our mapmaintainer's advice, we can move to a unique version of it as well.

Various Ports from SR: 
https://github.com/Scarlet-Reach/Scarlet-Reach/pull/840
https://github.com/Scarlet-Reach/Scarlet-Reach/pull/1325
https://github.com/Scarlet-Reach/Scarlet-Reach/pull/1234 
https://github.com/Scarlet-Reach/Scarlet-Reach/pull/1237
https://github.com/Scarlet-Reach/Scarlet-Reach/pull/1247 
https://github.com/Scarlet-Reach/Scarlet-Reach/pull/1186 
https://github.com/Scarlet-Reach/Scarlet-Reach/pull/1110

## Testing Evidence
image
image
image

## Why It's Good For The Game
I have been spending all my time porting individual changes piece meal, and it cut into my time to do projects.
If our server wants to stay competitive, we really need to keep parity with AP and Scarlet. To that end- this massive sync PR.
I'm thinking of doing this once every 2 months from now on.

PR #296, opened 2025-11-15
Sword's Bash Intent soft un-nerf.

CloverIsLucky:
## About The Pull Request

After parity( #253 ), both non-blunt strike and sword's bash intent have become copypaste of each other. This PR what it does is slightly un-nerf the changes from the parity no one asked for with increased damange modifier from 0.8 to 1.2 plus a small click delay.

Also Bash intent has been renamed from "pomel bash" to "crossguard bash".

## Testing Evidence

It will work, trust me.

## Why It's Good For The Game

I think I don't need to explain why is this good for the game. After noticing sword's bash intent became non-blunt strike 2.0, I felt the urge to make this PR to make it different again.

### DIFF
--- a/code/game/objects/items/rogueweapons/melee/polearms.dm
+++ b/code/game/objects/items/rogueweapons/melee/polearms.dm
@@ -107,7 +107,8 @@
 	attack_verb = list("bashes", "strikes")
 	penfactor = BLUNT_DEFAULT_PENFACTOR
 	damfactor = 1.2
-	swingdelay = 12
+	swingdelay = 13
+	clickcd = 13
 	item_d_type = "blunt"
 	intent_intdamage_factor = BLUNT_DEFAULT_INT_DAMAGEFACTOR
 


code/game/objects/items/rogueweapons/melee/polearms.dm
```diff
@@ -101,12 +101,13 @@
 	clickcd = CLICK_CD_CHARGED
 
 /datum/intent/sword/bash
-	name = "pommel bash"
+	name = "crossguard bash"
```
Draganfrukts: this should honestly remain as pommel bash, it's not very fitting to hit someone with your crossguard if you're holding your sword normally (most swords get one-handed bash), especially because the crossguard is what it is versus pommel strikes being the equivalent of a pistol whip

CloverIsLucky: We already have "pomel strike", so what you suggest is redundant.

CloverIsLucky:
Ah, thanks @Guidesu haha you caught me balacing this before you merged it 😅  I hope the PR's look okay ^w^

code/game/objects/items/rogueweapons/melee/polearms.dm
CloverIsLucky: 😛

CloverIsLucky:
Let's see how it goes...

PR #310, opened 2025-11-17
Removes hoarder heartbeast trait

Heheniy:
## About The Pull Request
removes absolutely coal heartbeast trait that makes it to steal from treasury with no counter

## Why It's Good For The Game
less coal


ProbablyCarl:
We're keeping this for now. Find a way to edit it, as opposed to removing altogether.

PR #319, opened 2025-11-19
Martyr Boons + Refluff

ProbablyCarl:
## About The Pull Request
Does a few things. Primarily uncouples all the Martyr stuff and shoves it into a folder, so it's not as messy. Same thing I did with unarmed at one point, though parity reverted that. Gives them expert wrestling, too, given they didn't have it for some reason.

Additionally, it refluffs the Martyr. Marking them specifically as guardians of the Prelate, so they've an assigned task and have something to play into. While also shoving onto them that they're _weapons_ of the Church. Not a hero to solo the Inhumen. Instead a conduit of the faith and intended to perish in pursuit of its goals.

Finally, it provides Martyr with three boons to choose from, giving them otherwise very rare or powerful traits to build how they're going to club the Inhumen. There'd been intentions to make patron specific boons, but I was drawing blanks, so those can come later. These can remain as generic fallbacks for all patrons, when such happens, so you don't lose out.

```
Endurance:
 - Provides them hard dismemberment. Recycles the rare trait.

Indefitigable:
 - Provides battle ready. Which means passive energy regen, with no drain from being in combat mode. 
   A trait recycling of an incredibly rare trait.

Hollow:
 - Immunity to magic.
 - Inability to cast, even if trained by the archivist in arcane and given scrolls / whatever.
 - Counter-Counterspell because magic doesn't exist.
```
On top of that, it also provides a new standard trait for the Martyr, built in, unique to them.
```
Conviction:
 - Prayer provides a small healing miracle for twelve seconds.
 - Prayer provides a small amount of nutrition.
 - Can hear all Tennite prayers.
 - Another Tennite praying provides them a powerful two second heal.
```
## Testing Evidence
image
image
image

## Why It's Good For The Game
Martyr is weird and left behind in terms of content.
As it stands, Templar blows it away outside of using the ult. In a lot of cases.
Is it a good idea? Probably not. Would it be fun? Absolutely.

Further stuff to set Martyr aside.


Pidgey-chan:
Why would the martyr hear prayers instead of the priest, the one who's actually responsible for tending to the spiritual needs of the populace? Why are we buffing an insane-o snowflake role like this to begin with? Templars being very strong is an issue with Templars, not cause to inflate the guy with fullplate and a flaming sword.

Murderfoxen:
> Why would the martyr hear prayers instead of the priest, the one who's actually responsible for tending to the spiritual needs of the populace? Why are we buffing an insane-o snowflake role like this to begin with? Templars being very strong is an issue with Templars, not cause to inflate the guy with fullplate and a flaming sword.

So your issue is with priests not hearing prayers, and templars needing a buff? 

Is that what makes this not good for martyrs? 
It has soul man. PRIESTS are already HUGE divine casters and the literal VOICE of the church. This is the VOICE of the people. It's a really cool idea!

CVSthePharmacy:
GRAHHHH I HATE MARTYR REMOVE THE DAMN ROLE ALREADY

code/__DEFINES/traits.dm
```diff
@@ -232,6 +232,7 @@
 #define TRAIT_RACISMISBAD "Heritage Vision"
 #define TRAIT_VENOMOUS "Venomous"
 #define TRAIT_OVERTHERETIC "Overt Heretic"//Applied on someone who has rites buffs. Armour, rituos, etc.
+#define TRAIT_CONVICTION "Conviction" //Can hear Tennite prayers. Praying heals and provides nutrition.
```
Guidesu: You could give it to priests too. That is my only grip i have.

Pidgey-chan:
> > Why would the martyr hear prayers instead of the priest, the one who's actually responsible for tending to the spiritual needs of the populace? Why are we buffing an insane-o snowflake role like this to begin with? Templars being very strong is an issue with Templars, not cause to inflate the guy with fullplate and a flaming sword.
> 
> So your issue is with priests not hearing prayers, and templars needing a buff?
> 
> Is that what makes this not good for martyrs? It has soul man. PRIESTS are already HUGE divine casters and the literal VOICE of the church. This is the VOICE of the people. It's a really cool idea!

Between the priest and the martyr, if anyone should be hearing prayers it isn't the frag role.

I never said to buff templars. Read slowly.

ProbablyCarl:
> Most of this is cool, but you absolutely can’t hand out hard dismember to anyone who has easy access to Necra’s Vow. Hard dismember locks every sever chance at five percent. Honestly, the trait shouldn’t have gone to non-antags in the first place.
> 
> Pairing it with something that passively feeds you blood, like the vow, turns it into a whole other monster. At that point, the only real way to kill someone running hard dismember + vow is to pray you hit that five-percent beheading roll.

Unfortunately, that's an issue with the _vow_. Not this. But, yes, you've a point.
Unfortunate as it is that some absurd synergy like that pops up. I hadn't even considered it, because I've not seen a Necran Vow another Necran in over a month. I could just add an exception to Martyr, I suppose, for the Vow. Disallow them from benefiting from it.
Alternatively, we could just rework the Vow altogether, but that's out of the scope for this PR.

Murderfoxen:
> Between the priest and the martyr, if anyone should be hearing prayers it isn't the frag role.


anyway, so I think this is a way to redirect the 'frag role' away from 'frag roleing' 

I cant comment on the combat traits beyond thinking that Martyr is already more of a HRP role; but making them tough to kill isn't an issue because it isnt as much of a 'frag role' as anything else. 

We have vampire towners/adventurers and we have wretches and bandits; I feel that they should be scared of martyr.

PR #321, opened 2025-11-19
Keybind options for subtle emotes

Useroth:
image

image

image


Heheniy:
whats the difference between subtle and big subtle

PR #327, opened 2025-11-20
Rockhill update 5: Dream Warriors

Jamdrawers:
## About The Pull Request

More rockhill work!

1) Adds a new weird special zone that adventurers spawn in before they warp into the proper map. I think it could be cool. Uses the old Underworld sprites.

2) Converts the knight captain's room into a knight's room (still incredulous that there are 4 knights in a round) and creates a fancy new knight captain's room upstairs in the keep overlooking the sea

3) Fixes the sea to use seawater (turns out that's important in APcode?)

4) Adds quest ledgers to the merchant and steward and mercenary's guild (considering that the one in rockhill is a fair bit further away than in Dunworld I think it makes sense)

5) Moves spawnpoints to be more consistent, with keep personel spawning together in the throne room, Men at Arms spawning together in the MAA barracks, and also trying putting the marshal in the old Mayor's building. Generally sticking to spawn points being in a main meeting area for an area rather than a private bedroom so you spawn next to your fellow gamer to get more quickly acquainted at roundstart

6) Changes the guild smith's upstairs layout, replacing the arificer and architect rooms with a guildmaster's office-and-bedroom

7) Fixes small benches from blocking the wrong way, and replaces them with a version that just doesn't block you at all. They're small benches, they're supposed to be small! Would do the same for Dunworld but might make issues with conflicts.

8) Increases fyritus spawns significantly in the Murderwood, giving a particular reason to go there.

## Testing Evidence

image
The Wayfarer's Dream - a little weird meta spawn point that adventurers start in before they start their journey. Simply walk into one of the portals to wake up from the dream!

image
Knight Captain's room - with a sea-side balcony!

image
New Guildsmaster's bedroom+office

image
Harbour turns to seawater as it meets the sea

## Why It's Good For The Game

Fixes and tweaks and content!

### DIFF
--- a/code/modules/roguetown/mapgen/bograt.dm
+++ b/code/modules/roguetown/mapgen/bograt.dm
@@ -1,4 +1,3 @@
-//genstuff
 /obj/effect/landmark/mapGenerator/rogue/bograt
 	mapGeneratorType = /datum/mapGenerator/bograt
 	endTurfX = 255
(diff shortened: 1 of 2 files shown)


Murderfoxen:
Hell yeah, I love Mapping PRs like this! 

It would be really cool if you added these adventurer spawns to dun_world as well! I love the idea of spawning in as groups like migration waves for those roles! its better than everyone doing everything at the edge of town

PR #332, opened 2025-11-21
Inquis Re-Pass

ProbablyCarl:
## About The Pull Request
Many changes. Awaiting a unique sprite. Considering a few other things.
```
Inquisition:
 - Uncoupled Naledi Disciple. It's now 'Sojourner', a unique anti-magic-slowing focused subclass.
  - Unique barrier spell, causing them to be entirely immune to magic while it's active. Has next to no durability or defensive utility otherwise, and breaks within a single strike or two.
  - Makes use of the old 'frostbite' spell, as the singular individual using it.
  - Their spell lineup is curated around slowdown and utility. 26 spellpoints in total. No free spells. All of a preset.
  - View the 'role' dropdown below for stats.

 - Uncoupled Arbalist Confessor, creating a unique heavy-ranged subclass.
  - Unique heavy crossbow, the 'sauterelle'. Access to three unique bolts. Very long reload time. Higher damage, pen and range. Can only fire the following unique bolts:
   - Heavy bolts are high damage, high pen. Nothing will stop these. All following bolts continue this trend. They get twenty in their quiver. Unobtanium otherwise.
   - Stake bolts act as holy bolts, though much higher damage. They get five in a unique belt. Unobtanium.
   - A singular 'tempest bolt' in their bag, which blinds and exposes the target. Unobtanium.
  - View the 'role' dropdown below for stats.

 - All other classes for the Inquistiion touched upon in some way. Old stat spreads returned. Classes cleaned up. Etc.
   - Confessor given torture mechanics, and the trait to fit it.
   - Confessor given miracles, the same as every other Inquisition class.
   - Psyaltrist lost its SPD/STR, got +2LCK/+1CON instead.
   - Psyaltrist athletics and knives lowered from expert to journeyman.
   - Psyaltrist devotion gain swapped with Adjudicator.
   - Adjudicator's wrestling raised to expert, in line with parity revert.
   - Adjudicator's devotion gain swapped with Psyaltrist.
   - Adjudicator's other armour choice no longer removes the speed malus.

 - All 'old' Psydonian weapons swapped their stats with the newer ones. Not sure WHY the actual non-adventurer variants were objectively worse. Additionally, removed blessing from the old sets across the board.
 - PQ requirements for Inquisition jumped significantly. Absolver and Orthodoxist raised to 20. Inquisitor raised to 30. Too many new folks played these roles like absolute muppets. Not sure why it was so low, when they're so absurd and capable of breaking the round's flow with a lack of RP.
```
## Testing Evidence
<details>
  <summary>Roles</summary>

image
image
</details>
image

image
image

image
image
image

## Why It's Good For The Game
Naledi disciple was absurd. This removes the infinikatar batman setup and makes them into something unique.
Arbalist was just out of place. Hopefully this gives them a purpose.

That aside, the varying changes are just returning stuff from before parity, or otherwise settling the classes about in a way that tries to wind them back from absurdity. Or make them unique, for that matter.


Pidgey-chan:
Silver stake bolts I get. But does the Inquisition need 100 damage 100 pen crossbows? For what, legbanging knights? They're already a capable fighting force.

Pidgey-chan:
> Is it absurd? I suppose. I just didn't consider it too big an issue, when we've other really broken stuff. A stake thrower that can drop a man felt flavorful and not as bad as some of the other stuff. Not hard to tweak the damage increase value on the crossbow itself, though, as the bolts are identical in base damage and it's just the modifier on the crossbow increasing it.

I would think the answer to really broken stuff is to trim it, not bless the Inquisition with the 100 AP leg disabler painscream-o-matic

ProbablyCarl:
I'll significantly drop the damage on the crossbow.
They've no other skills. That was their entire thing, and I'd taken to thinking it was fine. It's whatever, ultimately. A shame, though.

ProbablyCarl:
Identical in damage to the standard crossbow. Maintains the super pen. Not as if those don't go through everything, anyhow.

PR #345, opened 2025-11-24
Fix and 'Expansion' for both maps

HelennaC:
## About The Pull Request
Fix for both Rockhill and Dum world [and somewhat expansion].

Rockhill:
- Removed black/void turfs
- Area fix for inn (They get area buff, so the basement must follow the same rule)

Dum World
- Same area fix as Rockhill + Farm too (belongs to inn)
- Removed lonely saigas from the merc house
- Return of one saiga to keep's stables (renamed just like Warden's saiga)
- Expanded Goblin camp
- Swap of far travel points (Wrench [Expanded] and Vampire)
- Little love to both Dendor's temple and altar
- Small fix to avoid body blocking at sewers
- Some small changes related to misplaced mobs/tiles and unused dungeon

## Testing Evidence
Tested, it just works.

## Why It's Good For The Game
Some needed changes that should all together improve the flow and artistic feeling to the game, and 'new' content that was already done but never published due to TMs.

### DIFF
--- a/_maps/map_files/rockhill/rockhill.dmm
+++ b/_maps/map_files/rockhill/rockhill.dmm
@@ -4997,7 +4997,7 @@
 	dir = 4
 	},
 /turf/open/floor/rogue/dirt/road,
-/area/rogue/under/town/basement)
+/area/rogue/indoors/town/tavern)
 "bFU" = (
 /turf/open/floor/rogue/tile/masonic,
 /area/rogue/indoors/town/manor/rockhill)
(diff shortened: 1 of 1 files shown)


Jamdrawers:
There are a couple of mapping PRs already floating around so for compatibility reasons I think I might copy the rockhill inn area change to the current main rockhill-changing-PR (The black tiles are already fixed in my version for the next update) 

Could you compile some screenshots of your changes/be more specific?

Jamdrawers:
After I have a look at it I can see about merging yours with https://github.com/Rotwood-Vale/Ratwood-2.0/pull/349

HelennaC:
> There are a couple of mapping PRs already floating around so for compatibility reasons I think I might copy the rockhill inn area change to the current main rockhill-changing-PR (The black tiles are already fixed in my version for the next update)
> 
> Could you compile some screenshots of your changes/be more specific?

Rockhill or Dum world?

PR #349, opened 2025-11-24
Dungeoneer mapping fixes

cynical24:
## About The Pull Request

i've been playing dungeoneer recently and noticed that dun world's cells just don't proc the dungeoneer's trait, and places where it'd make sense to just don't work either

also fixes the prisoners to spawn inside the watchhouse dungeon and not the warden one

this is my first time actually doing any mapping so if i fucked up lmk

## Testing Evidence

image
image
image

## Why It's Good For The Game

helps dungeoneers do their dungeoning job in the dungeons 🙏🙏

### DIFF
--- a/_maps/map_files/dun_world/dun_world.dmm
+++ b/_maps/map_files/dun_world/dun_world.dmm
@@ -38398,7 +38398,6 @@
 /area/rogue/outdoors/mountains)
 "joo" = (
 /obj/structure/chair/stool/rogue,
-/obj/effect/landmark/start/prisonerr,
 /turf/open/floor/rogue/cobble,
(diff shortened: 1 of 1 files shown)


Jamdrawers:
Map merges are trickier because it's all one big file

Jamdrawers:
> > i was going to fix the prisoners to spawn in the watchhouse dungeon and not the warden one
> 
> changing the tiles to cell inside of the tower removed the warden's buff, so i decided to not change it tho

Can adjust that manually using variables, I think I did the same for rockhill

Jamdrawers:
Can close this now that it's been implemented in another PR

PR #361, opened 2025-11-28
Instrument changes

wow50000:
## About The Pull Request
Reorganises instruments in the instrument file to be alphabetically organised
adds shamsen, drum and trumpet to bard, cantor and spellsinger(musician subclasses of rogue, cleric and mage)

I DID NOT TOUCH MINSTREL as I'm not sure what to do and a bard player or someone who knows the music system better than me can tell me what to do

Did all of this after chugging two monster energies:tm: and eating a weird black flat thing that is giving me a really bad stomach ache

## Testing Evidence
image
image
image

SHOULD WORK

## Why It's Good For The Game
Currently it makes no sense for the music classes to not be able to pick every instrument unless there was some grand idea to make every instrument have different stats for whatever reason, I believe this counts as a bugfix as the performer virtue can be taken as a workaround to get those three items however this is just poor gameplay.

Also ordered the instruments alphabetically because I felt like it and it looks neat :3

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/combat/mage.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/combat/mage.dm
@@ -196,25 +196,31 @@
 		if(/datum/patron/inhumen/zizo)
 			H.cmode_music = 'sound/music/combat_heretic.ogg'
 	if(H.mind)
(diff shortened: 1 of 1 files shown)


ProbablyCarl:
This LOOKS fine.

PR #363, opened 2025-11-28
Constructs Toxin Immunity | Black Rot Fix

ProbablyCarl:
## About The Pull Request
Simply makes constructs immune to poison.
Additionally, it makes all toxin immune characters immune to the Pestran blight.

Constructs could catch the blight, and, given no surgery, were impossible to cure. Among other issues. Also, it doesn't really make sense for the undead to suffer it. Very weird.

There's a balance implication here or something with Rotcured becoming immune, but I don't care enough to touch that.
## Testing Evidence
image

## Why It's Good For The Game
Actual bug fix. Also I hate Pestrans.


ProbablyCarl:
> With this constructs can bypass the overdose check from stat potions that killed you if you tried to stack potions.

Could we not just take the SR approach, then, and prevent them from healing via potions altogether? All this does is add no toxins, which means anyone with rotcured can do this, too, so I didn't consider it a huge issue.

ProbablyCarl:
Adding the label, because someone forgot to label it.

PR #365, opened 2025-11-28
Taur Limb Overlay and Grabbing Fix

cynical24:
## About The Pull Request

fixes the limb damage overlay to work with taurs correctly, so you're not left guessing what your damage is on said limbs- or if you're bleeding. etc. pretty annoying bug

also fixes grabbing taur limbs -- grabbing their legs or feet will autocorrect to their taur body, so you can slow bleeding and all that. it also prevents them from kicking, so dw about that. bandaging also works as you'd expect.

## Testing Evidence

image
image
image
image

## Why It's Good For The Game

fixes some really detrimental taur bugs, yuuurh.

when do i get railed by them


cynical24:
i think their limbs still takes like a fucking gazillion omegadamage because i keep seeing them go to MANGLED so abysmally quick, but i might be tweaking

PR #373, opened 2025-11-29
Taurs get their feet slot

cynical24:
## About The Pull Request

bit of a controversial PR probably

re-adds taurs being able to wear shoes, and attaches the corresponding clipping mask. also removes their caltrops resistance to balance accordingly.

## Testing Evidence

image

## Why It's Good For The Game

i think it's stupid that you just get gutted an armor slot because you want to be "haha funny taur", like this is genuinely just an unfair advantage for any non-taur. i've killed a giant barbarian prawn in a few seconds simply because i kicked him over and he had no feet armor, so i could just rend at his feet and i win

### DIFF
--- a/code/datums/components/caltrop.dm
+++ b/code/datums/components/caltrop.dm
@@ -22,7 +22,7 @@
 
 	if(ishuman(AM))
 		var/mob/living/carbon/human/H = AM
-		if(HAS_TRAIT(H, TRAIT_PIERCEIMMUNE))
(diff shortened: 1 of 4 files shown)


ProbablyCarl:
Lamia specifically didn't have it because they're given longstrider and obnoxious to catch, alongside stronger kicks.
Pretty sure this screws with them.

cynical24:
> Lamia specifically didn't have it because they're given longstrider and obnoxious to catch, alongside stronger kicks. Pretty sure this screws with them.

fixed by checking for the lamian tail trait

Pidgey-chan:
I think taurs are ugly so downvoted

PR #374, opened 2025-11-29
Sexual Harassment (Examine Flavor)

cynical24:
## About The Pull Request

lets you.. uh.. look at peoples asses... n' dicks, and hips. hell yeah.

if you're 1 tile adjacent to a living player, and targetting groin, on examine it will run the usual checks, but before it reaches any of the other else ifs, the game will check if you're 1 tile adjacent of them, including diagonals.

when you are one tile adjacent, it'll check where exactly you're at in comparison to the player (infront, diagonally infront, to the side, behind, diagonally behind) and set the target (ass/crotch/hips).

before it's done it'll do the usual sneaking check, and then send the examine message. this shouldn't fuck up anything because of how many checks the initial else if statement performs. if you're in combat mode and do the same thing it'll still work because you're close enough, otherwise it goes to the last check.

i think old RW had this feature? so i brought it back.

## Testing Evidence

image
image
image

## Why It's Good For The Game

you get to look at twink butts and stuff and i don't have to bring back ratwood's old bodypart examining where you're staring at everybody's chest (i didn't really like it)

### DIFF
--- a/code/modules/mob/mob.dm
+++ b/code/modules/mob/mob.dm
@@ -483,33 +483,34 @@ GLOBAL_VAR_INIT(mobids, 1)
 	if(isliving(src) && src.m_intent != MOVE_INTENT_SNEAK && src.stat != DEAD)
 		var/message = "[src] looks at"
 		var/target = "\the [A]"
+		var/zone_text = parse_zone(zone_selected)
+		var/mob/living/T = A
 		if(!isturf(A))
 			if(A == src)
 				message = "[src] looks over"
 				target = "themselves"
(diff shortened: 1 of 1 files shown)


Murderfoxen:
hey can u make it so it does this all the time and not just in combat mode? I think we should all be seeing who's looking w/o any respect

cynical24:
> hey can u make it so it does this all the time and not just in combat mode? I think we should all be seeing who's looking w/o any respect

stares at you

ProbablyCarl:
Actually don't we have this already?
Turn on combat mode. It exists.

PR #375, opened 2025-11-30
Maneater problems

Rayforx:
## About The Pull Request
Removes the ability for man-eaters to take away people's limbs. 
It instead dislocates the limb and adds a wound to it.

## Testing Evidence
image

## Why It's Good For The Game
Removes one of the most hated mechanics in the game, a mechanic that people only deal with when they are alone, away from the town. It's stupid, and not much is lost from removing the limb-removing part.

### DIFF
--- a/code/game/objects/structures/maneater.dm
+++ b/code/game/objects/structures/maneater.dm
@@ -107,7 +107,7 @@
 		victim.flash_fullscreen("redflash3")
 		playsound(loc, list('sound/vo/mobs/plant/attack (1).ogg','sound/vo/mobs/plant/attack (2).ogg','sound/vo/mobs/plant/attack (3).ogg','sound/vo/mobs/plant/attack (4).ogg'), 100, FALSE, -1)
 		if(prob(chew_factor * 15))
-			victim.adjustBruteLoss(50)
+			victim.adjustBruteLoss(20)
 			limb.add_wound(/datum/wound/fracture)
 			seednutrition += 25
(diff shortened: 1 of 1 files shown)


CloverIsLucky:
Skill issue.

kepesk1:
> Skill issue PR.

Having watched well over 30 people quit in the last 2 years over maneater RR, and watched 3 brand new players quit in the last week because of them, sure -- it's a skill issue. But some maneaters grab you right off the road even when sneaking, and they RR you suddenly. Some juveniles can even do that at 5 mins timer if you're low strength and can't resist. It's a new player killer of any level, and a veteran player killer if you're weak and happen to spawn in the bog (which is basically a 50/50 chance).

The other day I played a 7 strength mage, got yanked off the road while sneaking on an empty tile, and lost 2 limbs after 18 failed resists. I spawned with no knife, so had no means to break the maneater. The maneater was, by the way, hidden behind a tree sprite. There was no way to detect it. This is consistently how veteran players still die to them, because they spawn behind trees (same as kneestingers, except kneestingers don't RR you). 

Once upon a time, the maneater density was low enough that players could RP in the bog and do homesteader stuff. Not anymore. It's way too dangerous, not because of mobs since sneaking helps with that, but because of maneaters. Mature ones will kill you sneaking or not if you aren't careful, and after 30 minutes basically all of them are matured in the code. It's made Rockhill bog a true hellscape, and the majority of adventurer spawns are located there.

PR #391, opened 2025-12-02
Digging Buff

ProbablyCarl:
## About The Pull Request
This is a return of [#2011](https://github.com/Rotwood-Vale/Ratwood-Keep/pull/2011)

## Testing Evidence
I did not test this. Slop PR.

## Why It's Good For The Game
Digging up multiple things on one tile won't crash the economy. In fact, it's a massive QoL improvement for people gathering both stone and bait. Does this kind of turn away the purpose of mining a bit? I suppose. We could reintroduce the check so you have to quickly remove the stone before digging again. As before. But it really shouldn't be too big an issue I'd think.


Guidesu:
i mean this is so small and we had this before.

PR #400, opened 2025-12-05
New Cleric Subclass: Shaman

Guidesu:
## About The Pull Request
The totem mechanics balances their now buffed costs... example is that they can only cast 10 lesser heals before running out of totemic energic,plus they also need to worry about their devotion, and if you are playing mage, two fireballs completly drains you.

They don't get witch shapeshift. So no shaman crows.

image
image
image
image
image

## Why It's Good For The Game
BECAUSE ITS FUN!

### DIFF
--- a/code/__DEFINES/traits.dm
+++ b/code/__DEFINES/traits.dm
@@ -74,6 +74,7 @@
 #define TRAIT_PURITAN_ADVENTURER "Interrogator"
 #define TRAIT_MIRROR_MAGIC "Mirror Magic"
 #define TRAIT_WITCH "They fear me, but I am useful to them."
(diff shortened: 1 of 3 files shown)


Hmeister-real:
> I don't think witches should be this powerful or this _versatile_ - other roles exist for a reason!

Same.. this is all cool but... for _witches_...?

Kepteyn:
It broke magic and miracles, but it's an awesome idea! Keep at it!

Ovilum:
To be honest, with the witch stuff? I feel like it needs to be its own stand alone class along with this shaman role underneath it. Would that be asking too much?

PR #404, opened 2025-12-06
Port Old RW Examine

cynical24:
## About The Pull Request

yes i know i just made a PR fucking with examines 5 days ago

this ports old ratwoods _FANCY_ examining system with some tweaks and fixes to work with AP's codebase.

most notably, you can now look at peoples chests, legs, arms, etc without the "looks at" examine text popping up anymore

from what I can tell it works by taking which direction both players are facing and dictating whether or not they're "behind" one another using that. this of course leads to inconsistencies like actually being in front of someone, but managing to examine them while they're behind you (just out of your field of vision likely) and the game considering you as "behind" them, but it's nothing to cry over i'd think.

using fixed eye will also allow you to get some unique examines, such as viewing their "thighs" instead of just "legs", same goes for grabbing someone, etc.

everything generally works fine; combat mode works as expected by falling back to the basic examine texts instead of the complicated ones, instead of saying "Elliot looks at Elliot" when you examine your own head, it'll now just say "Elliot looks over themselves", etc.

## Testing Evidence

image
image
image
image
image
image

## Why It's Good For The Game

awoooooga boobies!

### DIFF
--- a/code/__HELPERS/medical.dm
+++ b/code/__HELPERS/medical.dm
@@ -94,3 +94,219 @@
 		if(ORGAN_SLOT_REGENERATIVE_CORE)
 			return "regenerative core"
 	return slot
+
+/proc/parse_zone_fancy(zone, combat, combattarget, closeby, turnedaround, ontheground, grabbing, squinting, uncovered, dicked, pussied, strength, self = FALSE)
+	switch(zone)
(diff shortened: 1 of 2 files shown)


cynical24:
should probably note i can probably fix the behind variable things by pulling some of that code i have in my sexual harassment PR

PR #405, opened 2025-12-06
The not exactly re-return of racism again

Jamdrawers:
## About The Pull Request

Trying https://github.com/Rotwood-Vale/Ratwood-2.0/pull/156 again. Last time it was very swiftly refused within the day, but the maintainer who rejected it stepped down since, and I think there are a lot of people who still agree with this. Feel it warrants a little more consideration.

Restricts the following classes from Shunned or lower races:
Duke
Prince
Knight+Knight Captain
Marshall
Councillor

I left open other nobility roles for now, like court physician, steward and magos because I could imagine them more likely to be attained on some specialised merit, as well the Hand who is personally selected. Suitors are also welcome to try their hand, even if their odds might be slim.

For reminder, Shunned or Lower means Goblins, Verminfolk, Kobolds, Halforcs, Revenants and Lamia
We could adjust Lamias out of being Shunned if we want - I don't entirely know what our lore on Lamias is supposed to be and whether we're committed to that. The exact placements could be reconsidered.

Edit: At Useroth's suggestion I Unshunned Lamias

Edit: Shunned tieflings. Vile devil-spawn!

Edit: Been hearing from a lot of people that there isn't a big desire to treat verminfolk as shunned, since it's such a varied race and can be so many types of things like adorable squirrels and bunny rabbits, and *just* being small isn't enough to be treated the same as a goblin for example, and there isn't much lore associated with them to reinforce their place in-setting. So next update does the following
1) unshuns verminfolk
2) rename verminfolk to "Critterkin"
3) removes mood debuff for looking at critterkin
Also changed the description to mention how Verminfolk is a very derogatory and offensive word that some people use for them.
image
image

EDIT!!: Turns out I can't do this without wiping player profiles for anthromorphsmall characters. At least not without some coordination and help. Could go with it but might need to bug Useroth about it.

EDIT!!!: Bugged Useroth about it and fixed player profiles. Will be good for next deployment!

I ponder making a subrace specifically for rat-people that really IS shunned, since rat-themed characters are pretty common and pretty shunnable.

Could go either way on kobolds

## Testing Evidence

image

## Why It's Good For The Game

Gives a bit more texture and believability to the world and reinforces the game through mechanics. It's kinda hard to take goblin/orc/verminfolk/etc racism seriously when there's a goblin duke and a verminfolk prince, and "if the duke is goblin then just coup them" was, I think, not a very satisfying response. We have racelocks for metal constructs and I think that works fine.


Useroth (approved):

Jamdrawers:
> I agree with this PR.
> 
> And yeah, lamias could probably be moved out of shunned.

Done

Pidgey-chan:
I shun lamias personally bc they're usually major uggos

RadiantFlash7:
Truth be told, the point of rejecting racelocks early into ratwood 2.0 was entirely for population reasons. I wasn't willing to block people out of roles and drive them away when we were only hitting 30 players at most. With Ratwood's higher pop, that reason no longer really exists. Though, I imagine the some of the more popular races like kobolds and lamia will be displeased.

Pidgey-chan:
This shouldn't surprise anyone around for RW 1, these are the "legacy" racelocks, maybe even a bit more generous since drow and tieflings don't seem to be shunned.

Verminvolk I could maybe see unshunned since it's not really a race. Otherwise, about time. Goblin dukes spit in the face of believability in this setting.

Murderfoxen:
There comes a time we have to accept that people cannot be expected to roleplay super-seriously every round; and that in the golden moments; we are often laughing despite the grimdark expectations of the server. 

I for one, am happy lamias did make the Cut. I almost lost this glorious, very 'HRP' character who has NO VISIBLE CONNECTIONS TO ANYTHING FROM THE STAR-WARS FRANCHISE:

Glubba the fox!
Glubba the fox

and as we can all tell, Astrata has chosen him to the duchal. Because he is peak.

Troontjelolo:
https://github.com/user-attachments/assets/8e2f2405-122a-4c57-a318-57712ad69599

PR #414, opened 2025-12-08
Omnitowner Disabling

ProbablyCarl:
## About The Pull Request
Homesteader was removed pre-parity.
It'd been returned post-parity even worse, with the ability to select a handful of skills, traits, etc. Further diminishing the purpose of towner roles as a whole, alongside some other dedicated setups.

I don't know why we kept this. I've been asked by both player and contributor alike to remove this. I also really, _really_ dislike it. For it serves no real purpose beyond allowing someone to do everything.

Not even mentioning that you can select the equivalent of more virtues.

So, this disables it. Again. A second time, as opposed to the outright removal I did before from what I recall.
On top of removing it from subclass view.

## Testing Evidence
Two line change. It'll work.

## Why It's Good For The Game
See 'about'. We weren't even meant to have this returned.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/villager.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/villager.dm
@@ -30,7 +30,6 @@
 		/datum/advclass/cheesemaker,
 		/datum/advclass/drunkard,
 		/datum/advclass/fisher,
-		/datum/advclass/homesteader,
 		/datum/advclass/hunter,
 		/datum/advclass/hunter/spear,
 		/datum/advclass/miner,


AisuDracov:
Omnitowner was added as a solution for folks who wanted to go off and do homesteading away from town after AP added in skill limits. Honestly, if we just remove that, the skill caps, we'd not have a reason for needing omnitowner.

Jamdrawers:
> Omnitowner was added as a solution for folks who wanted to go off and do homesteading away from town after AP added in skill limits. Honestly, if we just remove that, the skill caps, we'd not have a reason for needing omnitowner.

I thought omnitowner was added waaaaaaaay before skill limits? It was here in the first rebase in september. Skill limits were a fairly recent addition iirc, added just before the parity ~a month ago.

Anyway, I think with the way skill caps are set up, you can do pretty much everything that is needed to homestead away from town anyway. Most towner roles get the Expert_Homesteader trait and can get up to legendary in skincrafting or cooking or whatever it is you want. And even as a role without Expert_Homesteader, if you want to Homestead as like a mercenary or combat adventurer or whatever, you can still grind high enough to do whatever it is you need to do, it's just that you don't have the carrot dangled in front of you to keep grinding for legendary and can't specialise in it the way a character who spent their entire life practicing it could.

Though that reminds me I did add a pair of roles to the pilgrims/refugees that were especially suited for out-of-town-homesteading that might have gotten obliterated in the parity.

PR #421, opened 2025-12-11
Allow Noccites to see in the dark again

Skyglazer:
## About The Pull Request

This PR adds back the trait that allows followers of Noc to see in the dark a little bit, which AP decided to give to Zizo worshippers instead for some reason. This PR does not remove the trait from Zizoids, they still get it. Can be changed separately if desired. Also includes a bonus fix of a missing period at the end of the "Night Owl" trait. As a treat.

## Testing Evidence

dreamseeker_2025-12-11_14-32-34

## Why It's Good For The Game

This is a change we never wanted that came with parity/the rebase/etc. AP changes this for 'balance' reasons, arguing that people would pick Noc JUST for the see-in-the-dark trait. Not a very strong argument, since people who would play in that way are still absolutely given the option to pick Zizo instead. Not like there's any downside to being a heretic of you don't go loud/get caught/etc.

Noccites are literally nocturnal and get tired in the day-time. Being able to see in the dark makes sense. People who would pick it to powergame would already pick Zizo instead, so we might as well have it for flavor.

### DIFF
--- a/code/datums/gods/patrons/divine/noc.dm
+++ b/code/datums/gods/patrons/divine/noc.dm
@@ -19,7 +19,6 @@
 		"NOC SEES ALL!",
 		"I SEEK THE MYSTERIES OF THE MOON!",
 	)
-	traits_tier = list(TRAIT_DARKVISION = CLERIC_T1)
 	storyteller = /datum/storyteller/noc
 
 // In moonlight, church, cross, or ritual chalk


Skyglazer:
> They still get proper darkvision at T1, according to this.

Should that be removed? I figure it's useful for Noccian clergy to be able to give Darkvision to others- something I regularly do on my Noc Acolyte, who got the darkvision trait anyway because he was a >T1 Noccite

PR #422, opened 2025-12-11
Reworks bandage system and adds bandages

john-pawson:
## About The Pull Request
This PR reworks, adds, and fixes a couple of things. First! I'll talk about the rework on how bandages work on the game:

# Rework
- Old system

On the old system, the bandages would work simply: 
There is a variable, called bandage effectiveness, that determines how good of a bandage something is (cloth being the only item, at 0.9)
If you put it on a limb that had bleeding < effectiveness, it stops the bleeding. Completely.
If you put it on a limb that had bleeding > effectiveness, the bandage got bloody and became useless instantly.

This was... to say the least, useless. Most normal cuts already achieve more than 0.9 bleeding and it's more likely that a cut that trivial would heal itself before someone would even manage to bandage it. Not only that, but on any significant bleeding where bandaging would be something you'd want to do, it would be completely ineffective.

- New system

There's three variables:
**Bandage speed**: How fast a bandage is applied.
**Bandage health**: How much a bandage can stop bleeding before being bloodied, and needing to be changed or cleaned.
**Bandage effectiveness**: How well a bandage stops bleeding. Effectiveness doesn't affect the rate at which the health goes down, only the amount of blood you lose while having a bandage.

When you bandage a limb, this limb's bleedrate will get multiplied by the bandage effectiveness.
For example, with an effectiveness of 0.3 and a bleedrate of 10, the bleedrate would be reduced to 3.
After that, the bandage's health will slowly go down depending on the bleedrate, the bigger the bleeding, the faster the health goes down.
For example, a bandage with 200 health, if we continue with the same effectiveness and bleedrate as before, you'd lose a total of 60 units of blood (140 protected by the bandage) before the bandage gets bloodied and you return to your normal bleedrate.
If the bandage manages to make the bleedrate go lower than 1, the bleeding's completely stopped by the bandage. But this doesn't mean that it'll stop taking damage.

I believe this is a way more useful and intuitive way of applying bandages, you apply them on a bleeding wound to help with the bleeding momentarily until it can be treated fully, and replace it if it's been on you for too long.

Arteries will bloody bandages very quickly but the extra few seconds of life may be enough to get you to retreat or heal with a potion before blacking out, and lesser injuries will get lessened by the bandage and given extra time to clot without much blood lost (You could honestly put a bandage over some wounds and just wait for them to heal passively, since the minor ones get stopped completely and won't take much of the bandage's health)

Here are some stats, for those that would like them:
**Cloth**
Bandage speed: 7 seconds
Bandage effectiveness: 50% (0.5)
Bandage health: 150

**Bandage**
Bandage speed: 4 seconds
Bandage effectiveness: 75% (0.25)
Bandage health: 300

Of course, all of these can be changed, hopefully, I get a test merge to balance these out properly.

# Add (New stuff)

Bandages!!
Yeah that's about it. There's actual bandages now, they can come in stacks of four and most medical places have had their rolls of cloth replaced with bandages, they can be bought form the merchant or made with medicine, alchemy or sewing skill, they bandage faster and better than normal cloths but they take extra space (smaller rolls) and are usually more expensive than regular cloth.
You can also use them *as* a regular cloth. Feel free to wipe your floors with bandages, why not.

Bandages will now tell the people nearby if they become bloodied, so now physicians can notice if a bandage needs changing on an unconcious patient, instead of only the bandaged being able to see it.

Un-bandaging is now also dependant on skill, higher skill, faster times.

# Fix

Bandaging speed will no longer depend on the one being bandaged, now it uses the bandager's skill, as it should.
Item bundles will now update their sprite if you insert a material by hand (Click bundle with stick) instead of picking the material with the bundle (Click stick with bundle)
Passive grabs will no longer stop 75% of your bleeding, and now will only stop 25% of it, as intended (The comment says it's supposed to only stop 25%, and aggressive grabbing *lowered* the amount stopped to 50%. Clearly, unintentional)

## Testing Evidence
Checking that the bandages apply correctly on loadouts and containers.
image
Crafting recipes work.
image
Testing that the system works! Different healths, bleed rates, etc.
image
image

## Why It's Good For The Game

I think it's an overall better and more intuitive system than the useless thing we called bandaging beforehand. Also adds new content that helps the physicians, also bug fixes.


john-pawson:
> Bloody bandages were useless? That's crazy.

Clean ones too!

ProbablyCarl:
A return of OldRW bandages? Kinda? This is cool. Doubly so with the super passive wound healing removed.

ProbablyCarl:
There's a conflict with [this](https://github.com/Rotwood-Vale/Ratwood-2.0/pull/418) PR, regarding surgical bags, otherwise we'd quick TM this shit!!!!

Elli-Skala:
there was a PR on azure or SR that allowed banages to be applied while moving to make them more useful mid combat. unsure if thats wanted here

john-pawson:
> there was a PR on azure or SR that allowed banages to be applied while moving to make them more useful mid combat. unsure if thats wanted here

Bandages will get ruined if the limb takes any damage while you're wearing them, so ideally you'd only use them after combat's done or if you're avoiding combat.

PR #425, opened 2025-12-13
Changes Vagabond's Examine Title into Beggar

yupup123:
## About The Pull Request

Adds an `examine_name` variable to `_advclass.dm` so it is possible to have different examine names separate from the name of the advanced class. They're all changed to "Beggar" except for:
Excommunicated is just "Excommunicated"
Exiled Apprentice is just "Exiled Apprentice"
and The Vagabond is just "Vagabond"

## Testing Evidence

image

## Why It's Good For The Game

After playing Vagabond a bit, and if you aren't wearing anything but rags, most assume you are an adventurer or refugee, and not someone that should be well known in the town as low-born scum dirt. Despite the title 'Beggar' not completely applying to every character or sub-role, I feel it's still a fitting title given how most would view them, and hits a lot harder than 'Vagabond' which feels too official and close to bandit. 

Vagabonds aren't 'wanderers'. They're from the town.

Plus it's a classic.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/youngfolk/orphan.dm
+++ b/code/modules/jobs/job_types/roguetown/youngfolk/orphan.dm
@@ -1,5 +1,5 @@
 /datum/job/roguetown/orphan
-	title = "Beggar"
+	title = "Vagabond"
 	flag = ORPHAN
 	department_flag = YOUNGFOLK
 	faction = "Station"
@@ -17,6 +17,7 @@
 	min_pq = -30
 	max_pq = null
 	round_contrib_points = 2
+	advjob_examine = TRUE
 
 	cmode_music = 'sound/music/combat_bum.ogg'
 	job_subclasses = list(
--- a/code/modules/jobs/job_types/roguetown/youngfolk/vagabond.dm
+++ b/code/modules/jobs/job_types/roguetown/youngfolk/vagabond.dm
@@ -6,7 +6,7 @@
 	allowed_races = RACES_ALL_KINDS
 
 	advclass_cat_rolls = list(CTAG_VAGABOND = 20)
-	advjob_examine = FALSE
+	advjob_examine = TRUE
 	always_show_on_latechoices = TRUE
 	job_reopens_slots_on_death = TRUE
 	same_job_respawn_delay = 10 SECONDS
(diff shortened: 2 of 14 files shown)


code/modules/jobs/job_types/roguetown/youngfolk/vagabond.dm
```diff
@@ -6,7 +6,7 @@
 	allowed_races = RACES_ALL_KINDS
 
 	advclass_cat_rolls = list(CTAG_VAGABOND = 20)
-	wanderer_examine = TRUE
+	wanderer_examine = FALSE
```
ProbablyCarl: False by default. Remove this line and give it another go. You'll see it works identically. No need to go false, when that's the normal.

code/modules/jobs/job_types/roguetown/youngfolk/vagabond.dm
```diff
@@ -6,7 +6,7 @@
 	allowed_races = RACES_ALL_KINDS
 
 	advclass_cat_rolls = list(CTAG_VAGABOND = 20)
-	wanderer_examine = FALSE
+	wanderer_examine = TRUE
```
ProbablyCarl: My bad, I should've been more clear. This line altogether, line 9, can be removed. Try compiling and testing without it present and you'll see the change.

yupup123: Pfft. Oh. Alrighty. Done!

code/modules/jobs/job_types/roguetown/youngfolk/orphan.dm
```diff
@@ -1,5 +1,5 @@
 /datum/job/roguetown/orphan
-	title = "Vagabond"
+	title = "Beggar"
```
ProbablyCarl: Apologies for the delay on getting back to this.
Name change necessitates you to to edit line 152 of jobs.dm, and line 107 of landmarks.dm

Maybe a few others I'm missing, too. Make sure your latespawns work and job listing stuff isn't screwed up via local testing afterwards.

yupup123: No worries.
So my intention with this wasn't to change the entire role's name to Beggar, just the examine text, but if you want I can definitely do that.

ProbablyCarl: By changing the title, you'd changed the role's spawn locations to latespawn exclusive, and the role job listing stuff, I'm afraid. Just ctrl+f "Vagabond" and you'll see all instances to replace with "Beggar".

PR #428, opened 2025-12-15
Boss Dragon Fogwall

ProbablyCarl:
## About The Pull Request
Implements a fogwall for the void dragon boss. Increases the health by a good bit. Has it drop a key to dissipate the fogwall. Removes three of the six(6!!!!) dragon rings. Zezuz Pyst. Places unbreakable walls around the surrounding areas to assure you can't just mine in and fire at it from the north, either.

## Testing Evidence
It just works. I tested but forgot to take images. Carl slop PR classic.

## Why It's Good For The Game
Prevents a serious exploit of just running past this awful thing. Fight it, cowards.

PR #429, opened 2025-12-15
Replaces all improper cogs with bronze cogs

Jamdrawers:
## About The Pull Request

there are lots of item/roguetown/cog's around, when they SHOULD be item/roguetown/cog/bronze. The former is like a dummy item that isn't supposed to exist in-game

## Testing Evidence
image

## Why It's Good For The Game

Fixes a bug.


ProbablyCarl:
Please add the fogwall, as per #428, to the dragon's lair. Otherwise, this is a gem PR.

Jamdrawers:
image
Should be all good.

PR #430, opened 2025-12-15
Antipope

ProbablyCarl:
## About The Pull Request
Adds the dreaded antipope, in the form of a role called the 'Doomsayer' for Wretch:
 - Has EVIL sermons, which are just Bishop sermons but for Inhumen. Debuffs Tennites, inversely to normal sermons.
 - Inverse priest statline. STR, as opposed to INT. CON/SPD, rather than WIL.
 - Heretic Seer usage, previously unused, now given to them. Can find their flock.
 - T1 miracles only. T2 spells, with 3 spellpoints.
 - Ritualist.
 - Resonance Caster, because it's funny and they'll hurt their undead companions. They're a charlatan, after all.
 - One slot. Think of it as a cult leader. That's more-or-less what it is.
 - Inhumen exclusive. Auto swaps the user to Zizo if they're not already Inhumen.
 - No bounty. A unique quirk shared with licker.
 - Zizite Cross lootbox, for the purposes of torture and his sermons. Any Wretch can get these, but it's fun and soulful.

Just a lot of silly stuff specifically meant to position them as an odd RP class. ~~This entire thing was done as a joke, but I see genuine potential for RP.~~ No longer a joke. Is it a good idea? No. Is it funny? Yeah, it is. I'll die on this hill.

## Testing Evidence
image
image

## Why It's Good For The Game
Funny content. Content, content, content.


Murderfoxen:
I really need this. I need this a lot. Why? Because I want to be evil pope. I want to do evil. I want to teach evil in my church of evil where I scold graggites for not finishing their kidneys and baothans for wasting their drugs.

GravyMaker:
Oh hey, good to see this suggestion manifesting... Let's see about trying it out.

Metro-Dweller:
This shit needs a buff. Truthfully toss out the magic shit and make the class purely faith focused like an actual priest. Maybe give it the same mechanics like actually priest with choosing their miracles and t4? If nothing else it needs more 250 faith and acting like its a lunacy embracer monk with 3-6 t2 spellpoints.

PR #431, opened 2025-12-15
Sexcon Reagents

ProbablyCarl:
## About The Pull Request
I hate you all.
Why would you make me do this?

In any case, I've cleaned it up some. The separation of the reagents was done in #158. This nabs the work otherwise from there, with a bit of an exception, as with that PR, this is a forward porting of OldRW content. I'm not sure how different it was on OldRW, but that PR even left in old comments, so it might've just been a genuine 1:1. I'll credit it, anyhow.

You know what this is. I'm not going to describe it. It's bad enough that I need to PR this.

## Testing Evidence
image

## Why It's Good For The Game
It really isn't.


Sirgir:
this will save ratwood, thank you carl

PR #436, opened 2025-12-16
cleans up an unused and broken curtain variant

Jamdrawers:
## About The Pull Request

There were some invisible fake curtains based on an old version of curtains that have since been deleted

## Testing Evidence

Hard to screenshot this

## Why It's Good For The Game

bit of cleaning


ProbablyCarl:
CentCom has a few of them, prompting an issue from the compiler.

PR #439, opened 2025-12-18
Cleric Touchup | Templar Bolt | Keeper Nerf

ProbablyCarl:
## About The Pull Request
Most contentious change out of the way, regarding keeper:
 - They've lost grab immunity.
 - Expert knives dropped to journeyman.
 - Wrestling given at novice.
 - Skills otherwise sorted.
 - Nobility no longer permitted. They're of a Pestran sect, not a house.

Why did they have grab immunity? They already had next to no strength, but they're a genuine pain to deal with when doing 90% of the stuff they do. You know who you are. Not so much a targeted PR though as it is just correcting something kinda odd and nonsensical.

Now, Templars:
 - Divine Blast given, as they're the direct opposition to the keep's knights at times in purpose, though not backed by MAA, nor given master weapons skill or plate. Only Acolytes, Missionaries and the Priest got this before. Now, Templars. It's flavorful I'd argue, though we can remove this if it's over the top.

Finally, Cleric changes:
 - For Monk:
    - ~~Monk lost its +1SPD~~, in exchange for +1WIL. 
    - +2STR exchanged for +2SPD.
    - Unarmed loadout given pugilist gloves, exclusive to both them and disciple.
    - All loadouts, with the exception of staves, given ~~crit resist~~ dodge expert. ~~This is a replacement of their dodge expert.~~
    - Staves loadout loses ~~crit resist~~ dodge expert, in exchange for ~~dodge expert~~ crit resist, ~~-2STR and +2SPD~~ -2SPD and +2STR. ~~This is for folks who want to play old dodge expert monk, I guess, but it's so poor by comparison that you're just gimping yourself.~~
    - Monk regen dropped from 0.5 to 0.3, on devotion.

 - For Cantor, their regen has been raised from 0.3 to 0.5, on devotion.
 - Missionary given T4 miracles, locked to T3 devotion limit, with their 0.5 devotion regen. ~~Additionally, they're now given ritualist. The only non-antag/non-Church to have it~~. At T3 miracles, sharing the slot with keeper who can just raise above it anyways, there was literally no reason to play them. You were worse than a vagabond in a lot of ways, even. Genuinely awful, compared to old Missionary. Why AP nuked them I've no idea, but they were fun and soulful. ~~As for ritualist, I'll remove it if we get Zizite murderrejects who rituoso themselves. God, I hate rituoso so much. Maybe I'll lock ritualist to Tennite for this? I 'unno.~~ Can you guess what happened with ritualist?

Also snuck in Churchlings not getting nobility. Why could they be nobility? Nonsense.
## Testing Evidence
image
image

## Why It's Good For The Game
Monk being a dodge expert without the speed was kinda odd. Just nonsensical and weirdly done.
Now there's at least a reason for it, I guess.

Missionary used to be super fun and flavorful. The only unaligned T4 caster, who could be an agent, their own entity, etc. Whatever. T3 restricted you from the major fun stuff, even though you were literally useless in every other regard. Maybe it wasn't an AP change that nuked them, and it was SR who buffed them? Either way, it was fun. Genuinely really great to play into. The only concern I see with this is a very specific group using it to mass rituoso themselves, but they already play borderline valid characters with how obnoxious they are.

Cantor just blows.

Templars aren't intended to be knight levels of strength, but they lack in comparison as a 3 slot role to the 8 MAAs, 2 knights and a KC, on top of all the mercs who gravitated towards the keep. Not counting the Martyr, but they're also a one trick sort of deal, and their skills / stats aren't even comparable to knights, who blow them out of the water even before the master skill change. Even the average merc has skills that best the Martyr without them popping the sword.

As for Keeper, _WHY_ did it have grab immunity? It's already locked to a ceiling of six strength max. Come on... It already has basically every single skill trait, every single possible skill it'd need at novice, etc. It's absolutely absurd as is for a non-combat class. Reminds me of omnitowner.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/combat/cleric.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/combat/cleric.dm
@@ -8,12 +8,11 @@
 	category_tags = list(CTAG_ADVENTURER, CTAG_COURTAGENT)
 	class_select_category = CLASS_CAT_CLERIC
 	subclass_social_rank = SOCIAL_RANK_YEOMAN
-	traits_applied = list(TRAIT_DODGEEXPERT, TRAIT_CIVILIZEDBARBARIAN, TRAIT_OUTLANDER)
+	traits_applied = list(TRAIT_CIVILIZEDBARBARIAN, TRAIT_OUTLANDER)
 	subclass_stats = list(
+		STATKEY_WIL = 3,
 		STATKEY_STR = 2,
-		STATKEY_WIL = 2,
 		STATKEY_CON = 2,
-		STATKEY_SPD = 1, //Base of +9, over the standard +7. Special clemency given to the Monk, as their playstyle is exceedingly lethal - light-to-no armor, while specializing in a dangerous melee style.
 	)
 	subclass_skills = list(
 		/datum/skill/combat/wrestling = SKILL_LEVEL_JOURNEYMAN,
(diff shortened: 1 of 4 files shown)


Sirgir:
Seconding this on the Quarterstaff, it feels a little counter-intuitive losing out on STR for SPD + dodge expert on a weapon with one of the highest defense values in the game. (Stave weapons are also often very weak due to the fact they're Blunt, so losing out on STR really hurts you even more, fuck blunt is bad!!!!!!). Feels like it should just follow the pattern of the other loadouts.

Now if staves were a swift weapon, it'd be a different story, I think.

Pidgey-chan:
The church doesn't need to compete with the keep in capacity for violence; Templars were better than knights basically entirely before knights got a master skill and plate.

Pidgey-chan:
> > The church doesn't need to compete with the keep in capacity for violence; Templars were better than knights basically entirely before knights got a master skill and plate.
> 
> Church is the only naturally opposing force to keep outside of direct antags (which are a generalized opposition rather then a specialized one)
> 
> Even then: templar getting divine bolt won't make them equal. It'll just give them **something** - for generalized threats, including the keep.

Actually they're the medbay

PR #445, opened 2025-12-20
Mercs Use Wretch Slot Setup

ProbablyCarl:
## About The Pull Request
Mercenaries now make use of the wretch slot setup, so they scale with pop. As opposed to eight roundstart, in an eleven player round.

They start at 3 slots. For every ten players above a count of twenty, it goes up by one, back to the original 8 slots. Is this harsh? Yeah. But it should prevent eight man merc deathstacks who outbid the keep's own roundstart.

I would be more harsh with this, but this is going to have enough pushback as is from the grenzel stacks and folks who play nothing but Ruma / Shaman / whatever to get their meta setup roundstart.

This also raises merc PQ requirement from 2 to 12. Play towner or something. Don't jump into your meta class and then die roundstart when you've no understanding of game mechanics. Thanks.

## Testing Evidence
Identical code setup to Wretch/Bandit. Nothing tested locally because it should just work, and I can't test a dozen dummy accounts locally as is.

## Why It's Good For The Game
We just had a round where out of the starting pop, none of them were keep, a handful were adventurers, and the rest were mercenaries exclusively. A full merc goon squad stack.

### DIFF
--- a/code/controllers/subsystem/migrants.dm
+++ b/code/controllers/subsystem/migrants.dm
@@ -258,6 +258,7 @@ SUBSYSTEM_DEF(migrants)
 	GLOB.joined_player_list += character.ckey
 	update_wretch_slots()
 	update_bandit_slots()
+	update_merc_slots()
 	if(character.client)
 		character.client.update_ooc_verb_visibility()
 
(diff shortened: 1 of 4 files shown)


vigeta:
I never thought I'd see the day. The cap rising with the pop is both a smart change and a fine compromise. Far too many rounds have been played in which the ENTIRE keep is empty and yet we have 4 mercs 4 advs etc. People who claim pop will disappear fail to realize the players who would depart over such a change are generally selfish and not really needed in the first place.

ProbablyCarl:
I can't be bothered.

PR #446, opened 2025-12-22
Levy: Towner Conscripts

Guidesu:
## About The Pull Request
A towner class limited to three slots with minimum garrison and keep access. 
image
image

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/villager.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/villager.dm
@@ -39,7 +39,8 @@
 		/datum/advclass/seamstress,
 		/datum/advclass/thug,
 		/datum/advclass/witch,
-		/datum/advclass/woodworker
+		/datum/advclass/woodworker,
+		/datum/advclass/pilgrim/levy
 	)
 
 /*


Jamdrawers:
I wonder if this concept would work better as like... something akin to Court Agent, where an existing role can be used rather than having a bunch of juiced up post-spawn options that overlap with all the existing towner roles. Though also wonder if it's a good concept or if it really works here to begin with. Can get kind of a similar result just going with a weapons virtue for 3 in a combat stat like spears - though even then that's not really necessary with how many towner roles already get 3 in spears (soiler) axes (woodsman) or maces (miner).

Thematically I like the idea of getting some noncombat skills as a recruit because your character was presumably doing stuff before they got recruited, but I do find when playing Warden (which is a bit similar in some ways) it can be easy to get distracted by your butchering skill and get lost making fat stacks out of meat, furs and hides and forget to actually do your job defending the town and interacting with other players. Can worry about the same happening but with farming or whatever - like it'd feel thematically appropriate for your farmboy-pressed-into-service to still be capable of farming, but I kinda wouldn't want them doing much farming while they're supposed to be being pressed into service. Being pressed into service is also a little tricky to justify in the game given the proportion of any given round that's already a combatant.

I'd also say to not give them Manor keys - the grubby peasant's jurisdiction ought be out in the town. Wardens don't get manor keys - the manor isn't their job.

Guidesu:
> I wonder if this concept would work better as like... something akin to Court Agent, where an existing role can be used rather than having a bunch of juiced up post-spawn options that overlap with all the existing towner roles. Though also wonder if it's a good concept or if it really works here to begin with. Can get kind of a similar result just going with a weapons virtue for 3 in a combat stat like spears - though even then that's not really necessary with how many towner roles already get 3 in spears (soiler) axes (woodsman) or maces (miner).
> 
> Thematically I like the idea of getting some noncombat skills as a recruit because your character was presumably doing stuff before they got recruited, but I do find when playing Warden (which is a bit similar in some ways) it can be easy to get distracted by your butchering skill and get lost making fat stacks out of meat, furs and hides and forget to actually do your job defending the town and interacting with other players. Can worry about the same happening but with farming or whatever - like it'd feel thematically appropriate for your farmboy-pressed-into-service to still be capable of farming, but I kinda wouldn't want them doing much farming while they're supposed to be being pressed into service. Being pressed into service is also a little tricky to justify in the game given the proportion of any given round that's already a combatant.
> 
> I'd also say to not give them Manor keys - the grubby peasant's jurisdiction ought be out in the town. Wardens don't get manor keys - the manor isn't their job.

Good ideas

PR #447, opened 2025-12-22
The Rookie

Jamdrawers:
## About The Pull Request

I saw https://github.com/Rotwood-Vale/Ratwood-2.0/pull/446 and was reminded that I had had an idea for something like that, albeit taking it from a slightly different angle. It's a new role that functions similarly to the Squire except it serves the MAA in town rather than the knights in the keep - they have a bit of extra out-of-combat skill utility from their time before they  were pressed into service (based on Guidesu's idea but toned down across the board), they fix dents in their superior's armor, and they're a spare pair of hands. But they aren't the peerless highly trained warriors that the MAA are. Not yet at least.
image

Leaving as draft 'til I finish descriptions and iron things out

## Testing Evidence

image

## Why It's Good For The Game

I think it's a cute archetype, and the guard does always benefit from a spare pair of hands to run messages and such.

### DIFF
--- a/code/__DEFINES/jobs.dm
+++ b/code/__DEFINES/jobs.dm
@@ -114,6 +114,7 @@
 #define MANATARMS	(1<<1)
 #define DUNGEONEER	(1<<2)
 #define SQUIRE		(1<<3)
+#define ROOKIE		(1<<3)
 #define BOGGUARD	(1<<4)
 #define SERGEANT	(1<<5)
 #define SHERIFF		(1<<6)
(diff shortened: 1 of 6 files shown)


Pidgey-chan:
What's the distinction from a squire here beyond attaching to an MAA? Couldn't it be some sort of subclass?

Sirgir:
Not sure if we need another garrison role (though this definitely seems like a far better role for newer players versus say squire). But if it were between https://github.com/Rotwood-Vale/Ratwood-2.0/pull/446 and this, I'd definitely go with this.

Troontjelolo:
Sovl, but better as an MAA subclass

PR #449, opened 2025-12-22
Blocks bandits from entering procedural dungeon

Heheniy:
## About The Pull Request
see name

## Testing Evidence
image

## Why It's Good For The Game
"Guess today i will rob this ancient tomb instead of local duke"


Pidgey-chan:
Adventurers are valid targets for robbery, even if they're in a dungeon.

CVSthePharmacy:
Don't merge this. People will probably abuse the dungeon protection... think of it like looting a high risk area where pvp is allowed. You're going there as an adventurer or bandit with the very real risk of getting your neck torn into bits of bone and tissue by another sentient being

PR #450, opened 2025-12-22
Add scissors recipe for iron and steel

chrapacz2000:
.


RadiantFlash7:
If I recall corectly, they exist still, they're just under engineering.

InspectorLestrade:
> If I recall corectly, they exist still, they're just under engineering.

You recall incorrectly.

PR #459, opened 2025-12-27
Armor Recipe Changes

InspectorLestrade:
## About The Pull Request

The intention of this pull request is to change armor recipes for cuirasses, half-plate, and full plate to mirror that of the Psydonic Silver Armor recipes and to effectively create an 'upgrade' system allowing for people to have the option of augmenting, rather than replacing their armor as a given round plays out.

## Testing Evidence

(https://youtu.be/yUZ4-nmnGFk)

The video there is a touch long, but it shows that we tested every instance we could think of where a Cuirass may upgrade into Half Plates and Full Plates (with the exception of Fluted Plates, we forgot to check those at the time.)

## Why It's Good For The Game

I feel like this is good because it introduced a standardized system for this armor 'family' (cuirass, half plate, and full plate) as well as encouraging people to hold onto things that could be useful later in the round rather than needing to throw it away the second they have a 'better' armor option available to them.


SpartanOps08:
Yeah, this makes sense!

PR #460, opened 2025-12-27
Smelter Amount Edits

InspectorLestrade:
## About The Pull Request

This PR is meant to address the amount of bars received from smelting recipes. I felt that as they are now, they are not very good and do not make for a good experience for the smith, or the smith's customers.

## Testing Evidence

(https://youtu.be/lIhV6kts-Gw)

## Why It's Good For The Game

I will argue for each of these buffs in turn.

Steel to four - This is how it was back in Old Ratwood, and in many places besides. Reducing it to 3 is one of those things that have happened as a result of Re-basing to AP's code. 4 ingots is a useful amount to work with in one blow, 3 is not so useful. You are rarely going to need 'odd' amounts of bars, and you will often need 'even' amounts given how many objects require two bars, or how many single object recipes someone wants to buy at once.

Ancient Alloy to four - This was previously set at two, and the number one reason people ordered this was to have Artificer Half-Plate engineered for them. This requires three bars meaning that you must go through the process of producing Ancient Alloy _twice_ if you want to produce a single item. This is immensely annoying, and time consuming. Not to mention it drives up the costs for ancient alloy to the point that players will not want to buy objects made from this material.

Blacksteel to two - Long, long ago back in Stonekeep it was four bars of blacksteel from a smelt. I am not saying we should go back to this, it would make blacksteel **too** plentiful. However, right now it is impossible to get a workable mount of blacksteel or offer it at a price that is good for the average player and good for the players running the smithy. A single bar is also not useful for making equipment, and can basically only be used as fodder to make treasures with rather than items that add to a player's progression. Changing it to two is my notion of a compromise between the ancient plentiful blacksteel, and keeping it as is where it's so rare, and so controlled that there's not enough of it to go around.

### DIFF
--- a/code/modules/roguetown/roguejobs/blacksmith/smelter.dm
+++ b/code/modules/roguetown/roguejobs/blacksmith/smelter.dm
@@ -247,15 +247,15 @@
 			blacksteelalloy = blacksteelalloy + 2
 
 	if(steelalloycoal == 1 && steelalloyiron == 3)
-		max_contained_items = 3
-		alloy = /obj/item/ingot/steel
+		max_contained_items = 4
+		alloy = /obj/item/ingot/steel // 3 bars is an awkward number to work with, considering most people order things that need 2 bars
 	else if(bronzealloy == 7)
 		alloy = /obj/item/ingot/bronze
 	else if(purifiedalloy == 11)
-		max_contained_items = 2
-		alloy = /obj/item/ingot/purifiedaalloy // 3 aaslag, 1 gold, makes 2 purified alloy.
+		max_contained_items = 4
+		alloy = /obj/item/ingot/purifiedaalloy // 3 aaslag, 1 gold, makes 4 purified alloy, 2 is too few to be worth doing anything with
 	else if(blacksteelalloy == 7)
-		max_contained_items = 1 // Blacksteel is supposed to be rare and inefficient. 3 steel and 1 silver into one.
+		max_contained_items = 2 // 1 ingot is too few to do anything meaningful with and makes prices too high to sell anything
 		alloy = /obj/item/ingot/blacksteel
 	else
 		alloy = null


john-pawson:
>This is how it was back in Old Ratwood, and in many places besides. Reducing it to 3 is one of those things that have happened as a result of Re-basing to AP's code. 

I would like to point out this wasn't done due to a rebase to AP, the change was applied on oldRW too, since it only needed 3 iron bars and 1 coal to produce, it meant that the coal was somehow creating more metal. It also meant that buying 3 pieces of iron and then selling 4 pieces of steel would always make a profit, due to the extra bar *and* the higher sellprice of steel. 

Not only that but with this amount seeing smiths with piles of steel that wasn't used by the end of the round was way more common, usually having more than enough for the entire week by halfway through the round. And I don't think "there's a lot of even numbered recipes" is that good of a reason to raise it back up.

Edit: Also, consider lowering the sellprice of purified alloy, it sells for 111 which means 444 mammon of profit for ~130 mammon worth of materials

Yurek1:
> > This is how it was back in Old Ratwood, and in many places besides. Reducing it to 3 is one of those things that have happened as a result of Re-basing to AP's code.
> 
> I would like to point out this wasn't done due to a rebase to AP, the change was applied on oldRW too, since it only needed 3 iron bars and 1 coal to produce, it meant that the coal was somehow creating more metal. It also meant that buying 3 pieces of iron and then selling 4 pieces of steel would always make a profit, due to the extra bar _and_ the higher sellprice of steel.

no smith in their right mind is going to sell their steel bars wholesale. that's absolute lunacy- they're going to turn it into things to sell, like swords and armor. this is entirely irrelevant, except for the fact that an even number of bars is easier to work with for smiths.

> 
> Not only that but with this amount seeing smiths with piles of steel that wasn't used by the end of the round was way more common, usually having more than enough for the entire week by halfway through the round. And I don't think "there's a lot of even numbered recipes" is that good of a reason to raise it back up.

smiths here have the _opposite_ issue. they run out of steel like you wouldn't believe, due to how hard it can be to get ore.
also, it absolutely is a good reason to raise it back up, because now they're not gonna be left with an odd stray bar that they can't do anything with.

> 
> Edit: Also, consider lowering the sellprice of purified alloy, it sells for 111 which means 444 mammon of profit for ~130 mammon worth of materials

nobody cares about the sell price of the bars themselves. there is MUCH easier ways to make money as a smith besides sacrificing your (relatively rare) gold ore to sell ancient alloy bars wholesale. they're going to smith it into statues, or jewlery, or something else that's actually worth money.

Jamdrawers:
I disagree with change 1 - it's 3 units of iron for 3 units of steel. The coal is there to treat the steel.

Yurek1:
> That and, in my own experience playing smith, there's enough steel so that you usually don't run out of it, there'll be times that you need to go ahead and wait a bit for a new batch to finish but as long as you keep the smelter running you're gonna have enough. There's no need for more.

are you sure you've played smith on 2.0? because unless you play exclusively on dead pop hours, running out of steel is a _consistent_ and _frequent_ issue for smiths. even if you run that shit 24/7, you'll get knights who roll in and want plate EVERYTHING, or you'll have a line out the door.

> 
> And the steel bar selling mostly applies to things like merchants, stewards and the such buying a riddle and smelting iron into steel as a money farming method.

they can't even get above apprentice in smelting without the virtue so why is this an issue? most merchants are exclusively interested in jewlery anyway because the profit margins on steel are so fucking low by selling bars in bulk. this isn't an issue.

Picklebootsjr:
This is so much nicer to work with now, as you can actually afford to be a blacksmith and get more metal to work with to help the vale. and distribute much much easier.

PR #468, opened 2025-12-28
Duchy Colors and Dye Bin Changes

yupup123:
## About The Pull Request

- Adds a new Throat command: "Change Colors". Lets the current Regent change the duchy colors.
- Adds all of the dye bin colors and removes all restrictions on duchy color selection.
- Fixes the bugged guard tabards on Rockhill and Dun World that had incorrect colors. Changes them to the Retinue Tabards that actually work.
- Makes dye bins update the preview color immediately after changing it.
- Makes dye bins immediately eject the item after using it.
- Adds 'Ducal Scheme' option at dye bins, automatically changing the color to whatever the duchy colors are, even if they are changed.

## Testing Evidence

image

## Why It's Good For The Game

- Sometimes a new flag must be raised for when a Duchy goes through a major change.
- I see no reason as to why the duchy color selection has to be so restrictive.
- Rejoice tabard users.
- Using dye bins was kind of clunky. This just removes a pointless button click.

### DIFF
--- a/code/__HELPERS/lordcolor.dm
+++ b/code/__HELPERS/lordcolor.dm
@@ -45,3 +45,15 @@ GLOBAL_VAR(lordsecondary)
 		O.lordcolor(GLOB.lordprimary,GLOB.lordsecondary)
 	for(var/turf/T in GLOB.lordcolor)
 		T.lordcolor(GLOB.lordprimary,GLOB.lordsecondary)
+
+// Ducal Scheme lordcolor implementation for dyed items
+/obj/item/lordcolor(primary, secondary)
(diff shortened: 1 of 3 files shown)


InspectorLestrade:
This seems like a very good change. Would it be possible to add a function to dye bins that is 'Ducsl Scheme' which colors an item with the colors that are current set by the Duke?

yupup123:
> Hey can we remove the auto-eject on coloring? If you have more than one color, you have to re-insert the object every time.

Yeah I meant to do that but it was quickly merged before I could get to it.

PR #472, opened 2025-12-29
Ports Offer and point QoL

yupup123:
## About The Pull Request

Ports everything from https://github.com/Azure-Peak/Azure-Peak/pull/4392 : 

> - Offering an item to someone now makes a physical clickable effect appear in the gamespace, making it easier to see when it happens.
> - Sneaking while offering an item to someone will now make it a sneaky offer, visible only to you and the person you're offering it to.
> - Pointing to an item held on your person will now make a speech bubble appear.

## Testing Evidence
It works in local testing but who knows what it might break in the grand scheme of things.

Screenshot 2025-12-29 132603

Screenshot 2025-12-29 132118

## Why It's Good For The Game

Quality of life good. These two things are specifically hard to notice, so a visual for them is nice.

### DIFF
--- a/code/modules/mob/point.dm
+++ b/code/modules/mob/point.dm
@@ -13,6 +13,16 @@
  *
  * overridden here and in /mob/dead/observer for different point span classes and sanity checks
  */
+/mob/verb/pointed(atom/pointed_atom as mob|obj|turf in view())
+	set name = "Point To"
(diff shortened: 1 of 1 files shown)


ProbablyCarl:
Move this to the emote tab. No need for an IC tab return.

PR #473, opened 2025-12-30
Throwing Buffs & Changes

InspectorLestrade:
## About The Pull Request

This PR is intended to beef up several throwing weapons a little bit (Tossblades, Hurlbats, and Javelins) while also ensuring they maintain distinct identities with clear use cases and drawbacks compared to each other. This is also why it increases the number of tossblades to five (that you can reload a tossblade from empty with one ingot) and increases the number of hurlbats to two (that you can fill a bandolier with two ingots).

This is because that Javelins were hardly any better than tossblades, while also being harder to utilize and carry (what with the expectation being you sacrifice a hip or back slot for their quiver).

## Testing Evidence

https://youtu.be/zfxMotWTdco

## Why It's Good For The Game

I feel this is good for the game because it creates more variance among the throwing weapons and encourages them to be used more often than as a gimmick, or because you just happened to spawn with them as part of your kit.


InspectorLestrade:
We have something that shouldn't BE there in this PR and it's going to be re-opened when it's debugged.

PR #474, opened 2025-12-30
Throwing Rebalance Try 2

InspectorLestrade:
## About The Pull Request

This PR is intended to beef up several throwing weapons a little bit (Tossblades, Hurlbats, and Javelins) while also ensuring they maintain distinct identities with clear use cases and drawbacks compared to each other. This is also why it increases the number of tossblades to five (that you can reload a tossblade from empty with one ingot) and increases the number of hurlbats to two (that you can fill a bandolier with two ingots).

This is because that Javelins were hardly any better than tossblades, while also being harder to utilize and carry (what with the expectation being you sacrifice a hip or back slot for their quiver).

## Testing Evidence

https://youtu.be/zfxMotWTdco

## Why It's Good For The Game

I feel this is good for the game because it creates more variance among the throwing weapons and encourages them to be used more often than as a gimmick, or because you just happened to spawn with them as part of your kit.

### DIFF
--- a/code/game/objects/items/rogueweapons/melee/axes.dm
+++ b/code/game/objects/items/rogueweapons/melee/axes.dm
@@ -228,7 +228,7 @@
 	grid_height = 64
 	grid_width = 32
 	force = 20
-	throwforce = 32 //You ever had an axe thrown at you?
+	throwforce = 35 // Buffed to not fall behind javelins
 	throw_speed = 6 //Batarangs, baby.
 	max_integrity = 50 //Brittle design, hits hard, breaks quickly.
 	armor_penetration = 40 //On-par with steel tossblades.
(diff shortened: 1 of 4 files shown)


Pidgey-chan:
These are for the most part low right now because throwing weapons target "piercing" defense instead of stab, which is arbitrarily low across the board.

Jamdrawers:
Would be great if 'heavy' thrown weapons had kind of a windup charge time. Kinda has to be balanced around the fact that you can draw and throw them as quick as you can click, which works for daggers but doesn't feel right on axes or javelins. Would definitely be a bigger code job though.

CVSthePharmacy:
> Would be great if 'heavy' thrown weapons had kind of a windup charge time. Kinda has to be balanced around the fact that you can draw and throw them as quick as you can click, which works for daggers but doesn't feel right on axes or javelins. Would definitely be a bigger code job though.

You could always make a throw intent that works as the Bow shoot intent

PR #475, opened 2025-12-30
Bandage soaking

john-pawson:
## About The Pull Request

WARNING: This could, potentially, be busted, or useless, balance-wise. I did my best balancing it on testing but live testing is always the best way to find out unexpected problems. Reccomended TM, careful merge.

This PR adds in a new mechanic: Soaking bandages in medicine! 4 medicines in specific: Health potion, Strong health potion, Aqua Vitae and blessed water.

It works simply! You can use a container with said medicine to soak a bandage on the medicine, when soaked, a bandage will slowly heal the limb and the wounds it has SPECIFICALLY on that limbs, this compared to drinking the potion is more efficient, but slightly more tedious and slower. So if someone has a broken leg or the only damaged bodypart bandaged is an arm, healing that specific limb by using soaked bandages may be better than drinking the potions or constantly using blessings.

Bloodied bandages will NOT apply their medicine, so they're not useful against things such as arteries or wounds that bleed a LOT

image

So far, the stats go as this:

Health potion: Standard, lasts about ~20 seconds on one bandage, and heals slightly more compared to when it is drunk.
Strong health potion: Stronger, still lasts ~20 seconds, but heals twice as fast as the normal health potion. Slightly less efficient used on a bandage than drunk directly, damage-wise.
Aqua vitae: Weaker, slower but heals more than the strong health in total. Lasts 2 minutes healing, half the rate of the health potion. Good for slow, thorough healing.
Blessed water: It's the worst. But, better than nothing. Fifth of the healing rate, lasts ~40 seconds, heals less than all the others medicines, but, it's practically free for anyone with Orison. 

## Testing Evidence
image
image

## Why It's Good For The Game

More ways of healing other than blessings, stitching, sleeping, and chugging an entire bottle of red. This could also be technically considered first aid, since it would be one of the little ways of slowing down bleeding *and* healing it at the same time. Either way, it's also a different way of applying potions! Usually you'd have to drink them and have to wait for your metabolism to do all the work, not only that but it heals your entire body, so it'd heal small scratches with the same priority as an artery. Now, you can focus that medicine on a specific bodypart if you need it.

### DIFF
--- a/code/game/objects/items/rogueitems/natural/clothfibersthorn.dm
+++ b/code/game/objects/items/rogueitems/natural/clothfibersthorn.dm
@@ -280,7 +280,7 @@
 /obj/item/natural/cloth/attackby(obj/item/I, mob/living/user, params)
 	var/obj/item/reagent_containers/C = I
 	if(!istype(C))
-		return
+		return ..()
 	if(C.reagents.has_reagent(/datum/reagent/medicine/healthpot, 10) && !medicine_amount)
 		to_chat(user, span_notice("Soaking the [src] in lyfeblood..."))
 		if(do_after(user, 3 SECONDS, target = src))
@@ -317,7 +317,6 @@
 			desc += " It has been soaked in blessed water."
 			detail_color = "#6a9295"
 			update_icon()
-	. = ..()
 
 /obj/item/natural/cloth/update_icon()
 	cut_overlays()


john-pawson:
> This sounds interesting, and I like it conceptually _but_ it reminds me of pre-bandaging.

Pre-bandaging doesn't work since attacking an already bandaged limb immidiately ruins the bandage, making it useless, and immidiately stopping the medicine from working.

> You mentioned that artery hits basically won't be healed by these, but what about other cuts?

Other cuts will get healed by these, yes. Minor cuts will usually get healed with a single bandage but the more deeper cuts would need 2-3 to get healed fully, the medicine usually runs out before the bandage gets bloodied, so it's not exactly wasted.
On bigger wounds you *may* be better stitching the wound, though if you're bleeding from multiple places it'd give you time to heal the wound passively while you treat other matters, and it is faster than stitching it by hand when you have no medicine skill.

ronfino:
What would be even better is soaking bandages in **poison**, maybe causing rot damage too...

PR #482, opened 2026-01-01
fix for blacksteel swords

RadiantFlash7:
## About The Pull Request
Fixes Blacksteel sword's missing sprites
image
It also increases their integrity to be on par with a longsword

## Why It's Good For The Game
They existed prior, but during the parity swords got moved from 32 to sword32, and these we lost.


InspectorLestrade:
Thank you for this. It was a tragedy the one time I recall someone making a Blacksteel Sword only to receive a busted ass sprite.

PR #495, opened 2026-01-05
language ports

lpeapnni:
## About The Pull Request
ports https://github.com/Azure-Peak/Azure-Peak/pull/4551 and https://github.com/Azure-Peak/Azure-Peak/pull/4712

## Testing Evidence
Capture

image

## Why It's Good For The Game
both of these were ported in https://github.com/Rotwood-Vale/Ratwood-2.0/pull/399 (RIP) and i really dont think that there's any reason that they shouldn't be here


lpeapnni:
#399 is reopened so this lowkirk doesnt need to exist anymore

PR #506, opened 2026-01-07
Adds Wilderness Guide Trait

yupup123:
## About The Pull Request

Adds a new trait called Wilderness Guide, blocking ambushes from occurring within 10 tiles.

Given to druids, but more roles could get it in the future.

## Testing Evidence

image

## Why It's Good For The Game

Gives druids a little bit more of a purpose and use.


Hmeister-real:
give this to bandits please, they trigger ambushes while carrying people even when sneaking

PR #512, opened 2026-01-10
(Bounty) Milking QOL, Deathless.

Guidesu:
## About The Pull Request
If they have blood (≥20 units):

Consumes 20 blood
Produces 10 milk
Continues to next cycle if container has space and still has ≥20 blood
If they have no blood (or <20 units):

Takes 15 burn damage
Produces 10 milk
Continues to next cycle if container has space and they're not dead yet
So it will keep looping, either draining blood or dealing burn damage each time, until either:

The container is full
They die (from burn damage if bloodless)

Deathless people use blood to make milk on demand
People with no blood get burn damage to make milk

Milking sounds.

## Testing Evidence
None....

## Why It's Good For The Game
I- ... don't know but it can be good!

### DIFF
--- a/code/modules/mob/living/carbon/human/milking.dm
+++ b/code/modules/mob/living/carbon/human/milking.dm
@@ -47,7 +47,7 @@
 	else
 		to_chat(user, span_warning("[container] is full."))
 
-// Blood-to-milk conversion for deathless beings (revenants, constructs, etc. with TRAIT_NOHUNGER)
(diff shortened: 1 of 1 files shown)


Jamdrawers:
I feel like construct *should* be excluded from there

ProbablyCarl:
This is weird, but actually something we could have, I suppose.
Could you see to preventing constructs?

PR #532, opened 2026-01-12
ports kazengun wretches

Rin053:
## About The Pull Request

This is basically a direct port of [Kazengun wretches from SR.](https://github.com/Scarlet-Reach/Scarlet-Reach/pull/912)

I kept everything the same except I added the bamboo hat to the merc classes as well as the funny ability to let your ears hang out with the bamboo hat using RMB.

## Testing Evidence

image

## Why It's Good For The Game

More evil things!

### DIFF
--- a/code/game/objects/items/rogueweapons/melee/blunt.dm
+++ b/code/game/objects/items/rogueweapons/melee/blunt.dm
@@ -67,10 +67,11 @@
 	icon_state = "instab"
 	reach = 2
 	clickcd = CLICK_CD_CHARGED
-	recovery = 10
+	recovery = 30
+	penfactor = 25
+	damfactor = 0.9
 	warnie = "mobwarning"
 	hitsound = list('sound/combat/hits/bladed/genstab (1).ogg', 'sound/combat/hits/bladed/genstab (2).ogg', 'sound/combat/hits/bladed/genstab (3).ogg')
-	penfactor = 40 //Same as a dagger.
 	item_d_type = "stab"
 
 //blunt objs ฅ^•ﻌ•^ฅ


Skyglazer:
Adding entirely new classes and buffing existing classes (which are already strong) should probably be separate PRs.

anconfuzedrock:
Ok, other side of it is all of these range from being the strongest existing wretches to being fucking nuts. you're porting it from a server with different balance philosophies. 
From least to most ridiculous:

even wretch knights don't get master weapon skills. the only ones that do are masked lunatic and berserker, and berserker is probably the strongest wretch. both of them only get unarmed skill, which struggles against anyone with armor.
"Disgraced Custodian" having master polearms is a bit ridiculous and probably oppressive, especially when they get a boar spear. only +1 strength is okay, though. this one might be fine.

Sword hunter gets master swords and dodge expert?? You can say its stats aren't great, but it's only 2 less stat points than a wretch outlaw, which only gets expert swords. And one sword they can choose is INSANE. A greatsword with swift is crazy. at least you removed the peel from it.

I thought a wretch with +4 str at the cost of journeyman weapon skills was really sketchy. Then I saw "Onimusha" still has expert wrestling. No! Then I saw they get UNCAPPED STRENGTH for some reason. NO!!! Even Berserkers don't get that! And calling the stats weighted towards 9 isn't considering how much better these stats are. Low perception was harmful in combat for berserkers; -2 int doesn't hurt them Onimusha in combat, and they even get basic medicine and crafting skills to work around it. -2 Fortune can be bad but can be ignored by having good mood. They might be discouraged from taking struggler now, but Thuggish is another +2 strength statpack with a fortune bonus, and it's especially appealing when you could have an uncapped 16 str (or 17 on a Drakian)

Also, it looks like you removed the peel from the sword itself, but the peel is still in the code. Please remove like- actually unused code from your ports. Makes me wonder if there's any more added bloat I'm missing.

ProbablyCarl:
We already have padded hoods. Just make them subtypes.
Additionally, a buff to Ruma? Really? One of the most absurd, out of this world merc classes as is?
Adding armour values to the cloth pants and sandals?
Buffing the naginata? A bunch of misc weapon changes?

That's not even touching on the Wretch classes, which are batshit insane.
+4STR from class, +2STR from statpack, +1STR from species, for a potential of 17STR roundstart, with STRENGTH UNBOUND. So not only do they immediately hit the softcap like brigands, _unlike brigands_, they bypass it outright, and can jump themselves further with potions. Combined with crit resist and immunity to painstuns. Oh, and master athletics, for some reason? And STRONG BITE? Wild.

Not touching the others, given they're not nearly as bad, but they're also wildly out there. Expert athletics and master swordsman, with something like a roundstart rapier / guaranteed DWielder setup too? The other one, with master polearms, is just objectively worse freifechter, I suppose, but that immunity to pain stun does more than you think.

One slot doesn't matter when it's one of EACH, and these all wildly blow everything else out of the water. Even licker would struggle to keep up with that bruiser that just bypasses the soft cap as is, lol.

We are not SR. I like to at least pretend we've a modicum of sense when it comes to the absurd stuff I introduce. Given we don't have stuff like instakill chems, for example... Nor do we have the same combat now, far as I'm aware?

I cannot in good faith even consider this, after seeing that one wretch setup, and then on top of that, a bunch of misc, unrelated stuff to the PR as is.

GravyMaker:
So, I'm not really a balancehead in the way that Carl is - I'm much more of a tech guy. With that said, as Skyglazer notes, you do mingle a smattering of seemingly-unrelated balance changes in among the actual role additions - I'd strongly recommend keeping these separate, unless there's an articulable reason that you feel they need to come alongside this role, that isn't just making sure this specific role isn't made too powerful by having pre-existing equipment, or too weak by being vulnerable to said equipment.

I can't give such a detailed breakdown on Carl's comments, though I will offer a little more insight into the cymbal-banging monkey that lives in the heads of maintainers. Balance is a pretty contentious topic - and it's never _more_ contentious than when we find that we have to nerf something. I get that - it really sucks when a role you enjoy is nerfed, no matter how good it is for the game at large. But for this reason we need to be especially careful about the state that roles are added in, to make sure they're not going to be _too_ potent, especially in the hands of a capable player.

In particular, Ratwood does not have the player count that SR has, and nor does it have many other features present in SR, for various reasons - that means that content typically shouldn't be ported 1:1, otherwise it will feel very overtuned here.

So with that in mind, I'd recommend reading and responding to what Carl says - he may sound a bit harsh at times, but hopefully you can look past that and understand that he's just trying to help steer you into a healthier direction for this addition. Try looking at existing roles you want to be comparable to the one you're making statistically, and bring things down to be in line with them (though keep in mind that some of those things may themselves be a little overtuned - it's well worth reaching out as you continue to refine the PR!).

Rin053:
I reverted the changes to the maces, but I kept reach smash distinctly for the tetsubo as it's a class-specific onimusha weapon.

I also kept the changes to the naginata since the sohei uses it.

awawaawawaawa:
If this just ported the Omnishuma & Other Kaze wretch's I'd be so down for that. I don't think Mercenaries need a buff however; they're already pretty strong here.

PR #541, opened 2026-01-15
runtimes fixes

Guidesu:
## About The Pull Request
aaaaaaaaa

## Testing Evidence
aaaaaa

## Why It's Good For The Game
aaaaaa


code/datums/weakrefs.dm
```diff
@@ -18,7 +18,8 @@
 
 /datum/weakref/Destroy(force)
 	var/datum/target = resolve()
-	qdel(target)
+	if(target)
+		qdel(target)
```
Useroth:
```suggestion
	qdel(target)
```
I don't think this is necessary, because qdel already has an istype() check in it.

code/modules/mob/living/living_movement.dm
```diff
@@ -16,6 +16,8 @@
 		unpixel_shift()
 
 /mob/living/CanPass(atom/movable/mover, turf/target)
+	if(!mover)
+		return TRUE
 	if((mover.pass_flags & PASSMOB))
```
Useroth:
```suggestion
	if((mover?.pass_flags & PASSMOB))
```

PR #548, opened 2026-01-16
Raises Keeper RCP From 3 to 5

EtheoBoxxman:
## About The Pull Request
Just bumps their RCP from 3 to 5.

## Testing Evidence
Variable change, it compiles.

## Why It's Good For The Game
Keeper's not just a combination of acolyte and apothecary, it has a higher impact on the round due to the heartbeast affecting all miracle casting, plus a few QOL boons for all doctors. Both acolyte and apothecary get +5 RCP, and have less round influence individually than Keeper does as a whole.

Just makes sense, given their general crowd gets more contribution points for less work.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/church/keeper.dm
+++ b/code/modules/jobs/job_types/roguetown/church/keeper.dm
@@ -21,7 +21,7 @@
 	give_bank_account = 1
 	min_pq = 10
 	max_pq = null
-	round_contrib_points = 3
+	round_contrib_points = 5
 
 	job_traits = list(TRAIT_MEDICINE_EXPERT, TRAIT_HOMESTEAD_EXPERT,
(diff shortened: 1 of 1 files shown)


ProbablyCarl:
It should be noted that 10RCP = 1PQ, to a max of 10PQ. Meaning, realistically, unless we make changes to the max amount? This doesn't matter. You can't play it until 10PQ as is.

It's a sound idea, for how much it can contribute, and _does_ contribute, but it has no player facing effect beyond number goes up.

EvenInDeathIStillServe:
> It should be noted that 10RCP = 1PQ, to a max of 10PQ. Meaning, realistically, unless we make changes to the max amount? This doesn't matter. You can't play it until 10PQ as is.
> 
> It's a sound idea, for how much it can contribute, and _does_ contribute, but it has no player facing effect beyond number goes up.

RCP's max reward is independent from how much PQ you have from commends. you get 10 PQ at 100 rcp no matter what

PR #556, opened 2026-01-17
Hand Selection

ProbablyCarl:
## About The Pull Request
Ports Redmoon's(I think?) version of hand selection. With my own edits.

Anyhow, I can't test this locally, given how it functions. If the Duke has a ckey in prefs, for Hand, it blocks spawning / readying unless you're that particular ckey. If he has none selected, it functions as normal.

~Saving is broken and I've not fixed it. So, DNM for now I suppose.~
I don't really care enough to have this saved. Arguably better for it to reset, even. Unless it begins to cause issues with random Hands joining mid round. It's saved temporarily to prefs regardless, and should persist until restart, barring screwing about with the button and such.

Additionally, I forgot to make mention of this, but the TMing Maint was notified. It disables Hand joining without a Duke.
## Testing Evidence
image
image

## Why It's Good For The Game
I'ma be real, this is a self port sort of deal.
Other server dies. We've an influx of a huge number of players. First round I can play Hand with a buddy on highpop, and we're set to do a gimmick? Wowzers. Great. Two other Hands competing for the slot? Even better, kinda unfortunate, but it is what it is. They'll get some great RP.

Reality of it? That Hand who sniped the slot from I and another great, genuinely fantastic RPer, AFK'd for an hour and then left the round, leaving us sitting on our hands. Pun not intended. Actual coal. Not ideal. You know the sort of things I'd wish to say, because I guarantee you'd deal with this too, and in less important roles.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/nobility/hand.dm
+++ b/code/modules/jobs/job_types/roguetown/nobility/hand.dm
@@ -256,7 +256,6 @@
 		return
 	GLOB.court_agents += recruit.real_name
 
-//Redmoon Hand selection stuff.
 /datum/job/roguetown/hand/special_check_latejoin(client/C)
 	if(SSjob.name_occupations["Grand Duke"].current_positions <= 0)
 		return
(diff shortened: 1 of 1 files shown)


kepesk1:
Hand selection was taken away with parity. I haven’t played much duke since it was taken. I don’t like having a bad hand. Yesterday, I tried to run the first post SR implosion round as duke. My hand was useless and left the round early when I really wanted someone specific. It made the round extremely tough not having a backup to help out. 

Really would like to see this speed merged. I’d also like to see duke slot prefer higher PQ players.

PR #566, opened 2026-01-18
The Big Maid PR

ophaq:
## About The Pull Request
PORTS: https://github.com/Scarlet-Reach/Scarlet-Reach/pull/2003
- Gives seneschal and servants keen ears + sleuth like SR had.
- Fixes the loadout fancy cloth sash item to use the proper one.

## Testing Evidence
meowster

## Why It's Good For The Game
It's my fetish. It's your fetish. And I am pretty sure this community wants it regardless of fetish.

### DIFF
--- a/code/modules/clothing/rogueclothes/shirts.dm
+++ b/code/modules/clothing/rogueclothes/shirts.dm
@@ -674,6 +674,7 @@
 	icon_state = "maiddress"
 	item_state = "maiddress"
 	mob_overlay_icon = 'icons/roguetown/clothing/onmob/shirts.dmi'
+	slot_flags = ITEM_SLOT_SHIRT|ITEM_SLOT_ARMOR
 	detail_tag = "_detail"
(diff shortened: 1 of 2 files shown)


code/modules/clothing/rogueclothes/storage/storage.dm
```diff
@@ -134,6 +134,15 @@
 	icon_state = "clothsash"
 	item_state = "clothsash"
 
+/obj/item/storage/belt/leather/suspenders/butler
```
GravyMaker: I didn't catch this at first, but this will need to be made a belt/rogue/leather, rather than simply belt/leather.

ophaq: OK now it works.

Githubaccount222:
Also, maids don't seem to get keen ears or sleuth. Seems the butler (not head butler, just butler) subclass doesn't either.

PR #570, opened 2026-01-18
Arcane Trickster Mage Class

EvenInDeathIStillServe:
## About The Pull Request

[ports the arcane trickster mage class from scarlet](https://github.com/Scarlet-Reach/Scarlet-Reach/pull/1143)
half mage stuff, half rogue stuff
gets 12 spell points, vicious mockery, lesser knock
T2 arcane, dodge expert and light step
image

## Testing Evidence

image
image

## Why It's Good For The Game

> I always remembered there being a spell-thief subclass but I think it got eaten in the big adventurer revamp felinid did back on AP months ago. Spellsinger is fun, but doesn't exactly 1:1 fit the sneaky trickster niche - you lack a lot of climbing and thief skills, so you have to overinvest in virtues to make up the difference. That, and you start with a best in class swift sword and martial spells, which don't really make sense for a thief. My idea with arcane trickster is to fulfill the fantasy of being a sneaky, tricky bastard but not really being much of a stand up combatant.
> 
> I think I have some concerns with a spellcaster that has dodge expert, but I think their lack of an expert weaponskill and low con/STR will make the susceptible to counterplay in the form of feints, charges, and grabs. Worst case scenario is I can make a separate 'utility/trickster' spell list to choose from if we have a suddenly influx of dodgemaxxing spitfire abusers, but that felt a little overkill to do preemptively.


ophaq:
Thank you. I was thinking of doing this, as it was one of my favorite classes on SR.

PR #572, opened 2026-01-18
Necromancer command code fixes, Necromancer balance Tweaks

RadiantFlash7:
## About The Pull Request
Necromancer NPC command code has been a touch buggy since the parity port. I've refractors the summoned NPC command segment so it should now function proper- That is: Mobs will follow a person when set to ignoring any hostile folks around, Telling them to move to a location will move them within 1 tile of it, and then loose them to normal hostile aggro.
I've also ported the option to choose your staff for hedgemages and necromancers so people can't always see a red staff and immediately decide you must be an issue.
Skeleton NPC summoned mobs had a slight loadout tweak. Worse damage dealing shield to allow for NPC AI to pick using it's weapons to attack with in all cases over bashing with it's shield. They also recieved an ancient alloy mask so they can't be one tapped as easy.
Necromancers also were given 4 more spellpoints, going from 12->16.

## Testing Evidence
image

## Why It's Good For The Game
The bugfix to NPC command AI for necromancers is somewhat key and needed. The choice to pick your staff (Rontz staff still being the best for casting out of the options) allows some mages to go about their business without being immediately pegged for a wretch.
Player controlled skeletons now get to pick their loadout on summoning.

As for balance changes: Ever since ritous was removed from necromancer, Necromancers been a bit pitiful. 12 spellpoints was not enough to justify doing much along side skeletons, especially with the skeletons face exposed. Now, necromancer NPC skeletons won't be onetapped any further with the mask in place, and an additional 4 spellpoints should help out necromancers significantly.

### DIFF
--- a/code/modules/mob/living/carbon/human/npc/skeleton/summoned.dm
+++ b/code/modules/mob/living/carbon/human/npc/skeleton/summoned.dm
@@ -76,7 +76,8 @@
 			command_target = null
 
 /mob/living/carbon/human/species/skeleton/npc/summoned/proc/receive_command_text(msg)
-	visible_message("<b>[src]</b> [msg]")
+	if(!IsDeadOrIncap())
+		visible_message("<b>[src]</b> [msg]")
 
 /mob/living/carbon/human/species/skeleton/npc/summoned/process_ai()
 	if(IsDeadOrIncap())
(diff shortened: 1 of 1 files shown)


Sirgir:
> If you're touching necromancer, is there any change you could make the summon skelly spells have a limited range? I feel like you shouldn't be able to summon them through windows and across rooftops and such

I concur, a necromancer just spamming long-range skeletons is awful to deal with-- especially since they're able to do through walls and windows. Feels like they should be summoned a couple tiles near them.

PR #577, opened 2026-01-19
Cursed Collar/Prisoner Collar Lovin'

Mryourface13:
## About The Pull Request

Yes, this is kinkslop. But also, why was this not already craft-able or otherwise able to be acquired?

- New Cursed Collar Sprite (Thanks to Yuni @yeetus5850)
- Makes cursed collar craft-able through leatherworking. 1 cured leather 1 Rontz Gem Novice skill level (same as other collars)
- Also, renames the other version of cursed collar that prisoners can get, and changes its description to make the two easier to differentiate. Now called "Prisoner's Collar". Prisoner's collar is still craft-able through smithing, as its simply a piece of decor/armor and doesn't have magic powers.
- Makes Cursed Collar unable to be self-unequipped, and leashable

## Testing Evidence

image

image

image
Disregard the gem being on the wrong side on the sprite in crafting picture. Fixed it and can't be bothered to send another picture.

## Why It's Good For The Game
The cursed collar already has full functionality for multiple pets, but right now there's no way to reliably get more than one, as you can't get more than one item from loadouts. Adding this will allow people to fulfill their dreams of being spymasters/slave masters <3


Purpleturtle3:
Cursed collars are already crafteable.
image

IronDragoon:
peak, thank you for making it unremovable this was on my to-do list

PR #580, opened 2026-01-19
Brings Arcyne Barrier AC In Line With All Summoned Armor Barriers

EtheoBoxxman:
## About The Pull Request
Title

## Testing Evidence
Variable change, it compiles

## Why It's Good For The Game
Conjure armor and dragonhide both have armor_class_none, I just missed this one


GravyMaker:
Thank you. Will TM this for a few rounds in case there are any unexpected balance implications, but it seems like a simple and fairly harmless correction.

PR #581, opened 2026-01-19
Druid Spell Glowup + Shifting QoL Ports

ophaq:
## About The Pull Request
PORTS: https://github.com/Scarlet-Reach/Scarlet-Reach/pull/1782
- Re-adds tame beast to druids.
- Adds vine wall spell.
- Changes stingers to be summoned in a wall instead of a + shape.
- Gives them wood walker trait, so druids can actually live in tree houses and not fall several z-levels trying to navigate. This is a big QoL, especially for harpy druids.

## Testing Evidence
I tested it. The spells work, the shifting works, it runs.
dendor

## Why It's Good For The Game
Much needed druid fixes, plus QoL. See the original topic for most of the specifics. The vine spell is one of my favorite spells, in combination with stingers to make druid camps made out of them as natural defenses. They're easily burnable, anyway... and mostly harmless except tripping non-dendorites.

Druid's tame beast spell was missing too, so I fixed that. I also let druids carry over their backslot items like bags while they shift. They can't access these things in their form, but now they can at least know someone isn't going to nab their stuff, and hopefully will encourage druids to shape more often, and more comfortably.

Keep in mind they will still have to dress light unless they put their clothes in their poncho or whatever before transforming! Having the neck + amulet on your ring slot staying over without having to pick it up is really nice, too.

Wood walker was added because druids easily climbing one of the thing they defend and being able to navigate them seems logical, and you don't have to waste a virtue slot for it.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/church/druid.dm
+++ b/code/modules/jobs/job_types/roguetown/church/druid.dm
@@ -23,7 +23,7 @@
 
 	//You're.. not REALLY a full-on church member, but being a druid implies you became a clergy-man of some sort; even if it's non-organized. So, still shouldn't be noble.
 	virtue_restrictions = list(/datum/virtue/utility/noble)
-	job_traits = list(TRAIT_SEEDKNOW, TRAIT_OUTDOORSMAN, TRAIT_RITUALIST, TRAIT_HOMESTEAD_EXPERT, TRAIT_WILDERNESSGUIDE)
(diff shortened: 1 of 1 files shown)


modular_hearthstone/code/game/objects/items/clothes/stockings.dm
```diff
@@ -5,6 +5,7 @@
 	icon_state = "stockings"
 	slot_flags = ITEM_SLOT_MOUTH
 	resistance_flags = FLAMMABLE
+	w_class = WEIGHT_CLASS_TINY
```
ProbablyCarl: I know what you are...

ophaq: Stockings should be small like underwear, no? Foldable and easily fits into a poncho or something. Do you want it removed?

PR #582, opened 2026-01-19
Dirt Digging QoL

ophaq:
## About The Pull Request
Adds a do_after for digging on muddy and regular tiles. You can trigger it by middle clicking a dirt or mud tile that has no hole!

## Testing Evidence
dig

## Why It's Good For The Game
I am a fisher and I am digging a hole. Diggy, diggy hole... Okay but seriously I'm tired of spam clicking holes to get bait and rocks/clay. And if you're a grave digger you can sidestep to stop the process and dig a pit as usual. Everyone wins.

### DIFF
--- a/code/game/turfs/open/floor/roguefloor.dm
+++ b/code/game/turfs/open/floor/roguefloor.dm
@@ -543,6 +543,30 @@
 			qdel(I)
 	.=..()
 
+/turf/open/floor/rogue/dirt/MiddleClick(mob/user, params)
+	. = ..()
+	if(!isliving(user))
+		return
+	
+	var/mob/living/L = user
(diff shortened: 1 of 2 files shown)


Jamdrawers:
Maybe if like... you *middle* clicked with a shovel it'd auto-scoop, instead of it happening by default

ophaq:
> Maybe if like... you _middle_ clicked with a shovel it'd auto-scoop, instead of it happening by default

That's a good idea. I'll see if I can figure that out later.

ophaq:
Added the middle click function, making default behavior with left click once again. It won't work on already dug holes, only dirt and mud tiles that have no holes. I tested it and it worked great.

PR #584, opened 2026-01-19
[UPDATED] SCOM 2 + Hermes Update (has conflicts is a liar edition)

HelennaC:
## About The Pull Request

Major changes to SCOM and updated with some minor fix on the HERMES.

- Some typo fixed on sergeant's door and
- Area fix on the south gate
- Area expansion on the inn (Innkeeper and related roles should receive their buffs on their own farm and basement)
- Fixed poor Iron duplicate locked in the keep's toilet (Often happens when you map and don't use space to move around to ended up clicking again...)
- Prelate's door fix
- Rockhill updated too (SCOM)
- Removed some private SCOMs in two houses (Not special, why they even have it?)
- Added a wider gap in the sewer's mouth (like the rest of the map)
- Inquisi. Manor area updated (Hopefully ambush doesn't spawn INSIDE anymore)

- For now, Ignore all Hermes updates. It fucked my map for some reason ;'(

## Testing Evidence
image

It just works.

## Why It's Good For The Game
Hermes updated: (Dum world)
image

SCOM updated: (Dum world)
scom4 0

SCOM before: (Rockhill) lmao it keeps going btw
image

SCOM updated: (Rockhill) more below, but too long and lazy to make edit like Dumworld *sniff
image


Jamdrawers:
I'm a little dubious about removing anonymity from SCOMs but I'm not hard-set against it. The postboxes change is nice.

HelennaC:
> I'm a little dubious about removing anonymity from SCOMs but I'm not hard-set against it. The postboxes change is nice.

I would think anonymity is nice, but wouldn't the Regent be the one behind paying for those, and technically, they would not only invest a good amount to be done, but also done right. Plus, it kinda prepares in case they become receive only, it's already named.

PR #588, opened 2026-01-19
Adds SCOMs to the Warden Fort, Refurbishes Warden Sergeant Office. 

JustMeTheIInd:
## About The Pull Request

While I understand the idea that the poor area of the Town is left to (mostly) fend for itself, and the much worse SCOM-coverage makes it hopefully easier for outlaws to hang about, I do indeed believe that there being **no** SCOMs (I checked) ANYWHERE in the lowtown to be a little silly, since it forces Wardens to run all the fucking way to the actual Town proper just to say literally anything to the rest of the Garrison. So in the spirit of the age-old saying, "Be the change you want to see in the (game)world" I went ahead and did it myself. 

The Warden Fort in lowtown now has all TWO SCOMs, yeah yeah, I know, I know, I am spoiling them. 

One of them is located in the Fort's courtyard, right below the entrance to the Barracks. The second one is located in the Barracks proper, in the sleeping area on the second story.

Furthermore, I also took the liberty to refurbish the (unused) Sargeant Office into a small chapel. Because why not?

## Testing Evidence

Zrzut ekranu 2026-01-19 234316
Zrzut ekranu 2026-01-19 232048

## Why It's Good For The Game

Hopefully, this PR strikes a balance so that Wardens won't be COMPLETELY cut-off from the rest of the Garrison, and can communicate and call for backup, while also not making SCOMs and communication too present, hopefully allowing Outlaws to still hang around the area (mostly) unimpeded.

### DIFF
--- a/_maps/map_files/rockhill/rockhill.dmm
+++ b/_maps/map_files/rockhill/rockhill.dmm
@@ -6631,11 +6631,8 @@
 /turf/open/floor/carpet/royalblack,
 /area/rogue/indoors/town/church/chapel)
 "cbA" = (
-/obj/item/bedsheet/rogue/pelt,
-/obj/machinery/light/rogue/wallfire/candle/r,
-/obj/structure/rack/rogue/shelf,
-/obj/structure/bed/rogue/inn,
-/turf/open/floor/carpet/royalblack,
+/obj/structure/fluff/psycross,
+/turf/open/floor/carpet/red,
 /area/rogue/indoors/town/warden)
 "cbJ" = (
 /obj/structure/fluff/railing/wood{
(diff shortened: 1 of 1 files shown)


Jamdrawers:
What I always did was write letters instead of sprinting to hightown to alert people to things.

Though yeah as the other guy says, I think it's nice to have an area that is more resistant to the guard-rush. Though I do appreciate that with the houndstones you would expect to use them better, pros and cons on both sides.

ProbablyCarl:
It's quite intentional that the wardens are given nothing of the sort, on Rockhill. They're not meant to have it.
Low town was intended to be cut off entirely for communication, and there's a number of ways around it as is.

Little area for prayer is neat though, admittedly.

PR #589, opened 2026-01-20
Increase spawn positions for Warden

Useroth:
From 4 to 6


ProbablyCarl:
These guys are scary at times, but I don't really see two additional slots being too big a deal, when the retinue is already against a dozen antags per round at worst. Having two over them, on a typical high pop round, is fine. Especially since they'll probably just die in the mire as is, and act as fodder on both maps. Especially the third, coming soon. I'll merge this if no one objects, and relay to a mapper of the change. Although Dun World already compensates with the tower. Rockhill's fort does not.

PR #590, opened 2026-01-20
Construct QoL Pass 1: Surgery + Misc Fixes

Spaghetti-bit:
## About The Pull Request

Adds a few misc QoL passes for constructs to make repairing them easier and make more sense overall.
Noc's TRAIT_NIGHT_OWL will no longer allow things that have insomnia to sleep. Exploit fix.

Anything with TRAIT_NOSLEEP can now learn skills / change their spells / rituos once per day. 

Repairing a construct now gives a little bit of XP, heals more and automatically stops when the construct has no more wounds. 
> Some wounds like missing eye / nose won't be repaired by such. Use surgery.
Repairing yourself as a construct adds 3 seconds. T'is a difficult task m'lord.
Lying down makes this process quicker, by 3 seconds. Minimum of 0.5 seconds per repair.

Constructs can now have surgery done on them. The engineering skill is used instead of the medicine skill. Get fucked up lil construct? Go to the artificer or ask another construct for help. Uses surgery tools.

Mostly flavor changes with how surgery works on constructs outside of a few specific ones:
- Extract lux:
> Kills the construct

- Cure Rot:
> Not possible.

- Cauterize Wounds:
> Not possible.

## Testing Evidence

Le Open Hatch:
image

Loaded in, did a few surgeries on a construct. It used engineering skill, gave me engineering experience.
Also did surgeries on non construct, it used the proper variant, used medicine skill and gave experience for such.
- amputation
- limb attachment
- plastic surgery
- cure rot (can't with constructs :) )
- extract lux (killed the construct )
- cauterize wounds (not possible w/ constructs)
- cure rot (Not possible w/ constructs)
- relocate bones
- sever external organs
- set bones
- manipulate organs
- incision
- infuse lux

Did the same with needles. Construct? engineering experience, didn't use a thread. (Only possible with an open hatch) Non construct? Medicine experience. Used a thread.

## Why It's Good For The Game

Currently; when a golem gets their eye, tongue, or nose taken / broken they cannot get it back without admin intervention. Fixes that.

Acts as a placeholder until / if we want to get specific tools to work on constructs.

My only concern is being able to grind engineering skill fast but... golems need someone else to give them the cog anyways. Dolls are weak as is, don't think them having a high engi skill will matter. That's also already capped with needing a trait to gain skills past apprentice.

### DIFF
--- a/code/modules/roguetown/roguejobs/blacksmith/tools.dm
+++ b/code/modules/roguetown/roguejobs/blacksmith/tools.dm
@@ -160,6 +160,10 @@
 		return
 	if(!ishuman(M))
 		return
+	if(!isliving(user))
+		return
+	if(user.doing)
+		return
 	if(M.construct)
 		var/mob/living/carbon/human/H = M
 		var/obj/item/bodypart/affecting = H.get_bodypart(check_zone(user.zone_selected))
@@ -179,7 +183,8 @@
 		
 		used_time = max(0.5 SECONDS, used_time)
 
-		while(do_mob(artificer, M, used_time) && (affecting.get_damage() != 0 || length(affecting.wounds)))
+		playsound(loc, 'sound/items/bsmith1.ogg', 100, FALSE)
+		while(do_after(artificer, used_time, target = M, same_direction = TRUE, no_interrupt = FALSE) && (affecting.get_damage() != 0 || length(affecting.wounds)))
 			if(get_dist(artificer, M) >= 2)
 				to_chat(artificer, span_warning("I need to be closer to them to repair them!"))
 				return


GravyMaker:
I'm intrigued by the possibility of creatures with TRAIT_NOSLEEP being able to refresh things which otherwise refresh when sleeping while tired - I'll have to consult with the others about this, both as applied to constructs and in terms of whether there are any particular creatures we'd need to disable this for prior to a TM. That said, these changes mostly sound quite valuable, thank you.

Spaghetti-bit:
> This here: `Anything with TRAIT_NOSLEEP can now learn skills / change their spells / rituos once per day.`
> 
> That concerns me. You're forgoing those by ignoring sleep altogether. What's the point to the vice? Sleep healing? I can just ignore sleep healing altogether, now, even as a side-effect to other functions. The only point to sleeping (for most) was to set skills and continue your advancement through stuff like rituos / resetting spells.
> 
> The rest of this is gem, though. Even if we don't chance the above. It's soulful.

You won't advance skills, the TRAIT_NOSLEEP only removes the TRAIT_STUDENT perk allowing them to read another book, which does nothing beyond apprentice level since you either need A) an archivist or B) to sleep to use said sleep experience.

PR #591, opened 2026-01-20
Rockhill Tavern QOL

EtheoBoxxman:
## About The Pull Request
Makes the basement slightly larger and actually gives a dorm to Tapsters, who had to share as single bed with the cook originally. 
As fun as cuddling can be, this was awkward in rounds full of tavern workers.
The basement has a link into the moat, with a climbable wall and solid ground so people can break into (or out of) the tavern discretely, and this also gives the tavern access to fresh water for filling their barrels without having to travel out of their way to get water for cooking or serving people.

The moat dropdown into the basement area:
image

The new basement area:
image

## Testing Evidence
image
image
image
image
image

## Why It's Good For The Game
Gives the tapsters a bedroom, and the basement allows for tavern workers to not have to leave the building for re-stocking water (which Ive had to do four times in an hour, with the new pop). Extra space in the basement is also good for tapsters who wish to go especially hard on brewing alcohols, what with the added space for farming if they invest in the activity.

### DIFF
--- a/_maps/map_files/rockhill/rockhill.dmm
+++ b/_maps/map_files/rockhill/rockhill.dmm
@@ -8824,14 +8824,6 @@
 	},
 /turf/open/floor/rogue/woodturned,
 /area/rogue/indoors/town/tavern)
-"cJh" = (
-/obj/structure/mineral_door/wood/red{
-	locked = 1;
-	lockid = "tavern";
-	name = "Tapsers' Quarters"
-	},
-/turf/open/floor/rogue/woodturned,
-/area/rogue/under/town/basement/tavern)
 "cJs" = (
 /obj/structure/fluff/railing/border{
 	dir = 1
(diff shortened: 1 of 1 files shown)


EtheoBoxxman:
> > Oh, is it live right now?
> 
> I added the tapster attic and it was merged like a few weeks ago

Gotcha! I went and did a pass over the tavern and I just missed it when checking the doors xD
Ill remove the spare bedroom, since the main thing I wanted to add with this was the water source

Jamdrawers:
Might just be me but I always kinda liked the short walk from the bar to the river to fill buckets with.

EtheoBoxxman:
> Might just be me but I always kinda liked the short walk from the bar to the river to fill buckets with.

Its cozy, yeah, but when you've got a constant flow of patrons for an hour straight, having to refill in the middle of the 'lunch rush' annoys everyone involved because of the walk from the kitchen to the river or fountain xD

PR #592, opened 2026-01-20
Short Race Cloaks and Tabards

yupup123:
## About The Pull Request

Adds Short Race versions of:
- Ladylike Cloak
- Lordly Cloak
- Direbear Cloak
- Wicker Cloak
- Stalker Cloak
- Stole
- Religious Tabards

Adds a white version of the `Tabard, Black` and `Tabard Shroud, Back`. They are both craftable and in the `loadout.dm`

Also fixes the `Nurse's Veil` raised sprite from being bugged.

## Testing Evidence
Screenshot 2025-12-25 214428

image

image

image

## Why It's Good For The Game

Short races are missing some pretty important clothes. This fixes that. Some of the sprites might not be the best, as I'm not an experienced spriter, but they are passable and better than nothing.

### DIFF
--- a/code/modules/clothing/rogueclothes/robes.dm
+++ b/code/modules/clothing/rogueclothes/robes.dm
@@ -321,6 +321,7 @@
 	color = null
 	icon_state = "feldrobe"
 	item_state = "feldrobe"
+	slot_flags = ITEM_SLOT_ARMOR|ITEM_SLOT_CLOAK
 
 /obj/item/clothing/suit/roguetown/shirt/robe/tabardblack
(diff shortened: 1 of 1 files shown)


ProbablyCarl:
I think we should gib all rat folk.

ProbablyCarl:
Slated for TM. Just a matter of it conflicting with another TM, so it'll remain un-TM'd for the moment.

yupup123:
Not sure how to fix these conflicts honestly... might have to just re-do the PR.

PR #593, opened 2026-01-20
AP Port: Sweetspots! Kind of!

ProbablyCarl:
## About The Pull Request
This includes my own edits, of course, and updates made to otherwise nonfunctional stuff. That said, lets get into what's different than the [original](https://github.com/Azure-Peak/Azure-Peak/pull/4980) PR. Credits to Free!

And double credits to @Spaghetti-bit, for actually fixing my mistake here and making it work. Whoops!

 - Scythe cutting given a sweetspot, of two or more paces.
 - The lance is given an effective range, of two or more paces.
 - Grand Mace / Godens given sweetspot mechanics. (Ranged thrust they use, really.)
 - Militia Goden given a one-handed replacement.
 - Spear one-handed thrusting speed increased dramatically, with no massive damage increase, as compensation.
 - Spear one-handed cutting speed increased, dramatically.
 - Sword lunging now has an effective range, of two paces. Given estoc and such.
 - Rending all has a double sharpness penalty.
 - Reaching rends have an effective range.
 - Eagle's Beak given a proper one-handed thrust.
 - Nomad's spear kept its original mechanics of being a one-hand exclusive spear, given that's the entire point of it. We can nuke it if it turns into a meta option, but I think I and like one other player are the only people who touch this as is. So, self buff, I guess. But it's unobtanium outside of a class that has it, on adventurer.

Sweet spot ranged mechanics and pacing was actually nonfunctional. It looked for ranged exactly above or below, for above/below checking, rather than just above or below. Just... trust that this works, please? It's tested. I don't know why it was the way it was before. But it works now.

## Testing Evidence
image
image

## Why It's Good For The Game
We're in an excellent spot currently. Rather than needing to remove peel like most communities, it just doesn't really get used. By comparison to straight up using armour penetrating weaponry, that is. It just isn't. I've watched, taken accounts and played alongside folks who previously swore by it.
Even something as catastrophically overpowered at times as the partizan is just a shell now, compared to what it was, because the better peel (when it had it) just doesn't matter. Given you can just stab anyways with it.

That's great, of course. But that also means those weapons that can do anti-armour gameplay take priority, and that means spears, halberds, etc, just outclass everything else. Hopefully with a ranged requirement, or punishment for using them in place of a more logical option? Some headway can be made in diversifying our roster of weapons in active use.


GravyMaker:
Let's give this a try. We'll need to be open to balance feedback but I think the concept here is a sound one.

Pidgey-chan:
This is a pretty big spear/polearm thrust nerf, despite the additional pen. They're already limited by the slow attack speed, this turns their reach into a downside!

Pidgey-chan:
> > This is a pretty big spear/polearm thrust nerf, despite the additional pen. They're already limited by the slow attack speed, this turns their reach into a downside!
> 
> That's the point, I fear. Kind of. It's to prevent you from rocking the spear and just facehugging someone with free AP. The way it functions is that if you're not in an appropriate range to use it, it'll remove the pen (change it to blunt pen, so nothing), and halves the damage. I'll be drafting this for now, in any case, since it's on the backburner while performance changes prevent its testing.

The thing that prevents you from facehugging with spears is that spear stabs are slowwww. If you're at that point, picking or smashing peeling or even just cutting at a piece of armor will probably break it faster than you can stab.

I also personally think that greatweapons (like all polearms) can and should be just better than weapons that can be worn as sidearms to compensate for being obvious and harder to carry.

Pidgey-chan:
Polearms are CUCKED. Greatweapons are now WORSE than shit you can store on the belt... or even a bag! See knucks. Now that range is meaningless, what fool would touch a spear over a warhammer they can hold in one hand for 20 more damage through armor? And the halberd (giant axe) STILL only cuts for .9 damage!

PR #594, opened 2026-01-20
Adventure + Wretch max

Rayforx:
## About The Pull Request
+ Increase the pop for adventures to 40
+ Makes the max to 15 for wretches, while they use the same 10 players to one wretch, scaling

## Testing Evidence
N/A

## Why It's Good For The Game
The adventures are pretty much the default "class" that people can play, joining and seeing every role being taken, with max adventures usually leads with the player saying "fuck it" and doing something else.

I think the scaling is fine with the wretches, so we do not have to worry about more wretch slots being added.


Hmeister-real:
Hell no we do not want massive wretch stacks like SR. If someone can't play either and is forced to play an impactful role or leave and help with the TiDi then that's great.

PR #603, opened 2026-01-21
20 max slots for towner roles

RadiantFlash7:
## About The Pull Request
Creates 20 max slots for differant towner types

## Testing Evidence
image

## Why It's Good For The Game
This isn't actually to limit the amount of a single type of towner, more so to let players know just how many there are of a type of towner. Theres nothing worse then joining as a character with an intent to play a type of towner, only to see 5-6 others of the same type, when you could've played a seperate character as a differant kind of towner.


ProbablyCarl:
This is such a hacky stopgap measure and I adore it. Actual gem.

Spaghetti-bit:
I am not the devil he speaks of upon his shoulder… slander.

PR #614, opened 2026-01-22
Natural armor virtue, loadout natural armor gone

Purpleturtle3:
## About The Pull Request

Removes the natural armor from the loadout options, adds a new virtue to replace it.

## Testing Evidence

image
Screenshot 2026-01-22 005122
Screenshot 2026-01-22 005321
image
image

## Why It's Good For The Game

The natural armor loadout item was making a pile of ahelps every other round, while people also complained of this being too weak. The armor is better while also taking a virtue slot to justify its existance.


ProbablyCarl:
Welcome back, Hestia subdermal armour...

Purpleturtle3:
> Weak is only 'weak' by virtue of changing the description, as it was meant for loadout use. If you want it to be similar, you'd have to remove the crit resists from it. Otherwise this is identical to disciple regenerating skin. Might be intended here, but just in case it's not.
> 
> The loadout item had no crit protection, with the armour values significantly tanked, just by how that system worked. It's part of why I called it a nothingburger. You gimped yourself by taking it, for anything but RP purposes. Even a gambeson was just objectively superior in all aspects.

No longer a subtype of disciple skin, removed the crit prot flags.

Pidgey-chan:
I don't think invisible inherent unremovable armor should be available to everyone regardless of role. It's not busted for all out combat or anything, a spear will stab right through this, but I think it's incredibly lame to get into a bar fight and a random lupian is immune to 300 damage of your punches

Genevesh:
> I don't think invisible inherent unremovable armor should be available to everyone regardless of role. It's not busted for all out combat or anything, a spear will stab right through this, but I think it's incredibly lame to get into a bar fight and a random lupian is immune to 300 damage of your punches

You can see tough skin on examine though, unless you have a very bad perception, so... don't get into bar fights with tough-looking lupians (literally), or grab something that isn't lethal but isn't your literal fists. A chair, maybe (though I don't actually know how powerful a chair is)

JGlitch:
I can't believe people thought tough skin was OP. It broke like nothing... And now people are gonna complain more cause its a bit better, even though it costs a virtue. I wanna remind yall about the acrobatic trait.

PR #619, opened 2026-01-22
The Stairs Ascend: Stair QoL + Stamina Cost

Spaghetti-bit:
## About The Pull Request

Stairs now cost a little bit of stamina when going UP, not down. Costs more when lying down.
Ladders are also affected.

No, you can't drain stamina from someone by dragging them up stairs / ladders.

Chain grabbed mobs now retain their grabbed mob / object as well.

## Testing Evidence

Went to a staircase by the rockhill bridge (Low-town to regular town) and observed stamina cost.

## Why It's Good For The Game

No more stair cheese / Z-level change spam.

kung-fu-panda

### DIFF
--- a/code/game/objects/structures/stairs.dm
+++ b/code/game/objects/structures/stairs.dm
@@ -152,11 +152,11 @@
 			while(P.pulling && isliving(P.pulling))
 				was_pulled_buckled = FALSE
 				pulling = P.pulling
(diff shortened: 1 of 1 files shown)


Jamdrawers:
It should also cost a chunk of stamina to climb up a wall. Probably a good chunk of stamina. Means you can have more of an advantage holding the high ground

PR #656, opened 2026-01-26
Lowers spawnrate of trees in bog

Heheniy:
## About The Pull Request
Combines with manually placed trees, current spawnrate makes small islands of land  in bog almost impassable, with most of their tiles covered in trees and rest by bushes and other things

## Testing Evidence
varedit

## Why It's Good For The Game
makes bog at least somehow passable


Jamdrawers:
I've got an alternate and moderately cleaner way of doing something like this to rockhill (which reduces spawnrates overall rather than specifically that particular type of tree) - I'd request we un-TM this, or at least un-TM the part that effects bograt.dm. I'm unopinionated on the dunworld bog.

PR #669, opened 2026-01-27
Increases WW infection chance + lick wound infection

Heheniy:
## About The Pull Request
Increases WW infection chance on wound 15% -> 25%
Adds back lick wound infection chance from old RW (10% to attempt to infect (you will need then to pass 25% check too)) 

## Testing Evidence
image

## Why It's Good For The Game
Playing WW rn is really a CURSE
You can spend whole night trying to infect one person for them just to die in the end

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/wildshape/cat.dm
+++ b/code/modules/mob/living/carbon/human/species_types/wildshape/cat.dm
@@ -218,8 +218,8 @@
 			rid = /datum/reagent/water
 			target.reagents.add_reagent(rid, ramount)
 			if(prob(10))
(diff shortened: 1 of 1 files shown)


DelineFortune:
> Do you remember when this used to give sewage, and despite my protests, we added that anyways and it became a new fragstun tool? Wild how that worked. Anyhow, good stuff.

Been made by me because wws were broken as fuck and we had crit fishing (a random stab could kill u)

So I gave them a solution to handle vs solo armored targets as they had 0 solutions to do anything 

Also I didn't like combat medics friendly wws

PR #677, opened 2026-01-27
Lights optimization

Rayforx:
## Light optimization - TM needed
**Lightning_source - Qdeleted stuff in sslightning**
Added early Qdelete check in SSlightning.fire() before calling update_corner
Skips deleted light sources immediately, avoiding proc call overhead for invalid stuff
Follows the same for lightning Object queue
**Lightning sources - Turf list optimization**
Replaced affecting_turf approaches with direct removal
Reduced memory allocation
**Lightning_source - corner list optimization**
Replaced effect_str with direct iteration to build corners_to_remove
Repeat effect_str and check !things in corners to identify that corners are no longer in range
Remove redundant corners to remove operation inside the main loop, since the removal list is already pre-built
Single pass processing all corners, instead of it being called multiple times

Probably some more, that I forgot.

## Testing Evidence
image

## Why It's Good For The Game
Less lag, experimental


Spaghetti-bit:
TM'd -- Lowpop

PR #683, opened 2026-01-28
Adjusts runaway prisoner migrant wave downgrades

HeroWithYay:
## About The Pull Request

Changes the slots for the Runaway Prisoner migrant wave progression to go 4, 3, 2, 1 instead of 4, 3, 3, 2
This allows for a singular player to spawn as a runaway prisoner rather than requiring at least two.

## Testing Evidence

Tis a single number change, sire

## Why It's Good For The Game

Very un-used migrant wave that I think wouldn't diminish the server for it to be playable solo in the case no one else readies. Might even be a fix? Depending on whether or not it was intended that you needed at least two people in the first place.


ProbablyCarl:
Actual oversight fix, I'm 99% sure.

PR #700, opened 2026-01-29
Dun World - Redux

ProbablyCarl:
## About The Pull Request
Oh, boy. This is a huge one.

The culmination of days worth of effort, by Alahan/Kepesk. This is a redux of Dun World's town and adjacent locations, featuring an obscene number of edits and entirely new locations built around said edits.

I won't provide too much in terms of images, given it includes plenty of secrets and special locations(you can just open the map editor if you care that much), but this has been built with player feedback in mind and attention to detail. If you've followed the map channel, you already know what this PR includes, anyhow.

I've made a few small edits here and there as corrections myself, or added something in place of another concept that wouldn't work. But overall this is entirely Alahan's. I had to go through and name every single scom and nervelock. I hate them so much.

Anyhow, given the nature of some of these changes, we've elected **NOT** to make it a new map, as originally intended, as this is effectively just a huge edit for the moment. Significant edits to further push it away further are planned, including an eventual rename.

Edit: Includes keys for the servants to lock up the kitchen, now, on both maps. Why? Requested, and it's such a small change I don't care enough to atomise it. John MAA can no longer walk in and Gronnish-hoover up the pantry. It also requires a map edit, which I'd done in this PR, to save the trouble on both fronts.

## Testing Evidence
image
image
image
image
image

## Why It's Good For The Game
Why is this good for the game? Do I have to describe why Dun World as is, just, overall, happens to be awful? Rockhill is an example of a good map. Excellent, even. The mire is hostile. Everything is strewn about where you'd expect it.

By comparison, Dun World feels like twenty different folks went through it and just smeared paint where it looked appropriate. There's no rhyme or reason for half of it. It doesn't have the theme or care that we had for something like Heartfelt, or OldRW's Rockhill. It's just bland and boring. Is AP's newest rendition of it better? Leagues, I've heard. But as it stands, we're trapped with a version that few folks like, and even fewer are willing to map for. So hopefully this gets us in that direction of folks wanting to participate again, or otherwise making the playing space a bit more appropriate for roleplay. Which is the biggest point of this all. Encouraging roleplay further, by providing the map in which to do it.

### DIFF
--- a/_maps/map_files/dun_world/dun_world.dmm
+++ b/_maps/map_files/dun_world/dun_world.dmm
@@ -4279,6 +4279,7 @@
 	lockid = "farm";
 	pixel_y = 14
 	},
+/obj/item/storage/roguebag,
 /turf/open/floor/rogue/cobble/mossy,
 /area/rogue/outdoors/town)
 "aUO" = (
@@ -21285,7 +21286,6 @@
 "eKL" = (
 /obj/structure/rack/rogue,
 /obj/item/clothing/neck/roguetown/psicross/inhumen/aalloy,
-/obj/item/reagent_containers/food/snacks/grown/rogue/fyritius/bloodied,
 /turf/open/floor/rogue/concrete,
 /area/rogue/indoors/inq/basement)
 "eKQ" = (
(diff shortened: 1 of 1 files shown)


anconfuzedrock:
why is the merc guild next to the keep... it's even on the same side of the moat. I don't think the untrustworthy gaggle of foreign mercenaries would get real estate right next to the tower.

PR #701, opened 2026-01-29
Adjust Ruma Clan Gun-in stats for the mercenary subclass

polvohime:
## About The Pull Request

Given the loadout of the gun-in, I believe it would better to make an adjustment in SPEED and STRENGTH.

Basically, we will be changing the -1 SPD into -1 STR

And the buff of +2 STR to +2 SPD

This change is made given the gimmick of the subclass. They are given a SWIFT balance weapon on spawn, which basically uses your SPEED compared to weapons that often use STR.

I don't think this will break or make this class too overpowered, and this may have been but an oversight from the past; and hoping this would at least give this class some love.

## Testing Evidence

image

## Why It's Good For The Game
Fixes an oversight(?)

### DIFF
--- a/code/modules/jobs/job_types/roguetown/mercenaries/classes/rumaclan.dm
+++ b/code/modules/jobs/job_types/roguetown/mercenaries/classes/rumaclan.dm
@@ -12,9 +12,9 @@
 	subclass_stats = list(
 		STATKEY_CON = 3,
 		STATKEY_WIL = 3,
-		STATKEY_STR = 2,
+		STATKEY_STR = -1,
 		STATKEY_PER = 1,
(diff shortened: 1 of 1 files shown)


polvohime:
> not every class is balanced to use all their weapon's traits. the knight captain doesn't have a speed bonus for their saber. (...)

the +4 SPD and +2 PER for the archer class is very good and it fits the loadout perfectly, even removing the STR and CON for them; compared to its melee counterpart which gets a swift weapon which scales off the SPD stat and a +2 to STR

image


while i understand the captain does not get a speed bonus, he does get a bonus in everything else without penalties-- as well as the choice of going glaive if you are not leaning into the sabre/shield combo

PR #705, opened 2026-01-30
rebalances most mercs to 7 stats and upgrades condioterro+vaquero 

CVSthePharmacy:
## About The Pull Request

rebalances mercs to be at a minimum of a +7 statblock to  be in line with our current statblocks

 ALSO buffs condioterro slightly in terms of gear, aka a messer and a pair of splint leggings, think a full coif and better gambeson too, since atm it was universally agreed that condioterro was the WORST of the mercs. Also gave them the proper sword skills- because for some reason they didn't come with expert? Huh? 

ALSO gives the friefetcher their punchdagger on their kriegmesser kit to add some avar in there instead of being all german. Their lancer kits all get the same punchdagger now, which is literally the same thing but reskinned

ALSO gives vaquero the equestrian trait and +1 riding to masters, why? Because the whole thing is centered around riding... unlike avar which is kinda just.... idk what they are but they're something. Vaqueros needed some love, I gave them it!

ALSO renamed the kits for samurais, aka hangyaku to be accurate to what the weapon gives. 

AS FOR STAT CHANGES!!!!

Desert riders
Janissary given +1 wil to a total of 3
Zeybek given +1 per to a total of 1
Almah lost give +1 per to a total of 0 Per

Friefetchers
Lancer lost 1 STR in exchange for 1 INT, class is based around feinting not strength rushing

Hangyaku
Samurai given +1 con to a total of 2 con

New moon
Given +2 per to a total of 2

Steppesan
Saber veteran given +1 spd to a total of 2 spd
Palya sapper (ordinance carrier) given +4 con, yes it's very high but it had a statblock of 3 prior with a -4 bonus to speed. They're meant to carry shit, this will make it shown. Also they have jman axes and no armor skills so LOL they'll take a lot of hits
Archers given +1 spd to a total of 3 spd
Light infantry given +1 wil to a total of 2 wil

Underdwellers
Removed spd bonus in exchange for +1 to str, for a total of 2 str

Vaqueros
Given +1 int for a total of 3 int

## Testing Evidence

number change, and GEAR. I'll check what github tells me is wrong

## Why It's Good For The Game

MERCS SHOULD BE THE SAME STATBLOCK RELATIVELY, IF THEY'RE LESSER THEN NOBODY CHOOSES THEM!!! Also gives condioterro medium armor cause it's already the worst class ever. ALSO masters riding apparently does nothing cause horses don't go over the max speed after getting to expert riding, I dunno why, might be a bug, but it's flavorful for now. The only big change is that vaqeuros get equestrian to not knock themselves out on doors. 

ALSO THE HANGYAKU KITS DIDN'T SAY WHAT WEAPONS THEY GOT SO I FIXED THAT


CVSthePharmacy:
> This is not number changes -- this has a bunch of item changes, some of them providing better armor prots and others exchanging bad weapons for better weapons. Another adds equestrian. Not sure what's going on with those but also they don't need rebalancing, these are already balanced roles

also the only item changes were condioterro's leather trousers for brigandine chausses + a better gambeson and changing the iron shortsword for a steel messer, not sure what else you mean that got changed. Also7hat they're fine but like, they're under the actual 7 stat requirement. Mercs are s'posed to be like that man

CVSthePharmacy:
> Mercenaries are not balanced around having +7 for their statblock, they are balanced around being +9, with STR and SPD weighted at 2 for each point.

I'll ask carl if this is true

CVSthePharmacy:
it is true

PR #711, opened 2026-01-30
Ports vanderlin reflections

Fatherless1Xx:
## About The Pull Request

ports reflections using a slightly modified variant of vanderlin's code 

## Testing Evidence

Facing mirror:
image
Facing right:
image
Facing away:
image
Facing left:
image

On non-reflective tile:
image

On reflective tile:
image

In water:
image

## Why It's Good For The Game
i use reflections to snipe people with arrows

### DIFF
--- a/code/game/objects/structures/mirror.dm
+++ b/code/game/objects/structures/mirror.dm
@@ -127,6 +127,9 @@
 	if(!L || obj_broken)
 		clear_mirror_reflection()
 		return
+	if(get_mirror_target() != L)
+		clear_mirror_reflection()
+		return
 	if(mirror_reflection)
 		cut_overlay(mirror_reflection)
 


john-pawson:
Is there a chance this'll affect performance?

GravyMaker:
I'm going to make a quick note here that this PR was only un-TM'd due to hard-to-diagnose technical issues which _probably_ weren't the fault of this TM itself - but which we needed to be aggressive about given their severity, hence the recent un-TMing of almost every active PR.

We're going to be re-TMing PRs slowly, a couple at a time, over the coming days, and this one should be among them. I'll re-apply the TM label to ensure we don't forget about it, but you shouldn't expect to see it back in the game for probably another couple of days.

Thanks for your contribution, and your patience!

PR #714, opened 2026-01-30
Hemophage Vice

ProbablyCarl:
## About The Pull Request
A new vice in place of vampire stuff. Provides +1 triumphs. Adjusts the triumph note on other virtues so they say 'TRI' instead of 'TRIUMPHS', because it was weird.

You can drink blood to regain nutrition and hydration, atop refilling your blood. Albeit at a slower rate than water, by a bit. Poor blood sources, such as the standing blood pools, DO WORK, but are poor by comparison to a proper source. It will convey this if used.

This was hardly tested.
## Testing Evidence
image
image
image
image

## Why It's Good For The Game
It's soulful. Or something. I 'unno. Content.

### DIFF
--- a/code/datums/character_flaw/_character_flaw.dm
+++ b/code/datums/character_flaw/_character_flaw.dm
@@ -751,4 +751,5 @@ GLOBAL_LIST_INIT(character_flaws, list(
 
 /datum/charflaw/hemophage/on_mob_creation(mob/living/carbon/human/vamp_wannabe)
 	ADD_TRAIT(vamp_wannabe, TRAIT_HEMOPHAGE, TRAIT_GENERIC)
+	ADD_TRAIT(vamp_wannabe, TRAIT_VAMPBITE, TRAIT_GENERIC)
 	vamp_wannabe.adjust_triumphs(1)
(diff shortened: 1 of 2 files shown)


Purpleturtle3:
A vampire reported that with this vice, drinking blood caused them to cough blood.

BriareosBlue:
I got a chance to test out with the adjustments and it seems that drinking from simplemobs still isn't possible, the bites don't do anything but play the bite sound. I've also tried biting a dead Deep One, and I get the message "My power is weakened, I cannot heal!" from trying to do so. Doesn't seem to satiate either.

Guidesu:
> I got a chance to test out with the adjustments and it seems that drinking from simplemobs still isn't possible, the bites don't do anything but play the bite sound. I've also tried biting a dead Deep One, and I get the message "My power is weakened, I cannot heal!" from trying to do so. Doesn't seem to satiate either.
> 
> Edit: Actually, it seems at some point in the round, I started getting that message after every simplemob I try biting.

Wdym you expect from drink from simple mobs? It was never possible.

BriareosBlue:
> > I got a chance to test out with the adjustments and it seems that drinking from simplemobs still isn't possible, the bites don't do anything but play the bite sound. I've also tried biting a dead Deep One, and I get the message "My power is weakened, I cannot heal!" from trying to do so. Doesn't seem to satiate either.
> > Edit: Actually, it seems at some point in the round, I started getting that message after every simplemob I try biting.
> 
> Wdym you expect from drink from simple mobs? It was never possible.

I remember that being a thing back when the vampire statpack was being TM'd. Unless somehow I'm misremembering.

PR #722, opened 2026-01-31
Directional Curtains

ProbablyCarl:
## About The Pull Request
Simple as the title implies. You can only open them from one direction. This doesn't change anything existing. Just something for a speed merge, regarding ongoing mapping efforts.
image

If these are made craft capable at some point, we'll uncomment the bits of it applicable. But this is JUST for mapping, as of the moment.
## Testing Evidence
image

## Why It's Good For The Game
Maint veto inclusion or some nonsense.
Add this. Now. Thanks.


kepesk1:
my hero <3

PR #734, opened 2026-02-03
Make rockhill great again

Murderfoxen:
We have to make rockhill great again.

To fight these scarlet tides, we must bring back the favorite past-time of bored wardens.

### The bog-gate. 

image

image

image

This mysterious structure was -always here- and manages to somehow have some of it's dishelved walls intact.

### Some very unfortunate news.

There was a collapse at the long-dilapidated mines to the east. 

image

This, and the recent dendorite growth nearby; have made the eastern passage very dangerous.

Only three dwarves were hurt in the accident. And a molebear. It was very tragic.

### The Malumites have gained class conciousness;

image

The smithy has moved to be near the bog-gate; as it will make a very popular spot for folk to give their mammons away, or to get gear before going out. -Almost- an official bog-smith!

### The Mercenary Guild hath moved;
image

The Mercenaries guild has moved to be in the pass between Hightown and Lowtown. This spot should give mercenaries much more of a centralized location between the flow of RP/combat

### The farmlands, they have shifted; and so too hath the abbey!

image

The soilsons have moved ever-slightly; and the church quite a lot. This should be a pretty quiet zone, but also a good spot for antags to still steal away folk; if they can figure out some genius way to get to this side of the river!

### DIFF
--- a/_maps/map_files/rockhill/rockhill.dmm
+++ b/_maps/map_files/rockhill/rockhill.dmm
@@ -357,9 +357,13 @@
 /turf/open/floor/rogue/cobble,
 /area/rogue/indoors/town)
 "afG" = (
-/obj/structure/fluff/walldeco/vinez,
-/turf/closed/wall/mineral/rogue/craftstone,
-/area/rogue/indoors)
+/obj/structure/table/wood/poor,
+/obj/structure/bars{
+	icon_state = "barsbent";
+	layer = 2.81
+	},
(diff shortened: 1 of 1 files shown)


ProbablyCarl:
There are parsing issues.
Please compile and run the map locally as well, to test.

Pidgey-chan:
SOVL. SOVL. BRING BACK BOGGAS

PR #739, opened 2026-02-03
Resonance Caster Return | Venomous Return | Antipope Adjustment

ProbablyCarl:
## About The Pull Request
Parity broke both resonance casting and venomous. This simply returns them. Which acts as a buff to Priest and Antipope for resonance casting, and a general content return to lamia.

If you're unfamiliar with venomous, it just enables you to inject a special poison when chewing someone, assuming it can go through their armour.

For resonance, it just fortifies people in an AoE of 2 tiles in view of the caster. Making it easier to deal with mass casualty events, yet not discriminating and able to aid the enemy. I really should finish this trait at some point.

This also nerfs Doomsayer. Heavily. On the opposite end. While also adjusting their notes somewhat. They've lost magic, alongside their arcane and alchemy skills. Polearm skills were adjusted to be stave instead, with them getting a stave in place of the toper staff, and unarmed was dropped to journeyman. Keep in mind that their STR/CON setup still allows them to do some funny wrestle gimmicks, but this wasn't meant to be a mainline class, as it has turned into. Meant to be support, instead. And that's not how it was played, soloing folks and the like, in an absurd, similar fashion to Iconoclast. So hopefully this pushes them towards that, as something more RP focused, with the potential to wrastle when needed.

Doomsayer, aside, has been given message and non-detection as spells. Given it's harder to communicate and lead otherwise.

## Testing Evidence
image

## Why It's Good For The Game
Folks will complain about doomsayer, but it truly was absurd at times. It was added well before any of the more frag happy folks returned, and so the concern about it being used to do some stupid stuff wasn't present. Now, it is, and I've had staff complaints relating to this. So, an adjustment is being made.

The other two are just yet MORE fixes to return stuff parity removed. I hate parity.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/wretch/antipope.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/wretch/antipope.dm
@@ -59,6 +59,8 @@
 	H.mind?.AddSpell(new /obj/effect/proc_holder/spell/invoked/convert_heretic)
 	H.mind?.AddSpell(new /obj/effect/proc_holder/spell/invoked/wound_heal)
 	H.mind?.AddSpell(new /obj/effect/proc_holder/spell/invoked/silence)//Shut that guy up!
+	H.mind?.AddSpell(new /obj/effect/proc_holder/spell/targeted/touch/nondetection)//For the purposes of meeting folks.
+	H.mind?.AddSpell(new /obj/effect/proc_holder/spell/self/message)//See above.
 	H.verbs |= /mob/living/carbon/human/proc/completesermon_evil
 	H.verbs |= /mob/living/carbon/human/proc/revelations
 


Guidesu:
Good to go after the freeze. Nothing particularly odd in the code. You should probably give the venemous trait to races like lamia though.

Guidesu:
> > Good to go after the freeze. Nothing particularly odd in the code. You should probably give the venemous trait to races like lamia though.
> 
> It already has the trait, brother. Parity broke it. Like resonance casting. This is a 1:1 return!!! WHAT DO YOU MEAN!!!!

i swear to god last i checked lamias did not had them!!

maker-of-nightmares:
Why the doomsayer nerf? Isn't it just a trader subclass?

Tekkomeka:
Wah, no 20 str 20 con doomsayer expert wrestling expert unarmed anymore!

PR #747, opened 2026-02-06
Blunt Integrity Damage Nerf

ProbablyCarl:
## About The Pull Request
Wow. Wild!
This drops integrity damage increase for weapons that have it from 40% to 20%. What this does is halving the increased damage blunt does to armour, basically. If they used the standard value, that is.

Why? Because we've chipping now, and you don't need the crutch of destroying plate as quickly. Like was needed _elsewhere_.

Additionally, technically, the executioner's sword sees a buff in that the -10% is actually just in line with standard blunt chipping. It has also been given miniscule level chipping, which is 10%, since it's, as per old fluff, basically being hit with a mace. If this is dumb we can drop it. Simple edit, anyhow.

Mauls also retain old armour integrity damage values, with updated comments.

Did I test this? No. Simple number corrections. It also removes my code comments because the joke is dead.

## Testing Evidence
Slop PR. No testing to be done. Sorry. Carl classic.
## Why It's Good For The Game
Destroying full plate for every mace in a few strikes was funny. Still is. This doesn't really entirely fix that. Just makes it less obscene by a tiny bit. If it's still too strong, we can drop the mechanic of increased damage from them altogether. But we shouldn't need to, I'd think.

### DIFF
--- a/code/game/objects/items/rogueweapons/melee/blunt.dm
+++ b/code/game/objects/items/rogueweapons/melee/blunt.dm
@@ -13,7 +13,6 @@
 	icon_state = "instrike"
 	item_d_type = "blunt"
 	intent_intdamage_factor = BLUNT_DEFAULT_INT_DAMAGEFACTOR
-	//We want chipping, m'lord.
 	blunt_chipping = TRUE
 	blunt_chip_strength = BLUNT_CHIP_WEAK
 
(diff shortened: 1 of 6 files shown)


ProbablyCarl:
I'd plans to make a nochip integrity damage modifier, but I can't be bothered right now. This works.

maker-of-nightmares:
This is a horrid change. Blunt weapons are already nearly entirely garbage with the only one worth a damn being the shock mace and mauls for special attacks. The, cudgel, iron mace, and steel mace are literally useless because you can't out DPS anything them crushing through plate is literally their only benefit.

Like you can barely hurt dodge experts with it and any class that specializes in maces literally can't use a sword to beat a dodge expert like the balance wants.

anconfuzedrock:
can we change the NONBLUNT_BLUNT_DAMFACTOR too? currently at 0.8 it means the butt of a halberd is BARELY more damaging to armor than the point will be (but way less accurate). it should just be 1 at this rate.

anconfuzedrock:
I'm pretty sure you're confused on a few things, Tekko. for one, only pure unarmed gets uncapped strength, so knuckle dusters don't have such a bonus. also I can grapple and hit them with a club too, really doesn't matter.

PR #748, opened 2026-02-06
Rage for barbarians and berserker

Fatherless1Xx:
## About The Pull Request

gives berserker and barbarians a rage spell

## Testing Evidence

image

## Why It's Good For The Game
i am a berserker main

### DIFF
--- a/code/datums/status_effects/rogue/roguebuff.dm
+++ b/code/datums/status_effects/rogue/roguebuff.dm
@@ -1281,6 +1281,29 @@
 	to_chat(owner, span_warning("My mind is clear again, no longer clouded with foggy peace!"))
 	REMOVE_TRAIT(owner, TRAIT_PACIFISM, id)
 
+/datum/status_effect/buff/barbarian_rage
+	id = "barbarian_rage"
+	alert_type = /atom/movable/screen/alert/status_effect/buff/barbarian_rage
+	duration = 30 SECONDS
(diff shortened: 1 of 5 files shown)


anconfuzedrock:
are you crazy? berserker is the strongest wretch. masked lunatic is basically just a significantly weaker berserker with better gear and it's STILL too strong. they'd be too strong even if they didn't get master weapon skill, and they can!

I don't even know what you're trying to insinuate with the "ironchad"

SneakySlinky:
"Does not get expert weapon skills"? I mean, I guess so, by technicality, it can choose to get master unarmed instead, if you're so inclined (why wouldn't you)?

Sirgir:
You are already more than strong enough as a Barbarian/Berserker. Hell, Berserker can run laps around the garrison and punch them to death. Critical resistance + immunity to pain is enough to shrug off the worst of wounds and run off scott-free while you let your crit resist tokens regenerate.

Barbarian/Berserker is already the, "I grab you then headpunch you until you skullcrack" role (especially if you take something like Struggler where your CON/STR/WIL is so high that people literally cannot resist out of it). They don't need any more STR when you can pump it up with statpacks and pick a STR race. If you really want to pump yourself up, you can just do what Osprey says and take Devotee and be a Graggarite. (Even then, you really don't need to do that).

CloverIsLucky:
Wawawawa stop crying folks. Let's go _ragin'_ with DnD/Pathfinder barbarians!

PR #757, opened 2026-02-08
Returns ye olde Longshoreman by way of Towner Thug subclass!

Jamdrawers:
## About The Pull Request
They're mostly the same as the Goon, though they start with a cutlass, a slightly different outfit, and lose some skills like mining and farming in exchange for fishing and swimming.

Also a few other tweaks to Thug

-Adds the Booze Healing to all thugs and a beer in the satchel. 

-Adds Mace option for Big Guy as well as giving baseline skill in maces and axes.

## Testing Evidence
image

## Why It's Good For The Game

-A bit of nostalgia for Ratwood 1's worst class.

-I think it'd be fun and flavourful but I can snip it out if it turns out that drunk healing is actually super powerful. But given the lack of armour skill or dodge bonus for the role their survivability probably won't be a big problem

-a big guy should be able to thwack people with sticks

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/pilgrim/thug.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/pilgrim/thug.dm
@@ -4,7 +4,7 @@
 	allowed_sexes = list(MALE, FEMALE)
 	allowed_races = RACES_ALL_KINDS
 	subclass_social_rank = SOCIAL_RANK_PEASANT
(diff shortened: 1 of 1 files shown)


ProbablyCarl:
Drunk healing was meant to be exclusive to dwarves, but this might be funny, I suppose.

CVSthePharmacy:
DRUNK HEALING IS ONLY FOR DWARVES!!! I HATE YOUUU

PR #761, opened 2026-02-08
Mpreg vice

IronDragoon:
## About The Pull Request

Adds 'Marked by Baotha' Vice. This uses the same code as the Baothan Mpreg ritual, and applies it to your character roundstart. This means the character has the groin marking and will be able to be impregnated regardless of genitalia. This is purely for RP/Flavour and should not affect balance in any way, it has been requested by some fellow freeks who want to have a character be cursed with this permanently. 

It now also applies a stronger version of the nymphomaniac trait. I wasn't sure the best way to do this, please take a look at the code, but from testing it seems to work as expected.

## Testing Evidence

image

image

image

image

(Groin tattoo marking visible) 
image

## Why It's Good For The Game

Fun freek ERP options are good and don't negatively affect anyone else. I especially want this because we have no 'evil' ritualist classes beyond wretch, so the mpreg ritual is *really* quite difficult to actually get to use.

### DIFF
--- a/code/__DEFINES/traits.dm
+++ b/code/__DEFINES/traits.dm
@@ -97,6 +97,7 @@
 #define TRAIT_WAGES_SUSPENDED "Wages Suspended" //Stops nerve master daily pay for this guy
 #define TRAIT_SCALEARMOR "Weathered Scales"//Mage armor, refluffed to scales.
 #define TRAIT_HEMOPHAGE "Hemophage"//You can drink blood for heals, but normal food and water makes you ill. Shitty vampire.
+#define TRAIT_BAOTHA_FERTILITY_BOON "Marked and shaped by Baotha" //Able to be impregnated, has permanent womb tattoo and stronger version of nympho vice
 
 //Hearthstone port (Tracking)
 #define TRAIT_PERFECT_TRACKER "Huntmaster" //Will always find any tracks and analyzes them perfectly.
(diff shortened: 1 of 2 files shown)


royalbeehavior-arch:
I think that all vices should have a drawback.

Quink-Fix:
I believe when it has been suggested before, a drawback a la nymphomaniac was suggested.

code/__DEFINES/traits.dm
```diff
@@ -489,6 +489,7 @@ GLOBAL_LIST_INIT(roguetraits, list(
 	TRAIT_WAGES_SUSPENDED = span_warning("The Stewardry has halted my daily wages!"),
 	TRAIT_SCALEARMOR = span_info("My scales can protect me from harm every so often."),
 	TRAIT_STANDARD_BEARER = span_info("The banner is my lifeline. Just as I am to it. The retinue know to rally around me, so long as I keep it safe."),
+	TRAIT_BAOTHA_FERTILITY_BOON = span_info("I have been marked by Baotha. I am branded visibly on my groin and am able to be impregnated regardless of physical states that would usually prevent this")
```
ProbablyCarl: End this line with `,`

Heheniy:
> It'd be awesome if the rune force-set your pantheon to Baotha, but players seem not to like such heavy limits.

just convert them after or before

Murderfoxen:
> really doesn't seem like much of a drawback. the inquisition isn't really going to do much about this, and the chances of them ever actually seeing it are pretty low unless you're already getting naked with them

My first few rounds when mpreg was introduced, I was in fact, found by the inquisition.  Maybe people just dont care enough to make it an issue. But it can be Roleplayed well. It's just most people don't want to engage with 'fetish' content canonically. :)

Quink-Fix:
praying for your upfall oomfie!!!

Heheniy:
why should person be punished more then nymphomaniac if mark itself not a buff

PR #768, opened 2026-02-09
Makes Riding Actually Trainable

ophaq:
## About The Pull Request
Riding XP is now gained every 5 tiles you move via RUN intent on your mount. It is scaled off your INT like all skills, giving you 0.1 XP for every INT you have (for example, 20 gives you 2 XP).

## Testing Evidence
ridingxp

## Why It's Good For The Game
The riding skill could not be trained at all, for some reason. This is a solution to that. Instead of how running works where you are constantly getting XP and based off stamina depletion or something, you are earning XP based off distance traveled while running on your mount. This means you'll have to actively zoom around and do circuits or something to train it, as it should be when learning how to ride. Also, it takes a while to get skills like athletics. Hopefully we see people using more mounts, now that people can actually learn how to ride instead of praying for a dream to get novice or use equestrian and then rush the archivist to train to journeyman.

### DIFF
--- a/code/datums/components/riding.dm
+++ b/code/datums/components/riding.dm
@@ -3,6 +3,7 @@
 	var/last_move_diagonal = FALSE
 	var/vehicle_move_delay = 2 //tick delay between movements, lower = faster, higher = slower
 	var/keytype
+	var/riding_xp_move_counter = 0 //counter to reduce XP spam - award XP every 5 moves
 
 	var/slowed = FALSE
 	var/slowvalue = 1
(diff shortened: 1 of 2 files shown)


ProbablyCarl:
I saw this and thought it would be the other riding to gain XP PR. You'd worried me there, a moment.

ophaq:
> Riding, much like Wrestling, is untrainable due to balance concerns. Saigas and goats, especially at high levels of Riding skill, are monstrously powerful. I don't know if we should be making it trainable without putting a bit more consideration into it. Maybe limiting the skill gain up to apprentice or journeyman without an appropriate trait would ameliorate this.

I wouldn't mind trying to implement that but it is pretty darn slow to train regardless, much like athletics. I could make the XP gain even lower if necessary. Otherwise if it goes the skill limit gain to journeyman, there might need to be a new trait implemented like homesteader uses or something.

Skyglazer:
>  Otherwise if it goes the skill limit gain to journeyman, there might need to be a new trait implemented like homesteader uses or something.

Yeah, I was thinking a trait like that would be ideal, if it wouldn't be too challenging to implement. Mind that it's less difficult to train than athletics because it doesn't take any stamina to sprint on a saiga. Will have to feel it in-game to tell how easy or hard it is to actually gain levels, though...

PR #776, opened 2026-02-10
Remove Justicar's Sight

OspreyIsHere:
## About The Pull Request
Ravox gets two traits: one makes weapons lose sharpness slower. Cool!

The other is the Validhunt Sight.
Whenever you examine someone's face (therefore, *mask off*), you CAN and WILL see if they have a bounty on their head. It leads to very much little RP - and mostly just going "mask off 321 mask off now now" then killing you.

At least have the decency to RP and actually analyze your targets.

## Testing Evidence
Its removing a single trait bruh

## Why It's Good For The Game
Validhunt sight does not incite RP - it's just a way to lock on someone and think about how to efficiently break their legs to take them to get paid. At least go to an excidium and get the scroll and actually cross-check people by appearances.


Spaghetti-bit:
Birb approval.

PR #781, opened 2026-02-10
Assorted Merchantry tweaks

Jamdrawers:
## About The Pull Request

Makes a smattering of changes to cargo/packsrogue - there's a lot of stuff that I think is either overpriced or underpriced and there's probably a lot that is missing, or things that have changed without the merchant tab being updated to compensate.

What I've put down so far is:
1) Made ozium and moondust a little more expensive. Drugs should be an expensive habit!
2) Made merchant boozes of all kinds moderately cheaper. Also just snipped out a few "Xhongese" drinks that were just, I think, cluttering up the list. I don't thing 'Xhongese' is even a place in this setting.
3) Made Shhig brand zigs a bit cheaper - I priced them initially when I ported them a while ago and hadn't really accounted for the merchant fee that'll get taken out
4) Made the Silver Dagger a lot cheaper - Silver weapons have been changed to be much less powerful but more numerous since in unrelated updates, but the merchant tab hasn't been updated to reflect this. Also feel it'd make sense to not put it in the 'luxuries' tab. Weapons (foreign) maybe?
5) Making some generic Wardrobe options (default robes, default plain dress) cheaper and some other ones (Fancy Formal dress) more expensive
ROUND 2
6) "six foot" and "eight foot pole" items clarified in name, gently reduced in price
7) Slightly reduce price of Mess Kit since nobody buys those
8) reduce the price of Rain Cloaks to not be crazy high
9) Mentha zigs a smidge more expensive
10) Fancy Tea-sets a smidge less expensive
11) Silver Psycross half as expensive (120 down from 250) - Same as silver dagger, silver is just not as big a deal now
12) Adds Sui Dust for 135
13) Can buy normal bottles as well as alchemical vials
14) Lipstick is a fair bit cheaper overall
15) Hair Dye is now under Cosmetics instead of Apparel
16) 'Consumables' now called 'Food' because you consume potions too!
17) takes drugs and magical supplies out of 'Luxury' and in to 'Substances' and 'Magic' so it's not so cluttered
18) Renames 'Weapons (Foreign) to Weapons (Exotic) and puts the silver dagger on the list

Will be a little bit of overlap in some of these but I think it's fine. Silver daggers are in both the 'magic' (since I understand it's used in magic rituals) and also in 'weapons (Exotic)' because it's an exotic weapon.
Open to any suggestions or feedback.

## Testing Evidence

Numbers numbers
image
image

## Why It's Good For The Game
(Usually explained in each point)

### DIFF
--- a/_maps/map_files/rockhill/rockhill.dmm
+++ b/_maps/map_files/rockhill/rockhill.dmm
@@ -377,11 +377,6 @@
 /obj/structure/roguewindow,
 /turf/open/floor/rogue/hexstone,
 /area/rogue/indoors/town/shop)
-"agg" = (
-/obj/structure/flora/roguegrass,
-/obj/structure/flora/roguegrass/bush/wall/tall,
-/turf/open/floor/rogue/grass,
-/area/rogue/outdoors/exposed/town/keep)
 "agj" = (
 /obj/structure/fluff/railing/border{
 	pixel_y = -7
@@ -422,8 +417,7 @@
 /obj/structure/arcyne_wall,
 /turf/open/floor/rogue/greenstone,
 /area/rogue/indoors/town/magician{
-	name = "Mage's Tower";
-	first_time_text = "Tower of the Magos"
+	name = "Mage's Tower"
 	})
 "agA" = (
 /turf/closed/wall/mineral/rogue/tent{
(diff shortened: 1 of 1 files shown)


Jamdrawers:
> As one of the few merchants I give this a big thumbs up, the changes are small but all make sense. I don't think I've had any price debated as much as the silver dagger, you can sell it at a loss and people will still be gobsmacked.

Yeah I think the crazy high price on silver daggers *used* to make sense when silver weapons were super scarce but were a big instant mega Fuck You button when you used it, but ever since ~some time ago they've been made way more common and way more gentle in effect.

PR #789, opened 2026-02-11
Reflections attempt #3

Fatherless1Xx:
## About The Pull Request

Takes Useroth's code and tries to patch out the bugs: github.com/Rotwood-Vale/Ratwood-2.0/pulls - keeping it vis_contents focused
Adds a very janky mirror reflection that needs more work
i can make the reflections better i just want to be sure that:
A) This doesn't glitch out
B) it doesn't end up being expensive for the server

works fine locally but i won't know for sure until it's live tested, once we've confirmed this actually works without glitching out i can polish it

## Testing Evidence

image
image


CaydenT:
It's appearing on random surfaces like carpets and stone tiles and door frames, creating more moving objects that screw with people. Going to be a huge mess in a crowd.

PR #796, opened 2026-02-11
Adds Sui Dust to the Purity vendor

RemervKeberb:
## About The Pull Request
Does what it says on the tin. adds the genderchange dust to the Purity vendor at the price of roughly the same as the other fancy vanity items like lipstick.

## Testing Evidence
image

## Why It's Good For The Game

Sui Dust is purely an RP drug that is barred behind journeyman alchemy. Meaning it's at best a somewhat annoying timesink grind to attain and at worst not something you can even attain at all if you just want to have some HRT fun.  
As is, crafting it is the only way to attain Sui Dust. Having it in the purity vendor not only makes it accessible to the interested crowd in a way that lets them invest their earned coin throughout a round in order to get to enjoying the RP side of it longer. But it also shows them it's a thing in the first place. It took me two months of playing before I ever even heard of it and that's a travesty...


halbdog:
Maybe a price more like ~200 mammon would be better? It isn't really that much more expensive with how inflated mammon is, but still encourages people to go talk to an alchemy role to get it for cheaper, leading to more interaction, more RP. It's the way I've gotten it before.

Also, the dust does have a another use, it's one of the 3 items required to upgrade the Eora Tree miracle, and that price puts it on par with a Rontz from the merchant, the only item of the 3 that can be bought currently. And.. the idea of lipstick and dust that transforms your whole body costing the same ammount is just a bit goofy.

PR #842, opened 2026-02-17
Removal of azure slop staves

Heheniy:
## About The Pull Request
Removes staves combat skill (which is literally used by 2 weapons) and replaces it with polearms

## Testing Evidence
idk, it compiles?

## Why It's Good For The Game
less azure slop


Heheniy:
it is literally not used in that way

if you look which else skills had roles with staves skill you will see that almost all of them have another combat skill on same level or even higher then staves

this skill is literally a peak of slop

Pidgey-chan:
They made spears suck so anyone using them is a nonissue

PR #845, opened 2026-02-17
Lycker mage gaming

HalfAwake691:
## About The Pull Request
Allows wretch lycker to become a sorcerer 

## Testing Evidence
image

## Why It's Good For The Game
Having variety in lycker seems like a good idea, I think im tired of seeing my one-hundredth lycker barb in a row. 

Plus they're magical creeturs by nature and might lead to some interesting roleplay


ophaq:
This wouldn't be a bad idea if there were penalties such as being mouth grabbable and not having infinite stamina for this class, but then that might not make sense.

HeroWithYay:
Infinite blue class probably shouldnt have magic.

PR #860, opened 2026-02-19
Silverface now makes some profit that the merchant can withdraw

Xander3359:
## About The Pull Request
Some of the profit of the silverface sales go to the merchant (Stored in the machine itself)

## Testing Evidence
image

## Why It's Good For The Game
It's less annoying that people use the silverface if you can actually benefit from it.
This allows people to buy from the silverface if they are willing to tank the upcharge when the lines get long at the merchant (or they forget to unlock it when they AFK)

### DIFF
--- a/code/modules/roguetown/roguemachine/merchant/_goldface.dm
+++ b/code/modules/roguetown/roguemachine/merchant/_goldface.dm
@@ -31,6 +31,10 @@
 	lockid = "merchant"
 	// Which job can access profit from this vendor
 	var/profit_id = list("Merchant", "Shophand")
+	/// How much % of the base price is added as profit
+	var/profit_margin = 0 // Goldface makes no profit since it's the merchant buying at the best price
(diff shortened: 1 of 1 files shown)


Jamdrawers:
I see the profit margin is 0.5? Is that 50%? Because that'd be crazy for something that requires literally 0 attention.

LennyfaceGaming:
I have to disapprove of it, mostly from the merchant's PoV.
There is already a vendor the nets profits to the merchant- It's called Peddler. That is if they use it, few minutes of preparation at round start and routine checkups go a long way.
Silverface, in my opinion at least, should not reward the merchant for doing nothing, if anything it should net no profits to incentivize the merchant to actually stock up basic supplies in the Peddler and save time and money wasted for all- The Silverface should, at best, be an expensive bandaid for the customers, not a reward for low effort/non present Merchant.

PR #879, opened 2026-02-20
DESERT TOWN - (Being intermittently tested)

Jamdrawers:
## About The Pull Request

Adds a WHOLE NEW MAP!!! With a WHOLE NEW THEME! Complete with an ARRAY of CUSTOM ROLES with CUSTOM OUTFITS!
AlAshurmapEA (not finished)

Some of these roles are Reskins - the SULTAN is the ducal equivalent with a different outfit, while in the place of the dungeoneer the SLAVE MASTER takes the stage as a fancy power role with one of the swankiest apartments in the keep, catching prisoners and putting them to work. Rather than Servants and a Seneschal you have SLAVES led by a HEAD SLAVE - all subservient to the slave master, of course.

Will be lots to iron out and balancejack but it'd be ideal to get this put in for now so that it can be kept updated with the assets and stuff

## Testing Evidence

081b0053d4cac4c6af943157cd4d36e1
sultan
image

## Why It's Good For The Game

It's really cool! And we've been waiting for it for -ages-


AlucardHellsingChapel:
We should rename bandits into crusaders

Jamdrawers:
Alright so the big issue is related to z-level configs, where the map briefly only has 1 z-level.

Second issue after that is that while setting roles to be map-specific-only does keep them from showing up in the class-select menu, they're still showing up in the Latejoin menu

JustMeTheIInd:
I would hate to intrude upon something that's not of my make, and I still consider the entire map to be great looking and very interesting to play, but from my brief overview of the Map in StrongDMM, it seems that the Inquisitorial Mansion has literally no entryways other than the front door, or just mining through half a mountain, which makes it pretty tough for any bandits and/or wretches that might want to break out their comrades out of Inquisitorial custody. Maybe a cave leading to the cells, like Dunworld has, could work?

PR #897, opened 2026-02-23
A Respectful Motion to Raise Adventurer Slots

Fatherless1Xx:
## About The Pull Request

With the utmost respect and in a spirit of sincere civic duty I would like to resubmit a humble proposal to the honorable court of Ratwood players that adventurer slots be raised to 99.

This petition was previously denied. However the maintainer who rendered that decision has since departed. As such I believe the matter is worthy of renewed consideration: https://github.com/Rotwood-Vale/Ratwood-2.0/pull/847

On the first occasion I regret that I did not treat this grave and weighty matter with the solemnity it so clearly deserved. I therefore return now in my serious voice. There shall be no frivolity nor any conduct less than wholly serious in this discussion. There shall be only the seriousness that so serious a subject demands.

## Testing Evidence

Having submitted this PR to the scrutiny of multiple expert teams I am pleased to report that they were unanimous in their conclusion that it operates exactly as intended.

## Why It's Good For The Game

It is my humble opinion as a member of this serious and distinguished community that adventurers provide more to the average round than vagrants or towners. Under our current player count normal there are frequently not enough slots to go around, forcing people to lock into a role they don't care for or to disconnect.

I do not believe it is healthy for the game to block people from playing the roles they want to play most. This is a game after all and not an occupation.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/adventurer.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/adventurer.dm
@@ -8,8 +8,8 @@ GLOBAL_VAR_INIT(adventurer_hugbox_duration_still, 3 MINUTES)
 	flag = ADVENTURER
 	department_flag = WANDERERS
 	faction = "Station"
-	total_positions = 20
-	spawn_positions = 20
+	total_positions = 99
+	spawn_positions = 99
 	allowed_races = RACES_ALL_KINDS
 	tutorial = "Hero of nothing, a wanderer in foreign lands in search of fame and riches. Whatever led you to this fate is up to the wind to decide, and you've never fancied yourself for much other than the thrill. Some day your pride is going to catch up to you, and you're going to find out why most men don't end up in the annals of history."
 	class_categories = TRUE


Pidgey-chan:
Hmm. Yes, quite. Indoobietably.

halbdog:
I've heard of admin-shopping before, but never maintainer-shopping. This PR was rejected for good reason once before.

The recent dramatic lowering of rp quality is no excuse to sink to their level, and pushing for more isolation will only make the problem worse.

coffeecoffeecoffee:
> I've heard of admin-shopping before, but never maintainer-shopping. This PR was rejected for good reason once before.
> 
> The recent dramatic lowering of rp quality is no excuse to sink to their level, and pushing for more isolation will only make the problem worse. But if you insist, maybe a more reasonable number like 30 slots, 40 slots would be better? 99 is just.. excessive. Anyone who played lowpop can tell you how much a round suffers from being 80% adv.

What? Those are the absolute best rounds.

Hmeister-real:
Allowing almost 100 people to play as a free-thinking, unimportant roles with 0 responsibilities and stats that allow them to put up a good fight against most roles alongside the option to have the versatility of towner roles to a slightly weaker level (blessings, magic), on a server with 120 pop, is baffling. Even the commonly discussed 30 adv cap buff is too much imo, this PR is a joke and if it got merged the RP quality would plummet.

TamedAchilles:
Polishing a turd doesn't turn it into gold, it just remains a turd unfortunately.

We quite literally, infatically, do not need 99 adventurer slots. If you only play Ratwood JUST to play Adventurer, this isn't the place for you. We have A PLETHORA of other roles begging to be touched. 

Break your bubble of non-RP slop and interact with people for once.

PR #911, opened 2026-02-25
The MIASMA (Buffs/Debuffs in terrorbog)

MrAsianFunTime:
## About The Pull Request
(DO NOT TM OR MERGE UNTIL CONSORT SUBCLASSES FULLY TESTED AND MERGED!)
(May cause issues, may not, currently is set up to have new Consort Subclasses in it to save for future issues, but better safe then sorry)

The Miasma adds back in debuffs for certain roles, buffs for some, and resistances (nothing) to the Terrorbog.

**Miasma Empowerment** - (STATKEY_SPD = 1, STATKEY_PER = 1, STATKEY_STR = 1, STATKEY_CON = 1)
Roles: Bandits, Skeletons, Wretches (Cant figure out good way to add to vampires just yet)

**Miasma Resistance** - (Nothing, fluff, shows you wont suffer in terrorbog)
Roles: Adventurers, Pilgrims, Migrant Waves, Mercs, Inquisition, Martyr, Prelate, Wardens

**Miasma** - (STATKEY_WILL = -2)
Roles: Towners, Church, Anyone else that doesnt quite get a Buff in the Town and Keep / Courtiers

**Sinister Miasma** - (STATKEY_INT = -2, STATKEY_PER = -2, STATKEY_WILL = -2)
Roles: Knights, MAA, Any other roles with Sworn Defender or Vigilant Guardsman

Also unadded the Lowlands from counting towards area for Sworn Defender and Vigilant Guardsman (Such had been there for the town and keep. Making the Lowlands other then wardens, the true neutral area when it comes to buffs or debuffs)

## Testing Evidence
https://gyazo.com/f6b21bd45aab3f6558b3f1372ecc9c0a
https://gyazo.com/61910931511155a9d80939bbdcbe1f82
https://gyazo.com/93ae4d6b242a2c5c46805c166dfcf852
https://gyazo.com/75712884bc084ddca103723d249276b2
https://gyazo.com/dfe0314ccaf212e949080bc9a8a28c63

## Why It's Good For The Game

First step in dissuading the Hyper Wars we always see at the Bridge into the Bog.
More Reason for Dukes, Towners, Other groups to use Mercs, Adventurers, Wardens and such to deal with issues in the bog.
Gives certain Antags some advantage if swarmed by keep and retinue and entire town still.

Encourages the Keeps forces to slightly think a bit more instead of simple charging in. Plan an attack or action.

Its the Bog, its full of evil stuff, time for some Evil Miasma as well.


Dedwolf:
mmmmmmm swamp gas
(can we call it something else?)

this is good, but i think towners should be unaffected

Pidgey-chan:
MAAs after spending their entire life 1 mile from the bog being somehow less used to it than foreign sellswords from the open plains of Grenzelhoft

Fatherless1Xx:
> 'guards have lower stats in the bog than in town' is just already a feature, (indeed it's a pretty aggressively implemented feature, the difference between the low-base-stats and the low-base-stats-plus-the-big-area-buff is pretty large!)
> 
> Gotta think not just about how powerful something should be in general, but how powerful it should be in some ways compared to others when you're making balance changes like these.

Losing a privilege is not the same thing as being debuffed.

Even without the guardsman bonus they still have seven stats. On paper that is equality, if you're generous and choose to ignore that they have expert in every weapon skill unlike every other class in the game who only have one.

In practice they still have a massive advantage and it has nothing to do with stats. It is the presumption of innocence from towners. It is the ability to buy effectively infinite health potions. It is having access to all the gear they could ever need. It is being able to use the scomstones to coordinate. It is by meta protections making it against the rules for antags to round remove you without you making the first move.

Jamdrawers:
> Losing a privilege is not the same thing as being debuffed.

If the baseline is adjusted to compensate then it is practically and mathematically identical. Y'know, like, 2+1=4-1

maker-of-nightmares:
This is just Sea Lich 2.0 NGL. The reason Sea Liches were such a nuisance is they had a direct and incredible advantage in an area no one else really had or the few who did weren't enough to solo the lich and his gang. 

Now you have basically half the map for the Lich and Evil Crew to just camp in because you physically can't engage them past the bridge and there's nothing you can do while they pot shot you with skeletons and spells. Like it's a minus 3 equivalent stat wise BEFORE you add they lose Vigilant Defender/Guardsmen buffs which is another 2 stats for a total of minus 5 for walking into the bog to engage the Evil Dude McGang. To put into perspective how genuinely awful in combat this would be going from 10 Int to 8 while your opponent remains ten means what was an even fight has just become A LOT harder. This is before you take into account the -2 Will on a suddenly tankier opponent. Not only are you fighting a target that is supposed to already be slightly stronger than you on average, that opponent has now become harder to bleed out and wound while you have become easier to feint and be feinted(potentially even by foes who couldn't previously due to stat differences), you now wind faster and easier due to lower Will and most importantly **lose spell casting potential due to the loss of the Blue Bar from a lower Will**. 

This while on paper is a great idea to keep towners inside and prevent deathballs into the bog the result will be inevitably the bandits will draw every encounter into the bog and whittle down the keep every single round and if you don't engage them during their fighting retreat they'll whittle you anyways in the town proper. **You've essentially made a lose lose scenario for playing the game as Keep and a Win Win scenario for the antagonists.** The better solution here would be to make them have a better incentive to stay in the town such as aura effects or spells that only work in town or near town like the Marshal getting a "Hold the Line" spell in town that is super strong but only works within high town for example. This makes it so the town wants to play more defensively so they can get their giga buffs from the Marshal/Sergeant while not punishing them for following them into the Bog to finish off a weakened foe this also more interestingly will help stop the endless hyper wars inside the town by making the Keep more of a reasonable threat instead of a bunch of tissue paper before the Church(Or at least make them more enjoyable)

PR #928, opened 2026-02-27
(TM ONLY) Enables Tracy

Kroto-Ker:
## About The Pull Request

Enables tracy.

Will make very large files that can take up server storage so use with caution.


Metekillot:
I would recommend conferring with the admins to handle enabling tracy through a server setting.

Metekillot:
After building on my system, as far as I know, this won't work unless you compile tracy specifically for Linux and include the libprof.so file in the repo root instead; assuming the server is hosting TGS on Linux

Useroth:
we still using that stuff

PR #935, opened 2026-02-28
Adds handholding

Xander3359:
## About The Pull Request
Adds handholding

## Testing Evidence
I tested on local and I'm pretty happy with how it came out.

## Why It's Good For The Game
This will save Ratwood it's essential.

### DIFF
--- a/code/_onclick/other_mobs.dm
+++ b/code/_onclick/other_mobs.dm
@@ -15,6 +15,8 @@
 		return
 
 	if(check_arm_grabbed(used_hand))
+		if(try_handhold(A, proximity, params)) // At this point the attack is blocked so we'll allow snowflake behaviour
+			return
 		to_chat(src, span_warning("Someone is grabbing my arm!"))
 		return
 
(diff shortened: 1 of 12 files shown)


AranyaPTH:
For the love of all that is good, YES!

Sgt-Solid01:
Peak

SmashyScotsman:
> Alleged bug someone stated in OOC: **Archimagirus:** Hand holding resets when going up and down stairs and swaps the active hand of the person holding the hand

Hello! This was me! To go into more detail. When holding hands with someone and going up and down a staircase it will drop the hand hold (But not the grab) and swaps the hand that the grab is currently active in. It can be easily resolved by getting the person following to redo the grab but targeting the opposite hand they had targeted before. 

It seems to consistently swap to the opposite hand of the last rather than particular one. For the sake of anyone wondering how it actually works here is a quick step by step as I understand it works. (Still getting the muscle memory down so may be incorrect slightly)

Person A grabs Person B's hand.
Person B targets the hand of Person A that is grabbing them.
Person B uses the GRABBED hand on touch and clicks Person A's sprite, This should then move their sprite to be holding hands.

RiceGateau:
handholding, where art thou..

Kroto-Ker:
Has been conflicting for 3 months.

PR #982, opened 2026-03-03
Return of the VANGUARD

Jamdrawers:
## About The Pull Request

The VANGUARD return as the poorly equipped, undertrained chaff we've always wanted them to be. Not TERRIBLE, they're still lightly armoured and have lower stats than the aloof, veteran rangers who go on long expeditions beyond the Terrorbog.

Warden slots are reduced to a more tight-knit elite force, while 4 rookie vanguard can fill the vacant space, bumming around, getting killed, complaining about their pay, and squatting at the bridge. Enabled only on Rockhill, Dunworld instead has the full 6 Warden capacity.

Wardens, being a usually more rare, are a teeny bit more elite, gaining skill in polearms meeting their skill in axes (so they can use that big cumbersome spear they have as decently as they use their convenient special axes, and giving them options to spend their coin on other than a giant double-ended axe every time - the objectively correct choice every time. And the forrester starting at Apprentice bows so that they can shoot from the battlements if they happen upon a bow.

Ready to test now that I figured out what was stopping latejoin spawns!

## Testing Evidence

image
image

## Why It's Good For The Game

There was just a certain Something Something to it that you don't get from being the cool elite skull-wearing baddasses that are Wardens.


Pidgey-chan:
Not sure about having Wardens and Vanguards both, especially if one of them gets to be the cool one. But really we need neither of them and Bog Guards instead.

n3tis:
> > I'd say wardens being elite would make more sense if they had better skills than the vanguard's archers. Currently the rangers seem to be.. rather forgotten in this PR. They don't have a high skill as the men at arms, and the foresters can easily level up to match the rangers at a much faster rate now, since they start as apprentices.
> 
> Warden rangers only get expert? Oh, right, weird.
> 
> Though that said, my impression has been that bow skill isn't _that_ important (especially compared to melee weapon skill), and notably vanguard archers don't start with the super mega juiced up Perception bonus that wardens do (like +5 with the buff?) which I think is more important?

Um... warden rangers get 2 perception while the vanguard gets 3, unless if something was changed and the 'class selection' infowindow wasn't updated. And both have the same talented woodsman trait... so.. uh.. the vanguard archers here have HIGHER perception than the wardens. Depending on the statpack, they have a more notable mega juiced up perception.

n3tis:
> Unless I messed up Vanguard get the confusingly named talent that uncaps tanning and lumberjacking while Wardens get both that and the similarly named talent that gives you another +3 perception on top of that when outside of hightown. Maybe they should get Expert Homesteader instead, though neither is super necessary.

The trait you're talking about is Expert Survivalist. The one that gives them +3 perception outdoors is called Talented Woodsman. And both of them have those. I'd suggest actually tapping into the ranger class to truly make it 'elite' if that's what you're going for, cause right now, they're no different than the Vanguard, and in a comparison, they're actually worse.

halbdog:
As a vanguard archer, I spawned with a dagger sheath but no dagger, and also no houndstone - despite a vanguard footman spawning with his. Otherwise, this was lovely.

PR #992, opened 2026-03-05
Gives halberd cuts the same damage mod as the bardiche

Pidgey-chan:
## About The Pull Request

Gives halberd cuts the same 1.2 damage modifier as bardiche cuts, up from 0.9.

## Testing Evidence

One line, take my word for it

## Why It's Good For The Game

The bardiche, the iron variant of the halberd, shouldn't have a better cut than it does, considering both of them are giant axes. As it stands, the halberd is one of the worst steel polearms, with poor throwforce, poor cuts, and only a bad chop and the usual spear stab to make up for it. It does nothing exceptionally well. This should make the halberd more versatile and viable as a weapon pick.

### DIFF
--- a/code/game/objects/items/rogueweapons/melee/polearms.dm
+++ b/code/game/objects/items/rogueweapons/melee/polearms.dm
@@ -70,7 +70,7 @@
 	sharpness_penalty = 2
 
 /datum/intent/spear/cut/halberd
-	damfactor = 0.9
+	damfactor = 1.2
 
 /datum/intent/spear/cut/scythe
 	reach = 3


Pidgey-chan:
> the bardiche is balanced totally differently from the halberd, it's not really comparable; the halberd isn't a steel variant of the bardiche, it's more of a variant of the billhook, which is a steel variant of the spear. Compared to the bill it has a chop but is worse one handed. It's perfectly fine, some polearms do different things.

bardiche

Pathed from halberd, description says "variant of the halberd", it's made of iron, it's the iron halberd variant. The halberd's chop is too useless to make up the identity of the weapon. The halberd at present cuts for less than 1 point of damage more than the boar spear, a weapon obviously not shaped for cutting, and quite a bit less than the glaive. There is no extra defense over the boar spear or billhook.

A 1-tile 30 AP chop does not do anything meaningful for the halberd besides marginally hasten the collection of goblin heads. It is a weapon carried by guards who spawn with it and no one else, because it offers nothing of note over the boar spear. It pairs worse with shields than the billhook. However the bill is kind of like objectively worse than the boar spear actually. It's been like that for a while. Could be addressed separately. 1.0 or 1.1 could be a good modifier for bill cuts, part of the purpose of the hook is to slash with.

Polearms doing different things is the point of this PR. With this change, the halberd becomes a viable alternative to the boar spear, while remaining worse than the more expensive glaive, which offers both this improved cut, the damage of a boar spear stab, and superior defense.

PR #1005, opened 2026-03-05
lets knights pick old again

Draganfrukts:
## About The Pull Request
This PR _restores_ the ability for knights to be old.

## Testing Evidence
It will work.

## Why It's Good For The Game
It makes sense. A knight's skill comes from experience, and knights are career soldiers.
It's for roleplay.
"But what if we're attacked and we have an old knight instead of a delicious adult fragger?"
IDFK dude he still has master and full plate and a cracked statline don't ask me


Jamdrawers:
I figure that's *kinda* what Veteran is for

Pidgey-chan:
Upvoted for old 15 int heavy knight feintmasters

PR #1006, opened 2026-03-05
No more kevlar gambesons

Pidgey-chan:
## About The Pull Request

Changes the piercing protection values of most armors to be higher for plate and metallic armor than fabrics

## Testing Evidence

WIP

## Why It's Good For The Game

Linen shouldn't be better at stopping arrows than steel plates

### DIFF
--- a/code/__DEFINES/roguetown/armor_defines.dm
+++ b/code/__DEFINES/roguetown/armor_defines.dm
@@ -80,38 +80,36 @@
 
 // Light AC
 // In general, light armor should always be vulnerable to stab, somewhat weak vs high pen slash (for now).
-// Also good protection vs arrows.
 // Minimal blunt rating should be 50, because normal blunt multiplier is 1.4, thus making sure it trade poorly
 // Capped to 90 blunt rating. This means light armor will take approx. 75% damage from blunt-weapon blunt attacks.
 // So blunt weapon can still be used against light armor but just not as effective.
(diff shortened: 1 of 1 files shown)


Jamdrawers:
Maybe it'd just be a historical accuracy thing but it seems to be the case that chainmail was kinda bad against arrows (compared to plate which was ridiculously good against arrows)

Pidgey-chan:
> Maybe it'd just be a historical accuracy thing but it seems to be the case that chainmail was kinda bad against arrows (compared to plate which was ridiculously good against arrows)

Mail was certainly worse than plate against arrows and these (mostly eyeballed) values reflect that, but pound for pound chainmail ought still be better than linen. Mail was a primary armor for over a thousand years, during times where spears and arrows were the largest battlefield threats- it should be decent against broadheads at least. With my values only 5 damage from a bodkin is stopped by mail. Honestly, I think I still left the padded gambeson too good as an underarmor option.

Ofc mail worn as a "primary armor" was heavier than what you'd wear under a cuirass, often, but there's no distinction between mail in this game besides iron and steel.

Pidgey-chan:
This PR leaves the legs without decent arrow protection for medium armor users, which, granted, is no different from how you can't protect the hands feet and eyes with any normal armor combination at current, but I'll have to think on a solution for that.

Jamdrawers:
What about splint..

PR #1012, opened 2026-03-06
Revert "AP Port: Sweetspots! Kind of!"

Pidgey-chan:
Reverts Rotwood-Vale/Ratwood-2.0#593

## Why It's Good For The Game

Un-cucks spears, which are borderline unusable in an actual fight with the sweetspot changes, especially in tidi. No one asked for these, bye-bye.


vitoras228322-ai:
Yeah, who would've thought that a most meta-pick weapon with the ability to pierce through almost all armor would feel slightly worse after the nerfs.

We don't need it back, not at all.

Pidgey-chan:
> most meta-pick weapon
> ability to pierce through almost all armor

Boy, do I have news for you!
Screenshot 2026-03-06 040710

Pidgey-chan:
This is practically a bugfix, I invite all doubters to observe warhammers or the longbow. Spears once offered one of few examples of "partial penetration" in RT, making them perhaps the most balanced weapons the game ever saw... but some people seem to think that 3 damage per point of STR through mail equates to perfect anti-armor. I wonder if they actually read any numbers?

It's laughably easy to rush down a spear as is, given that walking forwards is generally faster than backwards. Spears were left in an altogether pathetic state by Carl before his resignation, and desperately in need of a reworking... how they were was pretty good.

Kroto-Ker:
Don't know if this is gonna get through. (Closes it?)

Pidgey-chan:
> > > most meta-pick weapon
> > > ability to pierce through almost all armor
> > 
> > 
> > Boy, do I have news for you! Screenshot 2026-03-06 040710
> 
> Ah, yes, the swing with such an absurdly high miss cost AND delay AND 1-range that you won't actually hit if your opponent just moves...
> 
> ... Which is the same thing as the spear sweetspot! Wow! Thank you for bringing up an equivalent in balance as to why sweetspots are good!

One of these has 30 more AP, is one handed, fits on the belt, and it's not the spear. The silver whammer in particular offers the same force as a wielded spear in one hand(!) I don't think they're comparable at all. 1-tile range is frankly an advantage now, it's easier to facehug someone than to stay at exactly 2 tiles consistently.

PR #1028, opened 2026-03-08
Dragon War Banner(GIVE ME FEEDBACK)

RiceGateau:
## About The Pull Request

Adds dragons zeal, A banner that when planted in the ground give bonuses to bandits in a 7 tile radius, this radius is line of sight so if there is a wall or block of vision, it removes said buff. This item is exclusive to iconoclast but can be placed by other bandits, iconoclast is the only person who can punish other bandits with it though. This should be used like a stand and fight item.

The stats as the are at the moment : +1 stats and 1+ blue energy per tick.

## Testing Evidence

Not tested currently, spriting atm so there is probably bugs in the code. This is here purely to receive balance feedback about the stats and bonuses so I can adjust them. If you downvote this while in draft form, you like getting gang banged by deadites.

## Why It's Good For The Game

This should add cohesion and solidify the iconoclast as sort of the direction of the bandits which is kind of needed sometimes. It will also inform bandits of where to rally if they have their ring off or are just lost. The punishment ability is primarily fluff just like the punishment on the master rod but it adds flavor and kind of an IC explanation of why bandits even have a form of camp or structure in the first place, when in doubt, point to god.


kimesik:
Your PR doesn't compile. Double-check everything.

PR #1031, opened 2026-03-08
Cityguard for Rockhill

Jamdrawers:
## About The Pull Request

On Rockhill, Man-At-Arms slots are reduced, and in their stead is the City Guard, wearing classy blue uniforms that project Authority and Order, armed with stunmaces, mauls and crossbows. Their jurisdiction is more squarely the city, with a focus on keeping order and capturing criminals.
On Dunworld they are disabled, with MAA slots left untouched. On Deserttown it'll be completely different.

Also bundling in the ROOKIE. A one-slot role that joins the City Guard in their Guard Duties. They have bad stats, basic armour (but not worse than what the armoury has spare to save you the inevitable first-day-armoury-visit) and sub-expert skills. Though, compared to the regular guards, they do have a smidge more utility in the form of a smidge of crafting and cooking as well as Squire repair.

## Testing Evidence
image
image
image
image

image

Bundling in the Rookie
image
image
image
image

## Why It's Good For The Game

1) I think it gives the city some character
2) I think the uniforms were neat
3) I think in the past it would have more reasonable to not add these because they were historically underplayed, but since the population these days is quite a lot higher I think we can afford to split things up a bit more.
4) It doesn't increase the number of garrison slots overall since it reduces MAA slots to compensate, so it shouldn't swamp antags or anything. - EDIT - except for the rookie, who is very weak on purpose.


JustMeTheIInd:
The idea is great, albeit I'm unsure whether having possibly 6 stunmaces in circulation (4 City guard + 1 spare + 1 catchpole) is gonna be that good of an idea. They seem limited for a reason.

Pidgey-chan:
What's with the Oblivion guard outfits?

PR #1044, opened 2026-03-09
Changes skill leveling for ranged implements to be consistent with other skills.

n3tis:
## About The Pull Request

Changes the maximum skill ceiling for crossbows and bows from EXPERT to LEGENDARY so that they match slings as well.

## Testing Evidence

image
After some successful bowmanship, the change works without a problem, as it's a mere edit of two lines.

## Why It's Good For The Game

The skill ceiling should be consistent with how other combat skills are leveled up. Slings can be leveled up to legendary but not crossbows and bows; this simple change rectifies that oversight.
As an alternative; a custom trait could be made to give experience past EXPERT to 'ranged' classes.
I would be eager to try and create one if that would be more satisfactory to others. Might need a hand in implementing it, though.

### DIFF
--- a/code/game/objects/items/rogueweapons/ranged/ammo.dm
+++ b/code/game/objects/items/rogueweapons/ranged/ammo.dm
@@ -97,7 +97,7 @@
 		if(T.stat != DEAD) // If theyre alive
 			skill_multiplier = 4
 
-	if(skill_multiplier && can_train_combat_skill(L, /datum/skill/combat/crossbows, SKILL_LEVEL_EXPERT))
+	if(skill_multiplier && can_train_combat_skill(L, /datum/skill/combat/crossbows, SKILL_LEVEL_LEGENDARY))
 		L.mind.add_sleep_experience(/datum/skill/combat/crossbows, L.STAINT * skill_multiplier)
 
 //arrows ฅ^•ﻌ•^ฅ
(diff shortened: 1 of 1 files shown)


NotGoldenglow:
Legendary Bows/Crossbows is not as big of an impact as Legendary melee. It doesn't affect your damage output or accuracy, only the charge time for arrows, which fun fact! 13 PER Masters or 15 PER Experts will have the lowest possible aim times regardless. Anything more is actually a detriment to you because you'll charge faster than you're allowed to fire, and thus lead to misfires/false shots.

Source: I tend to play 18 PER Poachers&Wardens often for the funny longbow nuke. It's bad.

n3tis:
> Legendary Bows/Crossbows is not as big of an impact as Legendary melee. It doesn't affect your damage output or accuracy, only the charge time for arrows, which fun fact! 13 PER Masters or 15 PER Experts will have the lowest possible aim times regardless. Anything more is actually a detriment to you because you'll charge faster than you're allowed to fire, and thus lead to misfires/false shots.
> 
> Source: I tend to play 18 PER Poachers&Wardens often for the funny longbow nuke. It's bad.

I think a better way to work around with this would be for me to make a PR giving all ranged classes Toxophilite; which is a virtue. I think it's ridiculous that wardens and poachers can only get EXPERT skill in bows; but men at arms start out with MASTER? Toxophilite is described as a virtue that a character gets from having an early interest in archery; and it allows them to train all of their ranged skills to LEGENDARY. I believe that all ranged classes; keep and antagonists alike, should get this as a **trait** instead. Why would an archer; supposedly training for a long time- be unable to gain any further skill in something that is supposedly their primary purpose? Feels a bit unpolished.

Xander3359:
If anyone wishes to make this a reality, feel free to make the PR but lock the cap behind a virtue.

PR #1045, opened 2026-03-09
Improves the piercing defense of metal armors

Pidgey-chan:
## About The Pull Request

Title, sets the piercing values of studded leather/light brigandine to 50, mail to 30, and plate/cuirass to 70. Also assigns 80 piercing defense to the dragonscale necklace.

## Testing Evidence

Value change, take my word for it

## Why It's Good For The Game

Should prevent archers from completely bypassing the defense of a padded gambeson/coif by aiming for the eyes or hands or feet. Frankly I'd rather see plate have higher arrow defense than padding, kevlar gambesons make no sense, but that would require a broader rework. Band-aid for the armors with arrow defense being avoidable by aiming around it, and should make mail under plate a bit more viable.

### DIFF
--- a/code/__DEFINES/roguetown/armor_defines.dm
+++ b/code/__DEFINES/roguetown/armor_defines.dm
@@ -93,25 +93,25 @@
 #define ARMOR_LEATHER list("blunt" = 60, "slash" = 50, "stab" = 40, "piercing" = 30, "fire" = 0, "acid" = 0)
 #define ARMOR_SPELLSINGER list("blunt" = 70, "slash" = 70, "stab" = 50, "piercing" = 40, "fire" = 0, "acid" = 0)
 #define ARMOR_LEATHER_GOOD list("blunt" = 90, "slash" = 70, "stab" = 50, "piercing" = 50, "fire" = 0, "acid" = 0)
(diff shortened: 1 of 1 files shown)


Pidgey-chan:
> This even increases the piercing resist on some leather for some reason? increases the piercing resist on chain mail???
> 
> You might as well just halve the damage of all ranged attacks, that wouldn't be too dissimilar to this result. if you really don't want to get shot, shields block the vast majority of arrows, there are options. and archers can't entirely ignore your padded gear! accuracy is capped at 50% for hand and eye shots. this is probably going to hurt crossbows really bad too, and you'll usually only get one shot with those...

Yes, mail and studded leather/light brig are metal or metal reinforced armors with high stab protection, intuitively they should be capable against arrows as a result, although the values of both are still left fairly low. This does not directly lower the damage of ranged attacks nor introduce any especially powerful anti-ranged armor, but rather allows certain smallzones to be protected which were almost unarmorable to ranged in the past. It also cuts down on the "quantum warping steel that moves out of the way of arrows and bolts" issue somewhat, though padding remains the best armor for ranged overall. Frankly, padded gambesons are still too good, I can't imagine a knight ever seriously choosing a haubergeon over the mystical kevlar linen, but at least we can cut down on armor-piercing footshots.

PR #1046, opened 2026-03-09
Adds the hounskull to the hedge knight vendor

Pidgey-chan:
## About The Pull Request

Title, adds the hounskull bascinet to the hedge knight's hordemaster at the same price as the pigface.

## Testing Evidence

image

## Why It's Good For The Game

They already get the pigface bascinet, the hounskull is a more handsome reskin. It's a rather knightly helmet. No balance implications.


Pidgey-chan:
Will undraft after testing, never messed with these lists before.

PR #1058, opened 2026-03-10
Ports Vanderlin's Explosive Collars + Adds Initial Concept of Penitent (likely to change in the future)

vitoras228322-ai:
## About The Pull Request

1)Ports various types of control collars from SR, which were in turn ported from vanderlin.
Link to original PR - https://github.com/Scarlet-Reach/Scarlet-Reach/pull/1936
I decided to not give it to bandits rn + restricted the crafting to collars alone, but not detonators (Inquisition and church get one each, with inquiistion being able to order more if necessary. Inquisitor gets a combo of detonator + collar in their room).

2)Adds in the first draft of Penitent, aka Evil Templar, aka a Heretic undergoing Penance in the Tennite Church. Essentially - a slightly different heretic wretch for NOW (intended to be changed further with separate class for each patron). Has a bomb collar on their neck from the start, gets heavily debuffed if control collar is off. A high pq role (same as inquisitor, aka 30). Has some unique examinations texts for different people.

These changes are included into one PR due to the idea of penitents in any iteration requiring control collars.
## Testing Evidence

It compiles + video of collar boom: https://dropmefiles.com/oHhfW
image
(Likely to get unique sprites for clothes in the future)

## Why It's Good For The Game

1) Collars are fun mechanically and allow for some extra intimidation. As they were initially meant to be an Inquisition-exclusive item on Vanderlin - they could use some tweaking (I am not sure if you actually can bind a specific detonator to a specific collar, for example). Still - it provides an interesting venue for inquisition members to "Redeem" heretics and keep them in check.

2) Penitent is in a ROUGH draft and was done as a part of dev comm, but I will more than likely continue working and expanding on the idea as time goes on. The idea of Astrata, as the Deity of Order, subjugating her enemies and using them for her purposes is... kind of cool, imo?  Yeah, that's pretty much it.


Skyglazer:
I don't think the "Penitent" is a good idea. Speaking mainly from an IC perspective, most of our current lore would point towards the Church and Inquisition wanting to _convert_ heretics, not allow them to use their connection to the evil gods of the setting. Especially when the inhumen pantheon is often played subtly, as a subversive force- that requires things like the Inquisition to root out, so on. So, even with the threat of a 'bomb collar,' it'd be a very tough sell to get people to ICly trust the idea of using these evil forces for good, somehow, and also not expect them to try to be getting away with whatever they can the instant nobody is looking. It's how the inhumen gods- particularly Matthios and Baotha, are.

I think with our current lore, someone in the position of that role would be far more likely to be converted or executed, not allowed to be a heretic but with a bomb collar. I don't think the idea fits in our setting.

Pidgey-chan:
Bitters, penitent.

RadiantFlash7:
Long stale and abandoned.

PR #1060, opened 2026-03-10
Ironchads can forgo jman weapon skills for expert shields

Fatherless1Xx:
## About The Pull Request

This PR gives Ironchads the option to forgo journeyman weapon skills in exchange for expert shields, since defence mastery is the core gimmick of the class.

## Testing Evidence

1 line change

## Why It's Good For The Game

It gives the Ironchad a unique niche as a role that excels at defense but not much else. Honestly that is what it probably should have been from the start.


Pidgey-chan:
I'm normally wary of the notion that adventurers should have expert combat skills at all but this does sound pretty funny

PR #1089, opened 2026-03-14
partial port of Minor Integrity Damage Review by FreeStylaLT #4780 from Azure-Peak (without the unnarmed nerf)

zeroisthebiggay:
## Original PR

https://github.com/Azure-Peak/Azure-Peak/pull/4780

## About The Pull Request

Copy-pasted description

> -Punches now appropriately deal half damage to integrity (they didn't at all before)
> ~~-Default blunt damage modifier was bumped up to 1.6~~
> -Slash etc damage dealt to integrity is not mitigated by armor.
> -Blunt damage however IS mitigated.
> This resulted in them dealing roughly* equivalent damage except the bladed weapons ALWAYS dealt the same damage to EVERY armor, unlike blunt.
> *circumstances matter a ton
> 
> This now makes Blunt better at it across the board, which is the intended way.

changes punch's integrity damage to full to offset the fact that it wasn't being applied in the first place; still fixes the bug with unarmed doing its unintended amount of integrity damage, but now unarmed will do exactly the previous amount, but in an intended way

## Testing Evidence

i had screenshots here from when this pr would affect the balance, but now it doesn't

## Why It's Good For The Game

bugfixes, ~~which inadvertantly nerfs using your bare hands to break people's helmets super fast at high strength values~~ no longer does that

~~experimental change from the PR is making blunt weapons do ever so slightly higher integrity, open to discussion~~

### DIFF
--- a/code/modules/mob/living/carbon/human/human_defense.dm
+++ b/code/modules/mob/living/carbon/human/human_defense.dm
@@ -36,7 +36,7 @@
 			playsound(loc, get_armor_sound(used.blocksound, blade_dulling), 100)
 		var/intdamage = damage
 		// Penetrative damage deals significantly less to the armor. Tentative.
-		if((damage + armor_penetration) > protection)
+		if((damage + armor_penetration) > protection && d_type != "blunt")
 			intdamage = (damage + armor_penetration) - protection
 		if(intdamfactor != 1)
 			intdamage *= intdamfactor
@@ -259,7 +259,7 @@
 			var/stealthy = (m_intent == MOVE_INTENT_SNEAK)
 			if(H.try_accept_offered_item(src, offered_item, stealthy))
 				return TRUE
-	
+
 	if(..())	//to allow surgery to return properly.
 		return
 	retaliate(user)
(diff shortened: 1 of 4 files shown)


zeroisthebiggay:
ratwood might have some form of snowflaking to make blunt weapons do extra integrity damage so i may have just created a demon by fixing this
dokibird-doki
needs testing

maker-of-nightmares:
was more meaning if you fucked with blunt weapons to buff them you might make knuckles/katar/unarmed weapons go crazy because of the different stat scaling they use and the potency they _already_ have

code/modules/mob/living/carbon/human/human_defense.dm
```diff
@@ -36,7 +36,7 @@
 			playsound(loc, get_armor_sound(used.blocksound, blade_dulling), 100)
 		var/intdamage = damage
 		// Penetrative damage deals significantly less to the armor. Tentative.
-		if((damage + armor_penetration) > protection)
+		if((damage + armor_penetration) > protection && d_type != "blunt")
```
Jackofashepard, line 39: Wouldnt this nerf all blunt weapons?

zeroisthebiggay: blunt doesn't have AP

Jackofashepard:
So i dont know if it was this PR or another but suddenly blunt does 0 integrity damage to plate if you wear a gambeson beneath. the damage ghosts through the plate doing 0 and all applies to the gambeson.

zeroisthebiggay:
> So i dont know if it was this PR or another but suddenly blunt does 0 integrity damage to plate if you wear a gambeson beneath. the damage ghosts through the plate doing 0 and all applies to the gambeson.

not introduced by this or any recent pr
this is how it's been for a while

PR #1101, opened 2026-03-15
Swap INT values for heavy knight and royal champion

Actuallygoinginsanern:
## About The Pull Request

This is literally just a simple numberswap. it swaps the INT values between royal champion and heavy knight subclass

## Testing Evidence

image
It compiles! (I don't know why it wouldn't)

## Why It's Good For The Game

Right, in simple terms, heavy knight is already the most evidently powerful class of knight compared to other subclasses. It's given immediate heavy armor, buffs to it's str and con. it's only downside is -1 SPD, which will rarely make a difference in a fight.

Royal champion, on the other hand, is intended to be more of a duelist subclass of knight. Which INT plays a much heavier role, it doesn't make sense the royal champion, allegedly meant to be the best with the blade, can be out-feinted by your average heavy knight player. This just swaps the INT values.

### DIFF
--- a/code/modules/jobs/job_types/roguetown/garrison/knight/knight_champion.dm
+++ b/code/modules/jobs/job_types/roguetown/garrison/knight/knight_champion.dm
@@ -9,7 +9,7 @@
 	category_tags = list(CTAG_ROYALGUARD)
 	subclass_stats = list(
 		STATKEY_STR = 1,
-		STATKEY_INT = 1,
+		STATKEY_INT = 3,
 		STATKEY_WIL = 2,
 		STATKEY_SPD = 2,
 	)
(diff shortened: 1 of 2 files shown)


hugmeorelse:
Honestly, this makes a ton of sense. I played heavy knight the other day, and I was surprised by how amazingly good the stats were. That build is simply the meta to go with, so it's good to see other sub-classes get some love too with a swap. This also avoids stat inflation!

My only concern is if the +3 stat differential that champion has over the heavy knight is too much? It's hard to say, considering it's hard to qualify whether or not heavy armor is worth a -3 stat difference.

Pidgey-chan:
Good idea, but heavy knight has a weighted 7 after this. I'd stick the lost int into con and wil.

Pidgey-chan:
What why did you leave champion with 11 points LOL

PR #1102, opened 2026-03-16
Dukes and Merchants spawn with a single pistol

Jamdrawers:
Also fixes dukes spawning with two daggers!

## About The Pull Request

Puts a pistol (along with bullets and powder (and possibly a satchel to store them in)) in the inventories of the Grand Duke's Merchant and Inbred subclass (but not the warrior or magocrat subclasses) and the Merchant, along with some firearm skill for the merchant-duke and merchant (but not the inbred duke, they don't actually know how to use that thing)

Also comes with a tiny tweak of fixing the pistol's pixel offset, and making them repaired with engineering instead of blacksmithing
Also fixes the way that dukes always would spawn with two knives.

## Testing Evidence

image
image

## Why It's Good For The Game

Guns are cool and fun and exciting! And we've already gone through all the effort of porting them in, it's just a matter of implementing. The Merchant and Duke are important roles and it's nice to have a suitably flavourful form of self-defence for them.

### DIFF
--- a/code/game/objects/items/rogueweapons/ranged/firearms/arquebus.dm
+++ b/code/game/objects/items/rogueweapons/ranged/firearms/arquebus.dm
@@ -29,6 +29,8 @@
 	grid_height = 32
 	grid_width = 96
 	experimental_onhip = TRUE
+	pixel_x = 0
+	pixel_y = 0
 
 /obj/item/gun/ballistic/firearm/arquebus_pistol/getonmobprop(tag)
 	. = ..()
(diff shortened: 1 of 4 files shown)


Jamdrawers:
> Personally I'd like this to be a hard to craft artificer item than some token weapon to already strong roles.

I think it'd be cool as a high-level engineering recipe *as well* later on but I think merchant and those two duke subroles (the ones that don't get magic spells or heavy armour+expert swords) are good to start with. They're the classics!

Pidgey-chan:
There shouldn't be any obtainable guns until they have an implementation more interesting than being supercrossbows.

Fatherless1Xx:
> > The duke really doesn’t need to be made any safer than he already is. As it stands he’s basically unkillable.
> 
> bandithead downthumbs the literal coolest thing ever because it'd make the keep 1% stronger
> 
> Doesn't even make the duke more durable, doesn't even buff the most combat capable duke, just adds a cool gun to the one duke class which can't wear heavy armor or use magic.

You cannot have a true one shot kill weapon on a server with escalation rules. Those two things do not fit together.

On RT2 the pistol had built in counterplay. It took time to draw. It took time to aim. That delay mattered, because someone could just stab you to death while you were not expecting it.

Here the rules expect you to show intent first and give the other person a chance to react. If that is the standard then a weapon that instantly ends the fight the moment it is used just breaks the whole system. Escalation only works when there is actually time to escalate.

PR #1111, opened 2026-03-17
Renames 'Slurbow' to 'Handcrossbow'

Jamdrawers:
## About The Pull Request
As per title.

## Testing Evidence

image

## Why It's Good For The Game

While I do understand that 'Slurbow' IS a real term and that there IS a wikipedia page for it, the fact is that *nobody would know that*. Slurbow is a weird name that causes a lot of confusion, while Hand Crossbow is both more familiar and clearly communicates what it is without unintentionally suggesting anything false.


Pidgey-chan:
...Does it work one handed? Because when I hear hand crossbow I think the one from D&D.

Xander3359:
We have a crossbow already, slurbow is distinct and anyone can use google.

PR #1122, opened 2026-03-18
Unique keep clothes for rockhill

Jamdrawers:
## About The Pull Request

Reimplements some lost clothing items! Finally now that I figured out what was causing the missing sleeves, they look as sexy as they ever have.

When the map is Rockhill, the Duke, Hand, Steward, Councillor and Court Magician will have the cool swanky outfit from Ratwood 1. The exception is the prince/heir - their relatively unpopular pantsuits are instead worn as shirts underneath the other clothing that they start with.
The outfits are essentially optional in that the relevant roles will have spare copies of the alternate outfit in their luxurious wardrobes in the keep- they are all nobility after all. Given the vast riches of the wearer, most of them are quilted so as to have armour similar to leather or padding (other than the Hand's Jacket, which has armour similar to studded leather!)

Given that the RW1 duchess outfit is, while I think very cool, a bit more *niche* of a style than the ever-loved frilly dress, I might give like a little 'choose your armour' widget at roundstart to just pick which one your duchess wants to wear.

Also, as a side effect, this fixes a graphical issue with the Citywatch Armour sprite not overlaying on arms properly - though as a side effect, this makes it cover arms as well as chest mechanically armor-coverage-wise. To compensate, the integrity is lowered to less-than-plate-tier-but-more-than-cuirass-tier

## Testing Evidence
image

Duke

image

Duchess (hood optional)

image

Court Magos.

image

Prince - worn underneath the main outfit with all these cases

image

A Princess - suit worn under the dress by default

image

Another princess - the Nettlesome Scamp

image

Hand. So classy!

image

Steward. Lady Steward begins with the dress worn underneath it which I think looks nice

image

Counsellor

## Why It's Good For The Game

I think they look really cool and characterful. Would be a way to distinguish between maps visually, similar to how the Sultan and Sultana of Desert Map have different outfits.

Thanks to Infraredbaron for spriting them way-back-when.

🆑
add: Brings back old sprites from RW1 for nobility clothing, which nobility spawns in when on Rockhill
change: Renames 'Crown of Rotwood Vale' to 'Crown of the Realm'
change: Rebalances City Watch armour slightly, reducing integrity, adding arms coverage, and fixing a graphical glitch
/🆑


Pidgey-chan:
I like this implementation a bit better since there are a couple cases of roles getting these and their original clothes both. That being said, the iconic lord's cloak is much better than this 19th century admiral's uniform.

Fatherless1Xx:
These sprites were originally added to Ratwood Keep under a specific licence. Did you obtain permission from the licence holders before porting them?

https://github.com/Rotwood-Vale/Ratwood-Keep/blob/74f55d70294af6ac3650d7e2839ef70b3ce7db0d/icons/roguetown/clothing/licensed-infraredbaron/LICENSE.md?plain=1#L4

PR #1125, opened 2026-03-18
Update thug.dm

TheRealRatMan:
Changing stat nuke from WP to STR

## About The Pull Request

Miscreant is meant to be sneaky and fast, but you lose your breath the moment you try to run from danger like you should. After seeing stat balance of the other subclasses, I just removed the WP nuke while it still being in line with the rest of the thug classes.

## Testing Evidence

Its a 1 line change

## Why It's Good For The Game

Makes thug less AIDS to play, but still not power creeping in on adventurers and whatnot


eeshaa1:
woohoo

PR #1138, opened 2026-03-19
Killers ice rework

RadiantFlash7:
## About The Pull Request
This PR shifts killers ice from being a one tap poison tool, to providing a massive con penalty. Since it's no longer a deadly poison, poisoner roles get strong poison to use for damage. 

## Why It's Good For The Game
Killers ice is utter ass to deal with. Anyone with it can onetap a person and there is quite litterally nothing they can do. People can point to having antidote, but that both requires the antidote, and time to drink it before collapsing. Of which, you do not get. You drop from killers ice within about 3 seconds of getting it injected.

The changes provided still make killers ice incredibly potent- a 5 minute con malus that makes fighting a VERY poor idea. You're likely to bleed out rapidly, or hit paincrit if you take much damage while -10 con. Yet, the benefit of it is that it has counter play. You're not instantly in crit. 

I set up this PR about 2-3 months back, but never pushed it out of concerns people would call it a salt PR. It's something that still needs to be addressed.


RadiantFlash7:
> Like 3 people get tiny slivers of the substance. Maybe suitors too? The whole point is you can put it in someones drink and they'll die from it as a quick way to remove other people. By making it a debuff instead you still have to actively kill someone with a weapon, kind of ruining the point of it being a poison? Doom poison isn't viable either as you cough up your organs, meaning that assumedly whoever finds the body will know it was poison still. 
> 
> 
> 
> It's called **Killer's** Ice. Why wouldn't it kill?

Killers ice is not being used in drinks however. Its being used in daggers to instantly apply it and kill a target in 3 seconds. No setup, no intrigue, not counter play, just pick intent stab, done.

CVSthePharmacy:
The only dagger it can be applied to is the plaguebringer dagger. 
I do not believe killers ice should be removed, as this has never been an issue in the two years we've had it for. 
I agree that a con poison would be very cool, albeit as a separate poison. 
This poison should not replace killers ice, it has its place and is sitting nicely in it. 

Do not remove killers ice. This is very bad.

Add a separate con poison if you'd like. But killers ice as is, is unbrewable and you only get 1 drop.

Pidgey-chan:
I like it. Poison is disproportionately lethal in this game. Getting stabbed in the chest or your neck cut isn't usually immediately lethal, even naked, why is this one poison? Esp when poison daggers exist.

YakumoChen:
just blacklist the dagger from killersice

PR #1156, opened 2026-03-21
mending 'nerf' from azure peak port

zeroisthebiggay:
## About The Pull Request

https://github.com/Azure-Peak/Azure-Peak/pull/4761

> 4 Seconds of stand-still to repair.
> If interrupted, you get the cast back.

## Testing Evidence

image

## Why It's Good For The Game

kind of a saltjak but being able to mend twenty percent of fullplate midfight fucking rots

### DIFF
--- a/code/modules/spells/spell_types/wizard/utility/mending.dm
+++ b/code/modules/spells/spell_types/wizard/utility/mending.dm
@@ -41,6 +41,15 @@
 		revert_cast()
 		return
 
+	user.visible_message(
+			span_warning("[user] begins to concentrate on [I]!"),
+			span_notice("I begin to concentrate on [I]..")
+	)
+	if(!do_after(user, 4 SECONDS, TRUE, I, TRUE))
+		to_chat(user, span_warning("My concentration breaks! I could not repair [I]."))
+		revert_cast()
+		return
+
 	var/repair_amount = (repair_percent + (user.STAINT * 0.01)) * I.max_integrity
 
 	I.obj_integrity = min(I.obj_integrity + repair_amount, I.max_integrity)


Pidgey-chan:
This is practically a bugfix

NoShamNoWow:
> Anyone disagreeing with this pr clearly never actually fought a hard battle in their life.

Starting your argument with "If you disagree you bad at the game" is such a a leap of logic and assumptions.

OP specifically stated how strong it is in DUELS in their first reply. When you put things into a group fights it honestly doesn't matter due to too many factors to count. Playing Icono and saying you are a raid boss is not because of just mend. That class is one of the strongest bandit classes due to 'I grab you and punch your head in' that unarmed lets you do with high STR and I guarantee that you would get your shit rocked if you fought 2 rivalling keep roles alone focusing on weakpoints or the same bodypart. That class itself is NOT FUN to fight alone for tons of reasons, as all grappler/unarmed classes are.

It's not 'broken to shit' it's a very good utility spell for mage-orientated classes. Putting a 4-second timer just to repair something like a broken piece of clothing will take even longer because there is not a 'flat boost' repair. It's percentage based, they all repair at the same speeds regardless of what it is. And you know what armor-types have the most? Always chest. The chest armors. A area of the body that is the LEAST effective in attacking compared to head/neck/limbs. The only real problematic ones are heretical armors and the strongest plate-sets you can find/buy in game, and the former are rare as it is.

After testing enough myself I would disagree with a longer cast and more strongly suggest higher stamina use so it cannot be spammed in combat. A spell shouldn't be made tedious for everyone regardless of situation because it salted OP enough for them to scour azure peak for whatever nerf they did over there and bring it here.

Pidgey-chan:
Even smiths can't repair shit in combat lol get out of here. It's bad enough that mages can just invalidate skills and squire repair, combat repairs are egregiously stupid atop that. Practically a bugfix, speedmerge.

Jamdrawers:
It seems pretty plainly obvious that you shouldn't be able to mend your own armour during combat.

That said, it'd probably be fine to compensate by making it work a bit quicker - it doesn't have to be that much *worse*, it just should not be a mid-combat strategy.

CaydenT:
More importantly does this fix the blue light bug where you are sometimes permanently emitting the blue ambient light from casting it?

NoShamNoWow:
Please make this quicker its so god awful now.

PR #1161, opened 2026-03-22
Ports over Triumphs at Night from AP

Teragenic:
## About The Pull Request
Ports: https://github.com/Azure-Peak/Azure-Peak/pull/6199

Reasons:
Pros:
Literally roleplayers are not forced to go and sleep to gain triumphs, cutting roleplay short.

Cons:
Literally none. Triumphs are used for RP reason usually.

Reasons to still sleep.
Sleep is STILL the only way to have good active stamina regen and minor healing for most people.

Its the ONLY way to gain skill above apprentice.
## Testing Evidence

None. It genuinely shouldn't break anything or else AP wouldn't have full merged it. I did not see any bugfixes for it. 

## Why It's Good For The Game

Roleplay is good. Let people not have to interrupt chatting, sex, etc by needing to sleep for arbitrary roleplay currency.

### DIFF
--- a/code/controllers/subsystem/nightshift.dm
+++ b/code/controllers/subsystem/nightshift.dm
@@ -96,12 +96,11 @@ SUBSYSTEM_DEF(nightshift)
 				apply_status_effect(/datum/status_effect/debuff/vamp_dreams)
 			if(HAS_TRAIT(src, TRAIT_NIGHT_OWL) && !HAS_TRAIT(src, TRAIT_NOSLEEP))
 				apply_status_effect(/datum/status_effect/debuff/sleepytime)
-			if(HAS_TRAIT(src, TRAIT_INFINITE_STAMINA) || HAS_TRAIT(src, TRAIT_NOSLEEP))
-				handle_sleep_triumphs()
 		if("dusk")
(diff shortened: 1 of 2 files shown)


Teragenic:
> You're interrupting RP for... _triumphs?_
> 
> just feels like this serves to trivialize the things the triumph buy system is meant to make exclusive and rare like high-value virtues, combined trait virtues and taking "no flaw" round after round

Roles like Duke and Priest barely get time to sleep due to being incentivised to roleplay, so **Yes.**

Also, I don't think it will trivialize things more than they already are. If people are already taking it, people will already take it round after round. This just makes people not have to think about interrupting RP with having to sleep. AP has already shown as a good testbed. They have a BETTER Triumphbuy system with Virtues and more see here: 
image

They HAVE combined trait virtues and MORE things that cost more triumphs! That is my statistics I am putting forward to say that this change will probably not affect people who use those virtues already, BUT still allows people to roleplay.

The-Sun-In-Splendour:
image

Got a triumph for surviving the day as a rotting corpse in an unmarked grave.

zoktiik:
> Got a triumph for surviving the day as a rotting corpse in an unmarked grave.

And it got fullmerged in this state?

PR #1170, opened 2026-03-23
Even MORE mperature bug fixes (DO NOT FULLMERGE IMMIDIATELY)

john-pawson:
## About The Pull Request

Hey! It's me, again. Last PR got fullmerged pretty fast (Thank you for that Guidesa, not being sarcastic here, it helped) so I'm here with another PR with even more bugfixes! Two to be exact, one that works and one that MIGHT not:

- Shivering emote now check for invisibility, no more shivering giving you away while invisible.

THIS ONE NEEDS SOME TIME TM'D TO SEE THAT IT WORKS PROPERLY!! It'd be a great favor if any of the admins on a round that this is TM'd could put in a rain_storm weather and see if any tornados spawn, but:

- Fixes tornados not spawning. Previously, the tornado code was behind a check for the lightning cooldown, which meant that tornadoes only had the CHANCE to spawn every 6 seconds, instead of checking every tick as it was supposed to.

I wasn't able to test this on local since... for some reason forcing weather on the test map isn't working properly (This happened on all other bugfixes, these were merged and the weather still works perfectly on the server, so I'm guessing it's something to do with localhosts) so! DO NOT FULLMERGE THIS UNTIL YOU'RE SURE TORNADOES WORK AND RAIN_STORM WORKS PROPERLY ON THE SERVER!!
Thanks for your attention.

Edit 1: Everything seems to be working properly! Also, new thing: Looking up now gives you the warning message of incoming weather, if you were unlucky enough to be inside when the first message triggered, you can now litterally just go outside and look at how it's going to be. Hopefull this'll mean a bunch less "surprise snowstorms"

Testing evidence is below.

## Testing Evidence

image

Only shivers while visible!

Edit 1: 
image

## Why It's Good For The Game

buuuuuuuuuuuuuuuuuuuuuuuug fix.

### DIFF
--- a/code/datums/particle_weathers/datum_types/rain.dm
+++ b/code/datums/particle_weathers/datum_types/rain.dm
@@ -66,10 +66,9 @@
 
 	COOLDOWN_DECLARE(thunder)
 
-/datum/particle_weather/rain_storm/tick()
+/datum/particle_weather/rain_storm/proc/lightning_check()
 	if(!COOLDOWN_FINISHED(src, thunder))
 		return
-
 
 	var/lightning_strikes = 6
 	for(var/i = 1 to lightning_strikes)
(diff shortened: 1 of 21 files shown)


Pidgey-chan:
Bandaid that doesn't address the core issue of the entire system not being fun to begin with.

PR #1182, opened 2026-03-25
Court chaplain

Jamdrawers:
## About The Pull Request
Made for @goblins4lyfe on discord, figured it wouldn't be too difficult to put one up for him on his behalf.

A new Courtier role, whose job it is to advise the duke on spiritual matters and do diplomacy with the church, like a religious-themed councillor

Currently has stats similar to an adventurer missionary - we want them advising the duke, not spending all round grinding skills. They're also disallowed from Xylix (that's the Jester's job!) or Pestra (that's the physician's job!).

Considering to bar them from revival spells - they are meant very much to be advisors and diplomats, not Mercy from overwatch, and shouldn't overshadow the church or physician.

WIP, feel free to chime in.

## Testing Evidence

image
image

## Why It's Good For The Game

Could be a fun role to have around. Roleplay opportunities!

:cl:
add: Adds chaplain role!
/:cl:


Jamdrawers:
> And yeah, Church very rarely (in comparison to other 'factions') has the occasion to RP with others, most people visit Church as a corpse, and it is, generally pretty hard to do fun RP with a corpse.

I think the big thing with that is just all the acolytes have the extremely powerful ability to resurrect dead people, and with that extreme power comes the extreme responsibility and extreme *expectation* of using it.

It's why I wouldn't want the Chaplain to be able to revive people because it'd free them up from the responsibility of reviving people and let them instead spend their time doing different things.

Guidesu:
The Court Chaplain should not have any miracle tiers. At least T0 or at very max T1. They are middle man of church and keep and should not even be missionary-like level. Aside of that everything looks fine. @Jamdrawers

Dagonlives:
I don't like this P.R much unless it makes it clear that the chaplain is a holy see representative, not in the pocket of the keep.

The chaplain should be objectively loyal to the church not the keep. Otherwise it's an even greater weakening of an already hefty power disparity. 

Also the Prelate can excommunicate them and remove their powers.  Which any sane one will do the moment they speak against the prelate.

ChoirOfHeaven:
I played this role and I have to say it was very well received by both the Grand Duchess player and the Prelate player. The only thing I can think of.

I do think that the Chaplain should be primarily loyal to the Keep as its a Courtier role. Realistically your Court Chaplain would be someone picked for their bias _towards_ the royal family not the bias towards the Bishop. Obviously they're still a person of deep devotion so that accounts for adherence to the Prelate's (sane) commands.

Jamdrawers:
> The chaplain should be objectively loyal to the church not the keep. Otherwise it's an even greater weakening of an already hefty power disparity.

I think it's kinda important for intrigue for the player to have some wiggle-room with where exactly their loyalties are between the duke, the local church and the big setting church. Maybe some members of the church suspect you're in the pocket of the keep, maybe some members of the keep think you're in the pocket of the church, maybe you can persuade them of this, rumours, suspicion, arguments, trials, etc etc.

Dagonlives:
> > The chaplain should be objectively loyal to the church not the keep. Otherwise it's an even greater weakening of an already hefty power disparity.
> 
> I think it's kinda important for intrigue for the player to have some wiggle-room with where exactly their loyalties are between the duke, the local church and the big setting church. Maybe some members of the church suspect you're in the pocket of the keep, maybe some members of the keep think you're in the pocket of the church, maybe you can persuade them of this, rumours, suspicion, arguments, trials, etc etc.

The issue however is the church is already considered weak and irrelevant compared to the keep.  Adding a courtier that can effectively replace them with anastasis is a problem.  If a chaplain decides to betray the church they will almost certainly succeed.  The odds are that stacked. 

I get we want nuance but this is a real concern I have with the role.

ChoirOfHeaven:
I dislike that Court Chaplain has lost access to Acolyte level miracles and is now essentially a Churchling. I understand that there was a dislike of the miracle access already mentioned by a shotcaller in the server further up. Having played Court Chaplain though the nerf goes further than just mechanical.

If the Court Chaplain is supposed to be a liaison between the Church and the Court only, then the change makes sense, but they shouldn't be called a Chaplain because they're not even favored enough by the gods to access disciple miracles. Call them something else.

It is routinely reinforced within the setting that higher you rank within the faith the more miracles you have access to. At the current level the _Court Chaplain_ is a Churchling.

1. Bishop
2. Martyr
3. Templar | Templar Adventurers
4. Acolyte | Acolyte Adventurers <-- We were here.
5. Churchling **<-- We are here.**

Solely on mechanical level to show how odd of a call this was is that the Court Chaplain [Necra] cannot even perform a Burial Rite?

I was lucky enough to play Court Chaplain as it stood before the changes and after. This was easily going to become my new main role. But not with these current nerfs in place. They're overly penalizing.

CaydenT:
Uhh that's wrong though. Acolyte has full casting, Templar does not.

Jamdrawers:
> Adding a courtier that can effectively replace them with anastasis is a problem.

Court Chaplain doesn't have Anastasis

Lyroy:
Having tried it yesterday, I think this has overcorrected way too much towards the "just a guy wearing a robe" side of things. Being easily overshadowed by some missionary court agent who gets 2 more tiers of miracles than you do _and_ can actually use rituals does not feel good for someone who, supposedly, should be the spiritual advisor to the whole of the keep.

T2 (to avoid the resurrection spells) and Ritualist, sort of like a Templar, would be the best balance imo.

LittleEvil9:
I love this PR, having a court chaplain around made some rounds pretty fun.

ChuChu42:
chaplain's chamber door is written with all lower casing letters 
image
compared to physician's office
image
or knight captain's quarters
image

PR #1220, opened 2026-03-29
Increases Rockhill MAA and vanguard slots by 2 each.

anconfuzedrock:
## About The Pull Request
Just gives vanguard and man at arms 2 more slots on rockhill.

## Testing Evidence
Do people even do this anymore? I wish, but mine is just like three lines.

## Why It's Good For The Game
Having the garrison split between 3 smaller roles seems to have put a dampener on camaraderie and coordination. Man at arms feels especially hit by this, since a lot of its design is for 8 people- hell, there's four different subclasses, for just four people now.  The replacement watch and vanguard roles range from debatably equal to definitely weaker, six slots isn't going to break anything. Or maybe it will, but if we can test +9 statpoints for wretches, we can test this.

Vanguard are the coughing babies of the garrison so I don't think I even need to justify +2 slots for them. They're arguably weaker than adventurers and numbers are the only thing they have going for them, maybe some slightly higher numbers will make that more fun to play as.

## Changelog

:cl:
balance: Man at arms and Vanguard both now have 6 slots on rockhill, up from 4.
/:cl:

### DIFF
--- a/_maps/map_files/rockhill/map_adjustment_rockhill.dm
+++ b/_maps/map_files/rockhill/map_adjustment_rockhill.dm
@@ -11,7 +11,7 @@
 		/datum/job/roguetown/sergeant,
 		)
 	slot_adjust = list(
-		/datum/job/roguetown/manorguard = 4,//split with watchmen
+		/datum/job/roguetown/manorguard = 6,//split with watchmen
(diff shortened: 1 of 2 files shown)


Kroto-Ker:
Probably not.

PR #1225, opened 2026-03-30
vampires no longer dust on death

Xander3359:
## About The Pull Request
Makes it so vamps no longer dust when they die.

## Why It's Good For The Game
I think it's dumb

## Changelog

:cl:
balance: Vampires/Lickers no longer dust when they die.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/death.dm
+++ b/code/modules/mob/living/carbon/human/death.dm
@@ -40,14 +40,8 @@
 		SSdroning.kill_rain(client)
 
 	if(mind)
-		if(!gibbed)
-			var/datum/antagonist/vampire/VD = mind.has_antag_datum(/datum/antagonist/vampire)
-			if(VD)
-				dust(just_ash=TRUE,drop_items=TRUE)
-				return
-
 		var/datum/antagonist/lich/L = mind.has_antag_datum(/datum/antagonist/lich)
-		if (L && !L.out_of_lives)
+		if(!L?.out_of_lives)
 			if(L.consume_phylactery())
 				visible_message(span_warning("[src]'s body begins to shake violently, as eldritch forces begin to whisk them away!"))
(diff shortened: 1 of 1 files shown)


hugmeorelse:
> Gonna make it where you can use a stake on a vamp to dust them, drafting this for now.

Sounds good to me. This sounds cooler to me if I'm honest.

TamedAchilles:
> Do vampires even still have critical weakness? Like, Licker wretches are insanely strong right now. Heavy armor, uncapped strength, infinite energy. IMO, fighting to the death and not surrendering as a super strong antag slot meaning you get ashed is- for the best?

Just silver and sunlight as far as I'm aware honestly.

To contribute to the overall -1 I gave - 

I have to push back on the original framing in the OP. The idea that lickers are forced to fight to the death because of their tools or capture mechanics just straight up ignores a core part of vampyre lore: self-preservation doesn’t always mean mindless violence. A licker who actually values their unlife knows when to cut losses, slip away, or even surrender strategically since there are fates worse than the Inquisition, and most Vampire Lords would be teaching subtlety over suicidal last stands.

The original post treats every encounter in this binary fashion, but that’s ultimately the choice of the player, not a lore requirement. Overt heretic traits and ash death are known risks that smart vampyres account for. And the DNR bug, while annoying, isn’t some systemic conspiracy forcing accidental RR since it’s a technical glitch being worked on.

Blaming the class mechanics for what ultimately comes down to encounter design and player stubbornness just shifts responsibility. If every fight ends in death, maybe the issue isn’t the tools, it’s how we’re choosing to use them. And do we need to start hugboxxing our Antagonist roles because people choose to put themselves in situations where they know their character is going to die permanently? No, I don't think so and if we do it devalues Vampires a lot in this regard.

wow50000:
make it so vampires can have sex

PR #1232, opened 2026-03-30
Sweep of Towner Tweaks

Jamdrawers:
## About The Pull Request

Maybe not all of them can get in, but the ones that came to mind...
-Builder towner role built from the old 'Carpenter' role - expert carpentry+masonry, jman/apprentice mining/lumberjacking. Kind of a local town architect - it's something that got overwritten in the Parity update. Sometimes one might wanna be a dedicated Builder without taking up a valuable Architect Slot.
-Puts farmers and soilsons a bit closer to parity in terms of stats and skills - the latter got kinda powercrept while the other was left behind.
-Applies a *generous* pop cap on witches of 5. That's still a lot of witches - way too many, if you ask me, but we probably don't need eight witches knockin' about.
-Tones back some of the buffs that were applied to the refugee Harvester that somebody snuck in without anybody noticing (it's supposed to be a peasant iterant farmer, why's it got Alchemy skill?)
-Lots of buffs for Minstrel with Jman sneaking and stealing and such, making them a bit of a discount town thief. Also gives them a choice of instrument. Reduces the Music skill of adventurer bards so they're not better at it than minstrels (not that it REALLY matters much)
-Towners get Legendary in their signature skill (and sometimes an additional skill) when they're specifically OLD. OLD farmers get legendary farming, OLD surgeons get legendary medicine, OLD witches get legendary alchemy, OLD fishers get legendary fishing, etc. Tell me if any of these skills are super-mega abuseable in a way that even the 1% of Old characters having access to them would break everything, though. A fun way to make Age matter (and relatively unobtrusive - legendary isn't THAT big a deal for most of these, and Old characters are always gonna be rare.
-Edit - Towners ALSO get Master in their signature skill when they're specifically MIDDLE AGED.

Was reminded of it by https://github.com/Rotwood-Vale/Ratwood-2.0/pull/1222

## Testing Evidence

Simple stuff but I can double check

## Why It's Good For The Game
Described above

## Changelog
:cl:
balance: gives towners scaling skills (lumberjack for woodcutter, mining for miner, farming for farmer, etc) based on age (master at middleaged, legend at Old)
Caps witches at 5
Buffs towner farmer
Tones down refugee Harvester
buffs Minstrel with rogue-ish skills

/:cl:


code/modules/jobs/job_types/roguetown/adventurer/types/pilgrim/builder.dm
```diff
@@ -0,0 +1,77 @@
+/datum/advclass/builder
+	name = "Builder"
+	tutorial = "A skilled carpenter and mason, able to manipulate wood and stone to suit their needs \
+	building forts and stores, carpenting floors, putting up crosses. You can do it all with enough logs or rocks"
+	allowed_sexes = list(MALE, FEMALE)
+	allowed_races = RACES_ALL_KINDS
+	outfit = /datum/outfit/job/roguetown/adventurer/builder
+	subclass_social_rank = SOCIAL_RANK_PEASANT
+	traits_applied = list(TRAIT_HOMESTEAD_EXPERT)
```
annikaRU: I'd suggest adding TRAIT_MASTER_CARPENTER and TRAIT_MASTER_MASON, reason being this allows you to get two stone blocks from a single stone, and likewise just more planks, without it can be very tedius,

PR #1254, opened 2026-04-02
Lawless Vice from SR - roundstart bounties on almost every class

zeroisthebiggay:
## About The Pull Request

Originaly PR: https://github.com/Scarlet-Reach/Scarlet-Reach/pull/1182

> This PR adds a new vice titled Lawless, which will give you a customizable Wretch/Wanted vagabond bounty on most classes, unless you already have a roundstart bounty of some kind. If you have one already, you'll receive a randomized flaw instead.
> Conceivably, someone could pick something stupid like Grand Duke with this and have a bounty while being on the throne. Do we want to allow that? I dunno.

adds a vice that gives you a roundstart bounty, that's it
there's currently no restrictions in place as to what isn't allowed to take it, but I don't think that'll cause problems; it'll be funny if anything

## Testing Evidence

image

## Why It's Good For The Game

roleplaying minor criminals fun

## Changelog

:cl:
add: lawless vice from SR
/:cl:


zeroisthebiggay:
> I really really definitely like this but I think we genuinely will need to restrict it from certain round-integral pivotal roles. Otherwise a fucking kino feature.

i doubt most players will willingly take lawless on roles like duke
but i'm kinda curious what might come from this actually happening
lawless duke might be really kind of funny

PR #1281, opened 2026-04-07
Lets try it; Adds DNR to Prince/ess

AllCorruptingWyrm:
## About The Pull Request

As it stands; the previous PR that was just merged today was merged. After unknowingly realizing that the PR was merged, I immediately realized that the PR did not achieve much, if anything, indicating that people are still too safe. In order to make things less safe, I've decided to add **TRAIT_DNR** to Prince/ess to encourage them to be viable targets for antagonists.

I understand this might be jarring but bandaids need to be pulled off--people need to stop feeling so safe which is why I've elected to do this.

Personally, I think we should go further and KC/Knight roles should get this (while maa and co dont) but i'll be crucified.
anyway behave or i'll mark comments as resolved.

## Testing Evidence

it compiles

## Why It's Good For The Game

See above.

## Changelog

:cl:
add: DNR to Prince
/:cl:

### DIFF
--- a/code/modules/jobs/job_types/roguetown/nobility/prince.dm
+++ b/code/modules/jobs/job_types/roguetown/nobility/prince.dm
@@ -22,7 +22,7 @@
 	round_contrib_points = 3
 	social_rank = SOCIAL_RANK_ROYAL
 	cmode_music = 'sound/music/combat_noble.ogg'
-	job_traits = list(TRAIT_NOBLE)
+	job_traits = list(TRAIT_NOBLE, TRAIT_DNR)
 	job_subclasses = list(
 		/datum/advclass/heir/daring,
 		/datum/advclass/heir/bookworm,


Pidgey-chan:
Maybe if bandits didn't have a spell to explode princes instantly.

Pidgey-chan:
> > Maybe if bandits didn't have a spell to explode princes instantly.
> 
> They have like 15+ different roles explicitly meant to protect them from this happening. If you get got by a bandit, it's 99% because you put yourself in a situation where it'd happen.

I could walk into the keep as a bountyless bandit or adv Matthios holy caster and middle click the duke or either prince while standing in line w/ my face exposed to give them like 20 firestacks or gib them if they have over 1k in their account

Uhhhhhh so true though maybe they'll misclick on the knight 3 tiles away though

Jamdrawers:
Maybe instead of being tied to the role, it would kick in after you are declared regent. The Throne owns your soul.

Pidgey-chan:
> > > > Maybe if bandits didn't have a spell to explode princes instantly.
> > > 
> > > 
> > > They have like 15+ different roles explicitly meant to protect them from this happening. If you get got by a bandit, it's 99% because you put yourself in a situation where it'd happen.
> > 
> > 
> > I could walk into the keep as a bountyless bandit or adv Matthios holy caster and middle click the duke or either prince while standing in line w/ my face exposed to give them like 20 firestacks or gib them if they have over 1k in their account
> > Uhhhhhh so true though maybe they'll misclick on the knight 3 tiles away though
> 
> If you have more then 5 con you'll survive this. Also this is just like against the rules baseline? Make a ticket. Making hypotheticals that go against the rules of the game doesn't make the argument valid.

What's against the rules, killing people? RR? You don't really have a choice for the latter. At 1000 (900 bc noble trait gives you +100 for the spell's sake) in the bank it is literally just an instant kill that removes 1-4 limbs w/ no checks and 500 is _close_ since there's still an explosion

UntoldTactics:
> > Maybe instead of being tied to the role, it would kick in after you are declared regent. The Throne owns your soul.
> 
> We do already have it on coronation. Maybe.
> 
> I don't think we should merge this PR. We should be waiting like, weeks to a month to continue creeping DNR on more Keep rolws.

I think the very fact that people feel the need to creep this in, against community outlash and criticism, not addressing it, but simply saying you need to...what? Sneak it in? It's concerning and shows that a certain portion of the playerbase wants to override the other entirely, and shove in what they feel is better for the server.

I don't see how in any way this is an acceptable way to handle the matter and I genuinely believe that DNR as a whole is a wash. The players aren't ready and I doubt they ever will be. It's not just about keep or town players, but also about how antagonists will treat those roles who are DNR. I believe most antagonists would simply kill them and laugh at the chaos they cause, all the while being dull and uninteresting, because it is pathetically easy to off someone in the manor when its so pitifully easy to break into.

Flyguy715:
This shit sucks, dont push it.

Keep DNR to duke, leave the other keep roles alone.

DogExtrordinaire:
deeply unneeded. If you want a prince to stay dead you can just round remove them yourself. No one actually cares about princes and this won't make them care more.

PR #1295, opened 2026-04-08
Moves the bandit camp and wretch fort into an 'Other Z', gives them extra z-levels

Jamdrawers:
## About The Pull Request

Moves the bandit camp and wretch fort into an 'Other Z', gives them extra z-levels. They're now stored on a separate map instead of all sharing the same 'centcomm, activated when dunworld or rockhill are activated. Also expands them a little to make use of extra z-levels.

Also gives either prison a minable area so you can force prisoners to mine for you.

Will probably need to wait for the gnoll thing to merge before since the  

## Testing Evidence
image
image
image

## Why It's Good For The Game

1) lets them make use of Z-levels and stairs! Also a some back-end stuff that'll be relevant later on down the line.
2) Few fun additions to make it more of an interesting place.

## Changelog

:cl:
map: moves wretch base and bandit camp to a separate Z
/:cl:


Jamdrawers:
Alright turns out the problem was that the map.json is read before the game restarts, which is before it takes in TM's, so after TMing it I just had to restart. Easy.

NuIIpunkt:
@Jamdrawers the wretch camp's east exit seems to lead no where?

I played as hedge mage and was asked by another wretch to assist them in leaving
Turned out I didn't know the exit either!

West entrance works as intended

PR #1296, opened 2026-04-09
nerf watcher fireball cooldown

Xander3359:
## About The Pull Request
nerf watcher fireball cooldown from 3 seconds to 20 seconds

## Why It's Good For The Game
Nothing should have a full fireball spell on a 3 SECOND cooldown.
The normal spell is 15 and costs stamina, so this should be fine at 20 since they can cast it infinitely.

## Changelog

:cl:
balance: Nerfs watcher fireballs to 20 seconds cooldown
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/retaliate/summons/infernal/watcher.dm
+++ b/code/modules/mob/living/simple_animal/hostile/retaliate/summons/infernal/watcher.dm
@@ -45,7 +45,7 @@
 	aggressive = 1
 //	stat_attack = UNCONSCIOUS
 	ranged = TRUE
-	ranged_cooldown = 40
+	ranged_cooldown = 20 SECONDS
 	projectiletype = /obj/projectile/magic/aoe/fireball/rogue
 	ranged_message = "stares"
 


Jamdrawers:
Going from 3 to 20 is a pretty drastic change.

Jamdrawers:
Does this effect NPC watchers or just player watchers?

ALIEN19870:
Oh my god. Absolutely not. Fire has like actual counters that take 10 minutes at best to get: and then you're just immune. Fireballs are literally the only thing watchers do, there's nothing else in their kit and it's the only thing that makes them up to par with the other t3's. This PR comes from the circumstance in which there's more then one, in which case. Why are we punishing antagonists for having a more creative strategy then walking in and using cheesy spells or clicking on people till they die. Watchers ARENT bad enough for this PR to be made.

UntoldTactics:
> Oh my god. Absolutely not. Fire has like actual counters that take 10 minutes at best to get: and then you're just immune. Fireballs are literally the only thing watchers do, there's nothing else in their kit and it's the only thing that makes them up to par with the other t3's. This PR comes from the circumstance in which there's more then one, in which case. Why are we punishing antagonists for having a more creative strategy then walking in and using cheesy spells or clicking on people till they die. Watchers ARENT bad enough for this PR to be made.

On Scarlet Reach you'd have players that'd summon like 8 watchers and simply be able to run right into the keep with an army and crit everyone to death. There's only 1 counter, antimagic stuff. The other counters aren't full ones. Direct damage from a fireball will always do heavy heavy damage, even if you're enchanted with fire resistance. This will crit you eventually.

This 'creative strategy' is one of grinding mindlessly for an hour or more to then frag out immensely hard with funny orbs of fireballing.

The only accessible measures of antimagic to my knowledge are Being the Martyr with hollow trait, being a sojourner or having a null-magic ring or having someone to use counterspell on you. Of the non-permanent methods, null magic rings and counterspell are unreliable and infrequent at best, though in tandem and with strategy they can be effective, you can simply be ignored for the 30 seconds-2 minutes that these traits last, while they run away from you or are batted away by your summoners and their allies.

Watcher fireballs are very strong, the cooldown increase here is way too much, but I agree that something has to be done regarding them.

zeroisthebiggay:
twenty seconds really sounds like it's too much
especially with tidi potentially doubling that effective cooldown

Genevesh:
Twenty is too much. They're simplemobs and that's literally the only good thing about them. If you're going to 'fix' the meatballs in question, then please do fix their melee attack ability not working too - otherwise, just make it fifteen seconds.

PR #1297, opened 2026-04-09
implements broke_free

Xander3359:
## About The Pull Request
Implements a cooldown for re-grabbing someone after they resist out of a grab,

Curently I set it to 5 seconds, maybe adjust as needed according to testing? /shrug

## Testing Evidence
image

## Why It's Good For The Game
Trying to make grabbing less oppressive. 
If you manage to break out of a grab, being IMMEDIATELY regrabbed is suffocating, especially with how expensive resisting can be if you have lower con / skill

## Changelog

:cl:
balance: Resisting out of a grab makes you ungrabbable for a short duration
/:cl:

### DIFF
--- a/code/_onclick/item_attack.dm
+++ b/code/_onclick/item_attack.dm
@@ -92,12 +92,6 @@
 		worn_thing.hit_response(src, user) //checks if clothing has hit response. Refer to Items.dm
 	return I.attack(src, user)
 
-/mob/living
-	var/tempatarget = null
-	var/pegleg = 0			//Handles check & slowdown for peglegs. Fuckin' bootleg, literally, but hey it at least works.
-	var/construct = 0
-	var/burialrited = FALSE
-
 #define ATTACK_OVERRIDE_NODEFENSE 2
 
 /obj/item/proc/attack(mob/living/M, mob/living/user)
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -412,6 +412,7 @@
 			target.visible_message(span_warning("[target] slips from [src]'s grip."), \
 					span_warning("I slip from [src]'s grab."))
 			log_combat(src, target, "tried grabbing", addition="passive grab")
+			stop_pulling()
 			return
 
 		log_combat(src, target, "grabbed", addition="passive grab")
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -408,7 +408,7 @@
 				return
 
 		// Makes it so people who recently broke out of grabs cannot be grabbed again
-		if(TIMER_COOLDOWN_RUNNING(target, "broke_free") && target.stat == CONSCIOUS)
+		if(COOLDOWN_TIMELEFT(target, broke_free) && target.stat == CONSCIOUS)
 			target.visible_message(span_warning("[target] slips from [src]'s grip."), \
 					span_warning("I slip from [src]'s grab."))
 			log_combat(src, target, "tried grabbing", addition="passive grab")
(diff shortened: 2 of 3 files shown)


hugmeorelse:
I like the follow-up from the Discord discussion with the players, although maybe three, rather than five, is a good start. Something conservative to begin with.

coffeecoffeecoffee:
I don't think grabbing _is_ very oppressive outside of the fact that so, so many roles have really goddamn high levels in it. Could it be that we gave most of the garrison extremely high levels in it, and gave the bandits a class specifically geared for it as it's whole identity, might be the majority of the problem?

Jamdrawers:
once you've broken out of a grab you can move away, dodge, and parry the incoming follow-up grabs.

And if you're massively outnumbered and focused... of course you should be grabbed!

NotGoldenglow:
> once you've broken out of a grab you can move away, dodge, and parry the incoming follow-up grabs.
> 
> And if you're massively outnumbered and focused... of course you should be grabbed!

I actually enjoy having a short grab immunity, because it encourages less 'grab grab grab spam spam spam!!!' from 30 people in a fight, and more 'only grab them if you think you can hold them'. 

Previously, 8STR McBait Squire would keep grabbing even if they lost the grab a second after BECAUSE it still gets in their way and slows the person for a second. 

Besides, breaking out of grabs is easy enough.
**All this does is punish mindless grabbing. This is good.**

CameronWoof:
If grabbing someone is going to make them ungrabbable, then I just won't do it. My only other option is to stab them in the face until they stop moving. This is going to increase lethality significantly, especially in group chases where people won't want to risk fucking up The Grab (single) that you get before the knave bolts off at speed 20 and will just go lethal on them instead.

Crowbar764:
Can someone tell these devs to stop changing whatever they feel like.
Why is there no quality control or discussion, just straight to test merge and then commit.

vigeta:
Five seconds might be a bit long, but this helps combat the grabspam we've all become accustomed to.

coffeecoffeecoffee:
What I've seen so far is that anybody looking to take somebody alive goes
>Grab him!
>Shit, he broke free?
>Break every bone in his body, split his head open, I guess we have to. It was his choice.

ThinkArbies:
oh, another speed-TM'd pr.

admittedly in a better direction than the last one, but i don't think this is it either.

as i said in a previous PR, there is a **huge** valley of difference between John Garrison and his peers with expert-level wrestling and 12-13 strength and a Drakian or half-orc iconoclast with master grappling and struggler to hit 15+ strength. there's something about the way specifically master-level wrestling that makes it so outright oppressive, but anything less than that weak.

for reference, as a XIV strength character with expert grappling, on speed builds and thieves i still only average about 60% per resist attempt, and considering they can resist every 2 seconds that's a very, **very** good chance to escape the second or third grab.

target the way it scales with strength, and you can effectively nerf the power of specifically master wrestling without making it completely useless for everyone else

Floglor:
Good change. Grabs are the ultimate weapon, and they had no counter because if someone broke free, they were still grabbed the next frame. 

As for why there's no discussion or anything. Because nobody asked you specifically, that's why. Yes, that's a community-driven game, but developers fixing critical combat issues shouldn't be bottlenecked by a random ahh dude who doesn't consent.

CaydenT:
Tell me that this does not apply to mobs?? Grabbing is often the best way to handle mobs in the world and they often break free immediately and need to be re-grabbed.

Five seconds sounds WAY too long. To be perfectly honest it needs to be like two seconds, tops.

kimesik:
Good change.

Jamdrawers:
Is this still 5 seconds?

DogExtrordinaire:
i said this on azure to their changes but this feesls like an entire mechanic suffering for the sins of 3 classes (iconocalst, berzerker, and disciple)

RenGusta:
Can you maybe bugfix phantom grabs on NPCs and shit as a consequence of this

NotGoldenglow:
I second that. There's currently a bug present with NPCs being phantom-grabbed if they die? While being held, as well as if someone is dragged up a zlevel. The only fix is to grab and drop something else.

Xander3359:
> Can you maybe bugfix phantom grabs on NPCs and shit as a consequence of this 

Working on it.

PR #1300, opened 2026-04-09
[Yeoman/Town Content] Presence Flags (Ported from AP)

IronDragoon:
## About The Pull Request

(Details follow as direct from the original port as it's pretty much 1:1 and all credits to @Ketrai for another great little PR!)

(Original PR: https://github.com/Azure-Peak/Azure-Peak/pull/6197)

Adds towner flagpoles!

- Represent different roles in the town and indicates whether one is present and active.
- Corresponding mini flags allow those town 'groups' to raise or lower the flag (this affects all of the flagpoles on the map)
- They show a description of the sort of services usually offered, custom messages can also be set
- University based roles and the duke don't spawn with the flag, instead having them mapped into their buildings. Some spares are also mapped into church, apothecary, bathhouse, etc.
- Most flags are role locked to prevent them being used by those who should not.
- I've mapped some of the stationary 'main' flagpoles to both Dunworld and Rockhill, in areas I felt they fit, but Jams feel free to change this! 
- AP also made some of these in the wretch camp, to let wretches see what town services are available, but we have a centcom level rework coming up so will look at that once it's done.

## Testing Evidence

image
(Using their image, but i've tested this locally and works exactly the same) 

## Why It's Good For The Game

Adds a way for various roles inside the town to indicate they are open for business in a quick and visually appealing way, regardless of distance. Should make it a little easier to find out if the duke is accepting petitions, if the archivist is in, or if the apothecary is manned, for example. 

There's flagpoles for the following roles.
blacksmith
artificer
steward
duke
apothecary
church
fisher
university
innkeeper
tailor
bathhouse
merchant

## Changelog

:cl:
add: Added town flags in locations and mini flags to allow certain town centered roles to signify their presence in a round, with space for custom messages
/:cl:


john-pawson:
Finally, medieval crew manifest.

Lir-sys:
What's next? Flag colored uniforms and painted guidance lines with flag colored town sections?

IronDragoon:
@Jamdrawers I want to have this merged, but it has a small amount of mapping content. If I update to main and do the changes, will it be able to be TMed or is some mapping stuff already tmed? Just adds a few items and static flag poles on dunworld and rockhill

Xander3359 (changes requested):
Why is this a subsystem at all?
It does nothing a subsystem does whatsoever. 

Bruh.

Make it not a subsystem thanks.

IronDragoon:
@Xander3359 should be adjusted now, does this look better? Honestly, I just ported it and did not think too deep about why that was less than optimal, my bad.

EmilyFontaine:
Pretty sure this is messing with Merchant's starting inventory. They're meant to have a gunpowder flask and lead bullet pouch in their satchel, to load their glizzy, but the only thing inside their satchel is the miniature flagpole.

code/modules/jobs/job_types/roguetown/yeomen/merchant.dm
```diff
@@ -94,3 +94,6 @@
 	if(H.mind)
 		H.mind.AddSpell(new /obj/effect/proc_holder/spell/invoked/appraise/secular)
 		H.mind.AddSpell(new /obj/effect/proc_holder/spell/invoked/takeapprentice)
+	backpack_contents = list(
```
john-pawson: backpack_contents is an override here, just modify the one on line 63

Jamdrawers:
UN-tm-ing for until mapping conflicts are dealt with. Also I think the merchant bug.

IronDragoon:
> UN-tm-ing for until mapping conflicts are dealt with. Also I think the merchant bug.

what mapping conflicts? I can't fix them unless you tell me what is conflicting. I'll fix the merchant inventory.

Jamdrawers:
I would lean against this one, it feels a bit inorganic and I don't like the way it clutters the inventory - and it doesn't really tell you if the booth is manned, just if they've loaded in to the round or not, which isn't super useful information.

That said, it's inoffensive enough that it's not the end of the world if it gets merged- which we might as well do sooner rather than later to avoid getting into map merge problems

PR #1305, opened 2026-04-09
QoL - Item Examine Window Popups

ophaq:
## About The Pull Request
- Ports & Reworks with a slightly different coat of paint the examine functions of this PR: https://github.com/Azure-Peak/Azure-Peak/pull/4523

## Testing Evidence
I tested it pretty thoroughly. Hopefully nothing messed up. Didn't seem like it from when I tested locally.
examinetest1
examinetest2
examinetest3

## Why It's Good For The Game
We have been needing a nicer examine UI that isn't clicking every damn item on examine and using the ? tooltip. You can still do that but now you can check things at a glance after examining yourself or someone else. It's a big QoL!

## Changelog
:cl:
add: Pop up windows on mouse over for armors and such when examining people. Shows the description too. Custom items stay purple instead of blue to let you know it's custom! You can mouse over those too to read their flavor texts at a distance. You can check your exact durability on items too, but other players won't the exact numbers or % if they examine you.
qol: No more having to click every item on your person with the ? mark to check durability when you can just examine yourself and check item durability states.
/:cl:


john-pawson:
Don't know if showing exact durability numbers is the best idea, specially considering how the little damage messages and knowing the exact amounts of protection your enemy has already scale off INT, and usually can't be done by lower INT characters.

ophaq:
> Don't know if showing exact durability numbers is the best idea, specially considering how the little damage messages and knowing the exact amounts of protection your enemy has already scale off INT, and usually can't be done by lower INT characters.

I don't really think it is a big deal, but it could give people a leg up in choosing where to target next in PvP. If necessary, I could add a feature where it only shows the durability % and numbers for the player examining themself and not for others. Or just remove durability from the popup window altogether.

EDIT: I made it self-examine only. That way in PvP, players are in the dark on their enemy's integrity on examine. They can still see a generalized damage state though, but no more exact numbers or percent.

Jamdrawers:
sweet

john-pawson:
image
Bug!

ophaq:
> Bug!

I'll set keyrings to the exclusion list... later.

PR #920, opened 2026-02-26
NECRA REIGNS! OR: How I Learned To Stop Worrying And Gave The Duke The PSYDON Treatment. (DNR Trait.)

JustMeTheIInd:
## About The Pull Request

**As of right now, this PR only adds the DNR trait to the role of Grand Duke, and onto any newly-crowned Duke. Below I leave a 'legacy' version of the PR, as I still personally view the 'inital' version of the PR to be the 'true' and the 'desired' version.**

This PR gives the DNR trait to the following roles: The Grand Duke, The Prince, The Hand, The Knight Captain and
 The Marshal.

To maintain consistency, the Prelate's Crowning ability also gives the target the DNR trait. 

## Testing Evidence

Zrzut ekranu 2026-02-26 132956
Zrzut ekranu 2026-02-26 132844
Zrzut ekranu 2026-02-26 132810
Zrzut ekranu 2026-02-26 132744
Zrzut ekranu 2026-02-26 132634

## Why It's Good For The Game

Generally, the ability to just say "no" to the Grim Reaper really does kinda cheapen the death. At the worst of times it's just a Time Tax, and nothing actual comes out of it. And honestly? I personally believe that's fine. At least when it comes to towners or acolytes or mercs or whomever, because their deaths very rarely if at all result in any actual interesting scenario. No one really cares if John Fisherman, the local fisherman, died and cannot be revived. 

I believe the situation to be quite different when it comes to the actually "important" roles however. The presence, or  the sudden lack of it tend to greatly impact the round and the wholesale situation in the Duchy. For example, the irreversible death of a Duke is sure to cause some chaos, as the Church and the Keep try to agree on who exactly should inherit the Throne. Likewise, the death of a skilled Marshal or Knight Captain will be sure to put the Garrison into chaos, opening up avenues for the Evil People (TM.) to swoop in and take advantage of the situation. 

Through this, the PR aims to improve RP potential around the major Leadership roles and the possible, sudden and irreversible death of them. 

Maybe the Prince and the Hand will actually consider their safety and security? Maybe they shall requisition Guards or hire mercenaries as bodyguards? Maybe they shall grow paranoid and dare not stray from the well-lit spots of High-Town?

Maybe the Duke will think twice before enjoying the wine that the strange drow servant so eagerly handed him? Maybe he'll designate someone to ingest some of the wine, to ensure it's free of poison? (actual thing btw. Medieval royalty often used food tasters or cupbearers to ensure that the food and drink they are about to eat wasn't poisoned.)

[...]

PR #1306, opened 2026-04-10
Revert NECRA REIGNS! OR: How I Learned To Stop Worrying And Gave The Duke The PSYDON Treatment. (DNR Trait.)

Pidgey-chan:
Reverts Rotwood-Vale/Ratwood-2.0#920

**Why It's Good For The Game**
no lore reason for this + the duke can be oneshot by the Matthios T3 with no counterplay + raises awkward rules questions

also the original pr had a ratio like this
image

### DIFF
--- a/code/modules/jobs/job_types/roguetown/church/priest.dm
+++ b/code/modules/jobs/job_types/roguetown/church/priest.dm
@@ -173,7 +173,6 @@ GLOBAL_LIST_EMPTY(heretical_players)
 /mob/living/carbon/human/proc/coronate_lord()
 	set name = "Coronate"
 	set category = "Priest"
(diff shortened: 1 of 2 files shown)


YakumoChen:
just port the churnwealth nerf from ap

zoktiik:
You're intentionally leaving out that the PR originally had DNR for a bunch of other roles, then changed it to only be the Duke and his *coronated* successors. A lot of the downvotes were from that before-time, and you know it.

zeroisthebiggay:
nerf churn wealth instead

XxWawawiwaxX:
If Matthios T3 is such a problem, then you can tweak that instead

PR #1307, opened 2026-04-10
Syicide

DelineFortune:
## About The Pull Request

adds self round r*moval emote
https://github.com/Scarlet-Reach/Scarlet-Reach/pull/638

## Testing Evidence

image

## Why It's Good For The Game

/me farts and dies

High heavy role play


Lir-sys:
Idk if we want people a tool to avoid IC consequences by committing suicide whenever they get caught.

DelineFortune:
> Idk if we want people a tool to avoid IC consequences by committing suicide whenever they get caught.

they already can
by pressing hold breath

code/modules/mob/living/emote.dm
```diff
@@ -2446,3 +2446,44 @@
...
+    key = "praysuicide"
+    key_third_person = "utters their last words"
+    message = ""                   
+    emote_type = EMOTE_AUDIBLE
+    stat_allowed = UNCONSCIOUS      
+    show_runechat = FALSE
+
+/mob/living/carbon/human/verb/emote_praysuicide()
+    set name = "Suicide"
+    set category = "Emotes"
```
zoktiik:
```suggestion
    set category = "IC"
```
I think it'd be better if this wasn't included in the standard Emotes tab. You could say the same about Hold Breath, maybe, but Emotes is already populated enough as is. I also think seeing this verb in the Emotes list would be a little too grim for what that tab is meant to be.

NotGoldenglow:
Half the people I capture post-fights expresses displeasure about being captured, and stonewall RP or just, y'know, kill themselves.

Let them die so I know they don't want to be here and can bury them appropriately.

PR #1309, opened 2026-04-10
Ports some bronze armour from somewhere

Jamdrawers:
## About The Pull Request

Porting from Twilight Axis - I'm not sure if they ported it from somewhere else. 
I edited it to just use Iron stats instead of having a unique special integrity that's ever so slightly different from iron and balanced around giving Fire and Acid resistance. They used defines that we don't have and also I thought it was just unnecessary.
Four pieces, a cuirass with leg-bits, a heart protector (light, but only guards chest), and a 'panipoly', which is like full plate. It has a special mechanic that gives you a -1 speed and a +1 con when worn.

## Testing Evidence

image
image
image

## Why It's Good For The Game

Mostly intend to use it on Desertmap, but it's nice to have a use for bronze.

## Changelog

:cl:
add: Adds craftable bronze armour
/:cl:


SciDragon:
I don't blame you for not finding where it was from, it was part of this massive unatomized PR on Azure Peak LOL
https://github.com/Azure-Peak/Azure-Peak/pull/5331

PR #1310, opened 2026-04-10
Removes knuckles punch intent integrity damage bonus

UntoldTactics:
## About The Pull Request
Removes the modifier that adds 40% integrity damage to knuckles punch intent.
As the comment states, tis was a mistake.

## Testing Evidence
One line change that compiles.
image

## Why It's Good For The Game
Knuckles are above almost any other weapon in relative time-to-kill(or skull crack in this case), barring exposed armor circumstances. Due to the way helmet armor works, the ease of which unarmed classes can obliterate swarthes of people relative to other combatants in the game is simply stupid. A disciple or berzerker can cause more damage in combat than 3 other combatants of any given class. Considering this and the fact that unarmed classes likely have the highest number of people with master skill level (a bygone relic when unarmed was significantly weaker/useless), unarmed users have unparalelled ease to attack people, bypass parrys or dodges and obliterate peoples skulls.

They're so good that many players of classes typically not expected to wield unarmed weapons have gone around using them alongside a shield or just plain grappling and smacking heads in - because they are that good.

This is an admittingly lazy but simple fix that reduces the effective integrity damage they can do, reducing time to kill significantly, putting them more in line with other weapons considering their quick attack speed.

This comes from significant experience playing otavan orthodoxist disciple, as well as seeing what unarmed users on the antagonist side can do. It's fun to play, but absolutely mindnumbing to everyone else. In theory, you can counter these users, in reality they have builds that allow them to sustain much more abuse and thus it hardly matters if you get a bait or block off- they'll just keep coming at you. Playing literally anything else in combat feels like playing hard mode whenever a grappler or knuckles user is around. The sheer disparity between strong unarmed classes and everything else is to the point that when I play a disciple? I have absolutely 100% confidence that I can win a fight against pretty much any player on the server. No class should give anyone that level of certainty, ever.

## Changelog

:cl:
balance: Knuckles punch intent no longer recieves a 40% bonus integrity damage modifier.
/:cl:

### DIFF
--- a/code/game/objects/items/rogueweapons/melee/unarmed.dm
+++ b/code/game/objects/items/rogueweapons/melee/unarmed.dm
@@ -36,7 +36,7 @@
 	swingdelay = 0
 	icon_state = "inpunch"
 	item_d_type = "blunt"
-	intent_intdamage_factor = BLUNT_DEFAULT_INT_DAMAGEFACTOR // This might be a mistake
+	intent_intdamage_factor = 1
 	//We want chipping, m'lord.
 	blunt_chipping = TRUE
 	blunt_chip_strength = BLUNT_CHIP_WEAK
--- a/code/game/objects/items/rogueweapons/melee/unarmed.dm
+++ b/code/game/objects/items/rogueweapons/melee/unarmed.dm
@@ -36,6 +36,7 @@
 	swingdelay = 0
 	icon_state = "inpunch"
 	item_d_type = "blunt"
+	intent_intdamage_factor = BLUNT_DEFAULT_INT_DAMAGEFACTOR // This might be a mistake
 	//We want chipping, m'lord.
 	blunt_chipping = TRUE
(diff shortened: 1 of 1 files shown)


Lir-sys:
Uh this is not how you reduce the integ modifier by 40 percent
Just turn 1.4 to 1.0 here. This PR might actually break something if you remove that line. Other than that this is a very reasonable PR.

https://imgur.com/XCJwlV8

UntoldTactics:
> Uh this is not how you reduce the integ modifier by 40 percent Just turn 1.4 to 1.0 here. This PR might actually break something if you remove that line. Other than that this is a very reasonable PR.
> 
> https://imgur.com/XCJwlV8

It shouldn't matter as when a variable is lacked it inherits from the parent. If it defaults to datum/intent that means it'll be set to 1 automatically, but I'll do it anywho.

PR #1316, opened 2026-04-11
Draft: Councillor Minister system

PPooch:
## About The Pull Request

I faced the question, "What do councillors do?" They can get assigned to special tasks, but are quite mediocre at them. I also noticed some conversation about a "Keep Chaplain" role, but how that would disenfranchise the Church itself, et cetera. I thought, what if these councillors could go out and ingratiate themselves to the power roles of the town, be given a snippet of that role's abilities, in exchange for being a "voice inside the keep" for their purposes? That seemed like an interesting dynamic to encourage Councillors to get out and move and shake.

There is an old archive cabinet in the councillor's meeting chambers that only the councillor job can use to spawn a default "writ". They can go out to the following jobs; Guildmaster, Bishop, Innkeeper, Bathmaster, Inquisitor, Court Mage, Court Physician, and Merchant, and try to earn the patronage of that role to establish a ministry. The faction head uses the writ, which finalizes it. 

The councillor brings the writ back to the archive cabinet and submits it: this gives them some bonuses to skills apparent to that role, an appropriate key, some spells in the case of the Mage and a surgery bag in case of the Physician, as well as some devotion and miracle mechanics for the Church and Inquisition.

This also spawns a paired signet ring / seal combination (sprites pending, maybe). The councillor will give the seal to the faction head and this allows a direct, two-way communication between the faction head and the minister. I haven't put any sort of reclaim or respawn code in, but that is certainly possible to add to the cabinet if wanted. 

Councillors with a valid ministry will show up as "Minister of X" in their examine string. The councillor gets a bonus to their noble estate income to incentivize their minister position.

## Testing Evidence

Testing writ use and ring / seal / keys spawning and skills + special bonuses.

https://github.com/user-attachments/assets/84dc947f-63f4-4320-926d-923e77e795f7

## Why It's Good For The Game

Councillor is very much the "newbie noble" role. They are one of the few "free-form" roles left in the game, someone said. I wanted to incentivize the councillor to go out and be a "go-between" with a faction head and the keep - not only just for newbie nobles to learn the role, but also so that experienced players have enough meat on the bone to make something happen. This also handles the question of, "Do we have X but keep-oriented?" in the case of Court Chaplain, etc. In order to not step on the power role's toes, this ministry creation is completely voluntary and (ideally!) mutually beneficial for both parties. The councillor gets utility in skills and abilities; the power role gets a direct line to the keep. This in-round decision would hopefully let councillors do a benefit analysis of what roles are present, whom to be the middle-man for; as opposed to locking in at round start like subclasses would accomplish.

N.B. llm assisted code, hence draft.

## Changelog

:cl:
add: Councillors can now establish Ministries with heads of town factions, gaining some of their appropriate skills.
/:cl:

### DIFF
--- a/_maps/map_files/roguetest/roguetest.dmm
+++ b/_maps/map_files/roguetest/roguetest.dmm
@@ -338,6 +338,10 @@
 /obj/item/inhand_tester,
 /turf/open/floor/rogue/cobblerock,
 /area/rogue/indoors)
+"mO" = (
+/obj/structure/roguemachine/minister_archive,
+/turf/open/floor/rogue/grass,
+/area/rogue/outdoors)
 "mW" = (
 /obj/structure/curtain/red,
 /turf/open/floor/rogue/concrete,
@@ -2938,7 +2942,7 @@ zo
 wQ
 yK
 vq
-vq
+mO
 vq
 vq
 eg
--- a/code/modules/jobs/job_types/roguetown/courtier/councillor.dm
+++ b/code/modules/jobs/job_types/roguetown/courtier/councillor.dm
@@ -68,3 +68,413 @@
 		armor = /obj/item/clothing/suit/roguetown/armor/leather/newkeep/councillor
 		cloak = null
 		mask = null
+
+// ===== MINISTER SYSTEM =====
+// Councillors draft a writ at the ministerial archive. They bring it to a
+// faction leader, who finalizes it. The councillor then returns the finalized
+// writ to the archive to be sworn in, gaining an income boost, signet ring,
+// keys, and archive benefits (skills and traits that match their position). 
+// The patron receives a seal to contact their minister, 2-way private comms.
+
+/datum/ministry
+	var/name = "Ministry"
+	var/display_title = "Minister"
+	var/income_bonus = 20
+	var/list/ministry_keys = list()
+	var/list/archive_traits = list()
+	var/list/archive_skills = list()
+	var/ring_type = null
+	var/seal_type = null
+	var/mob/living/carbon/human/councillor = null
+	var/mob/living/carbon/human/partner = null
+	var/active = FALSE
+
+/datum/ministry/proc/archive_bonus(mob/living/carbon/human/H)
+	return
+
+/datum/ministry/guild
+	name = "Guild Ministry"
+	display_title = "Minister of Crafts"
+	ministry_keys = list(/obj/item/roguekey/crafterguild)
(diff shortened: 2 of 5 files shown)


JackSage:
Really cool idea, although not a big fan of the minister of health or magic as both of the "heads" are already court members. I'd rather see councilors representing/communicating with non-court members to represent that part of town/faction in the duke's court.

maker-of-nightmares:
The only suggestion I had you already hit on in the code though I would say making the minister of vice something less on the nose. Like minister of Appropriations or Minister of Benalities something that isn't inquisitor bait

Makximus:
huge fan of this. would get me to play councilor again

coffeecoffeecoffee:
Hey so, I really love this idea! After playing around in local with some of what these ministries give, a few things probably need to be fixed:

1) What _tier_ of spells does a minister of the mage's guild get? I think it seems to say it doesn't give a tier? Which would seem a little broken in a useless kind of way. Otherwise seems really cool - T2 or something would seem pretty fair.

2) Inquisition minister is **busted,** and it's not the weapon skills; expert in climbing and athletics is **insane** for a character like a councilor, and allows them to do feats of acrobatics that feel ill-befitting of a _minister_ of anything. I'd be very comfortable seeing it reduced and made up for with something like whips & flails - so maybe they can join in on some of the torturing! What's more, you have to keep in mind that the inquisition is designed around _being an all-psydonite team..._ Because the basic inquisition manor key gives access to **_everything, including the inquisitorial vault, their silver weapons and the heretic relics._** A councilor that (I believe?) isn't patron-locked and happens to be Matthiosite or Zizite would be able to steal things like the avantyne blade. I was testing the basic inquisitorial key in local and it gives them access to the **entire manor** except the actual Inquisitor's personal room. This might be more of a mapping issue with their manor than anything, but something should probably be done to separate "basic access" from "the evil fucking inquisitor's relic vault" before representatives of the Duke are given this key.

Aside from this, this seems awesome. There is sort of the issue that the councilors of the batthouse and Inn could swipe the private room keys, but this is also kind of up to the faction leader of those places as the issue is that they spawn out in the open on shelves. Aside from the inquisitorial manor's issue with keys, this seems like it rocks.

maker-of-nightmares:
@coffeecoffeecoffee I think that might be a mapping issue at the end of the day. Just make a new key and set the "Secret don't want people to see" rooms to it then give the various roles that new key/give the inquisitor a bunch of them he can hand out as he pleases

PR #1336, opened 2026-04-13
Quieter Sex (Stops Sex sounds/emotes bypassing walls/Z-levels)

IronDragoon:
## About The Pull Request

Added a control for sound propogation for emotes to mark them as ignoring or not ignoring walls (this includes z-levels). This defaults to true, so normal emotes will not be affected. I've gone through and added all the sex related emotes I could find to now not propagate beyond walls/z-levels.  This also requires some specific emotes having separate sex-specific keys to prevent normal emote usage being affected. Performance impact should be nil (or at least, functionally so)

TL;DR: Sex noises will not bypass z-levels or walls anymore

## Testing Evidence

Tested on localhost in a mixture of situations and seems to work as intended. 

## Why It's Good For The Game

We have long had issues with intrusion onto private scenes and the fact that any sex related noises were bypassing walls AND z-levels was a big factor in this, which should now no longer occur. Additionally, some areas of the map could become rather frustrating to work in (the inn, keep kitches, etc), due to not being able to actually do anything about having to hear these sounds. 

## Changelog

:cl:
sound: Sex-con sounds and emotes no longer bypass walls or z-levels
/:cl:


coffeecoffeecoffee:
Sovl removal, our greatest filter for fun-destroying ERP hunters to get caught and banned-removal
I am fully against sanding down every potential point of conflict that encourages people to find out-of-the-way places to boink, and think that having things like this to catch the people who abuse them to ruin other's rounds will lead to a significantly worse community in the long-term. Admins should and do ban for this (and should probably ban them for much longer.)
I really, really hate saying this, but this is legitimately one of the few cases of an actual policy / staff action problem, not a code problem.

I would say perhaps you could reduce the range and I'd be in favor of that, though. One floor up and down at max, or something. We already HAVE options implemented to do gentle sex acts quietly / "doing subtly" - I strongly suggest expanding that, instead, so that people who want to have quiet sex can, and people who want it otherwise (me) can go at it at our own risk! Maybe we could have "normal" force work like you suggest, and leave "brutal" untouched?
Or just apply your change to "normal" force sex interactions?

glitchbow:
> I would say perhaps you could reduce the range and I'd be in favor of that, though. One floor up and down at max, or something. We already HAVE options implemented to do gentle sex acts quietly / "doing subtly" - I strongly suggest expanding that, instead, so that people who want to have quiet sex can, and people who want it otherwise (me) can go at it at our own risk! Maybe we could have "normal" force work like you suggest, and leave "brutal" untouched?
> Or just apply your change to "normal" force sex interactions?

This, this could be reworked as subtly / normally / LOUDLY.
Subtly is the quietest sneaky option, same as it is now.
Normally is sex sounds only within the room, like this PR has it.
LOUDLY should be as it's been, where it passes Z-levels and is obvious.
Default should be either subtly or normally, so you know LOUDLY is a choice that was actively made.

The-Sun-In-Splendour:
make it a toggle so i can choose wether i want the entire garrison to hear me absolutely goblin the princess or not

coffeecoffeecoffee:
> This, this could be reworked as subtly / normally / LOUDLY. Subtly is the quietest sneaky option, same as it is now. (...)

I'm not actually sure if the "gentle" force option _is_ actually quiet and sneaky, to be fair. The "doing subtly / doing loudly" is it's own option for certain ERP interactions that doesn't exist for many acts! It could totally be that gentle/normal force sex options are disproportionately loud right now, and that's something that could be alleviated. Might test this to see how bad it really is and how loud "gentle" actually sounds.

IronDragoon:
> make it a toggle so i can choose wether i want the entire garrison to hear me absolutely goblin the princess or not

I am an exhibitionist but I personally don't think it's very fair to have to force others to listen to your sex if they do not wish to. I've had plenty of times trying to work as a servant and constant sex noises just drove me insane. I'm merely putting this out as a potential solution as it was discussed many times, and this was often suggested. I'm going to leave it as is and it's up to maintainers/Useroth as to whether this is desired or not.

zoktiik:
> Sovl removal, our greatest filter for fun-destroying ERP hunters to get caught and banned-removal... *[snip]*

You do realize that the rules are written to deter such interruptions from happening right? Why tf do we need to encourage randoms to lock on like heat-seeking missiles? It's more trouble than it's worth. And all the levity that sexo brings, when you hear it from *way too fucking far away*, has certainly overstayed its welcome.

ophaq:
I'm in favor of making it a toggle. I always thought it was funny I could hear people fucking across a building. Maybe have it based off of perception instead for hearing range? And the toggle for it, or make rough / brutal modes cause moans to be loud, and anything below that is quieter. And if you are directly under the floor where the sex spot is or near the door where it is happening, perhaps you can hear it from there too.

Anyway, hearing sex across z-levels never bothered me too much but I do agree that it could use a reduced range at the least.

Jamdrawers:
Alternatively it could be Area-based, so that as a mapper I can just set certain areas to be more soundproofed than others.

ThinkArbies:
Could make it so if you stand on a tile adjacent to a door or a window you can still hear the sex sounds.

That way a conscientious thief or intruder might hear them and opt not to disrupt a scene or someone can go and figure out where the KC's been for the past hour by putting an ear to his door.

zeroisthebiggay:
it being force based would be kinda cool
i tried quietly eating out a princess once but people in the kitchen could hear it that shit is crazy why's it so loud

PR #1350, opened 2026-04-15
reduces spellcost of message and mindlink by one

zeroisthebiggay:
## About The Pull Request

maintainerman told me i could do this as a subsitute to making them more accessible to court mage roles by making them free to them

## Testing Evidence

it's a webedit sir

## Why It's Good For The Game

theres are spells that mostly exist to facilitate roleplay and coordination, but people often skip out on them because they're the same cost as things with actual usecases like mending
i think bringing them down a point and especially message down to the one point mark where you spend all of your spellpoints and have just one point left and you have to choose between light or campfire or some bullshit and you'd really rather just have mending like c'mon will encourage court mages to actually take message instead of going all in on fireballs and other way more interesting utility spells

## Changelog

:cl:
balance: reduces spellcost of message and mindlink by one
/:cl:


Xander3359:
Good code 👍

PR #1355, opened 2026-04-15
Salt-mine tweak: adds time-served multiplier and sentence manager

MickDick1:
## About The Pull Request

There will now be a time served multiplier that increases, maxxing out at 5x at the 1 hour mark.
image

In addition, this PR adds a management machine the dungeoneer can use to customize the quota for each prisoner.
image

It also ensures return prisoners have the multiplier for their sentence reset.

And finally full variable control is now available on a separate machine.
image

## Why It's Good For The Game

This solves a few issues wrapped up into one package. If you're not actively mining salt, then you're not progressing towards your sentence. Which means doing anything else actively halts the prisoner's momentum and forces them to single-mindedly whack rocks if they'd like to eventually meet the quota. It now serves as both a timed prison sentence and a labour one wrapped into the same package, while also allowing the dungeoneer to customize the quota/sentence length and return prisoners have their sentencing reset after successful serving.

## Changelog

:cl:
add: New machine for managing prisoner sentencing/quota/variables
code: Added a multiplier to the saltmine tracking machine that increases over time
add: Nervemaster now gets reports on prisoners in the mine
/:cl:

### DIFF
--- a/_maps/map_files/generic/CentCom.dmm
+++ b/_maps/map_files/generic/CentCom.dmm
@@ -2071,6 +2071,7 @@
 /area/rogue/dietroyt)
 "rk" = (
 /obj/effect/landmark/start/dungeoneer,
+/obj/structure/roguemachine/ticket_manager,
 /turf/open/floor/rogue/naturalstone,
 /area/rogue/dietroyt)
 "rl" = (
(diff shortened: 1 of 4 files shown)


Elli-Skala:
Salt PR

Wesreidau:
The Salt Mine, as it stands, is a player-abuse level put into the game without consideration for how it would operate mechanically to create a soul-crushing player experience most prisoners throw themselves into lava rather than engage in.

Please, please, please merge this so that the salt mine can be made lenient enough prisoners engage in the experience, get into their prisoner RP, and feud with one another over tickets.

As it stands, one salt = 0.5% chance of freedom. Its absolutely horrendous. This will let us set it to 5% per salt or better and let minor and major criminals work things off. While they work, they interact, they quarrel, they get into fights, roleplay happens, and they're engaged in the play.

Again. Merge this please, from a former dungeoneer main.

PR #1359, opened 2026-04-16
New Refugee Sub Class: Tinkerer!

Didosforte:
## About The Pull Request

Always wanted an artificer that was not so locked/binded to the guild and such. Being allowed to roam more frequently, and offer their services to other factions ( keep, church, lowtown, or even the 'evil' forces, work is work for a Malunite after all ).

## Testing Evidence

Final Name Adjustment, maybe
The New NEW desc
Changed Skill Set
[ note about the skill level in engineering which is at expert, this is with a bronze arm virtue, which gives a level in it ( yet they start at Journeyman ) ]
Stats   With industrious stat pack
[ note about stats, this is with the industrious stat pack, as an human ]
passives
Items and Spells below
Backpack stuff
Drip

## Why It's Good For The Game

Currently, very few people play as an artificer. One of the reasons I see this happening is because of the limitation to the environment they are in, always in the workshop and not being able to leave much without major problems. [ Unlike blacksmiths and others, which have their variants like refugees and even towners ].

They can't explore what they can do being always confined.

So I decided to implement this to provide this variety/alternative and perhaps get more people to try the class. So they can play in a manner more relaxed, without much obligation to test this one.. See if they like it.. And perhaps create all sort of odd things. (in a similar way to what I do)

## Changelog

🆑
add: A sub class to the pilgrim/refugee, The Artificer
balance: Adjusted its stats, and starting items. It has some tools necessary to work, while having less skill levels and stats, in comparison to the normal one.
balance: Improved one thing to allow them to defend themselves ( mace skill ), to journeyman. They are a refugee. They will be in the wilderness. They need to at least be able to fend off a rous.
balance: Removed weaponsmith and armorsmith skills from them. They still can do it, yet will have to practice in game for it.
qol: Named it Tinkerer, instead of Artificer, to difer the two [ guildsman and refugee ]
/:cl:


Jamdrawers:
You'll need to rename it at the very least.

Didosforte:
> You'll need to rename it at the very least.

I presume you mean the current name of 'artificer', as to difer it from the normal one? In order to avoid confusion IC and all of that? If so, that can be done easily, yes, yes!

john-pawson:
Btw, changelog uses all the "cl" "add:" "qol:" "balance:" things to work and decide what symbols to use when in-game, so I suggest you use the original template so people can see the changes.

Didosforte:
Adjusts have been made!

Shadovarn:
Take away its smithing. There is no need for it to have that. Especially with everything else.

Didosforte:
> Take away its smithing. There is no need for it to have that. Especially with everything else.

~~The smithing stuff is the same as the normal one for the guildsman artificer variation, which has it [ same level and all of that ]. Why should it be removed? Has it been abused in any manner already since it was added..?~~

Spoke to them, will be making adjustments.

Didosforte:
Right, with the little balances made, and other little things adjusted.. It appears to be all working! A refugee exclusive, with the skill set they need to do their work. Lovely.

Jamdrawers:
I think 'tinkerer' might be a better name for a travelling iterant one.

Didosforte:
> I think 'tinkerer' might be a better name for a travelling iterant one.

Not a bad idea! I shall adjust to that!

PR #1361, opened 2026-04-16
Makes it so you can surrender so long as you aren't dead.

Xander3359:
## About The Pull Request
Makes it so you can yield so long as you arent dead.

## Testing Evidence
image

## Why It's Good For The Game
You should always be able to yield even in crit/skullcrack

## Changelog

:cl:
qol: It's always possible to yield as long as you aren't dead
/:cl:


maker-of-nightmares:
that'll make "Accidental" deaths alot harder to do given surrendering makes it harder for you to die(last I knew)

PR #1362, opened 2026-04-16
Fix combat mode right click blocking most interactions.

Xander3359:
## About The Pull Request
Changes the snowflake checks to just isobj() for special attack

## Testing Evidence
I go on local and combat mode right click sword to alt grip, which worked.
I was still able to feint/parry/bait

## Why It's Good For The Game
Bugfix

## Changelog

:cl:
fix: Fix combat mode blocking most right click interactions.
/:cl:

### DIFF
--- a/code/_onclick/click.dm
+++ b/code/_onclick/click.dm
@@ -956,7 +956,7 @@ GLOBAL_LIST_EMPTY(reach_dummy_pool)
 	return FALSE
 
 /mob/living/try_special_attack(atom/A, list/modifiers)
-	if(!rmb_intent || !cmode || istype(A, /obj/item/clothing) || istype(A, /obj/item/quiver) || istype(A, /obj/item/storage))
+	if(!rmb_intent || !cmode || isobj(A))
 		return FALSE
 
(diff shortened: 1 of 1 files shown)


Jackofashepard:
Yo xander i was very excited to try this but i discovered a bug that kinda makes halfswording useless. If someone wears a plate helmet and a padded coif under ALL integrity damage goes to the coif first completely ignoring the plate helmet effectively making the person blunt immune. One would expect it would first damage the plate helmet and THEN the gambeson would start taking integrity damage once the plate breaks in which case you can switch to proper blade and continue.

zeroisthebiggay:
> Yo xander i was very excited to try this but i discovered a bug that kinda makes halfswording useless. If someone wears a plate helmet and a padded coif under ALL integrity damage goes to the coif first completely ignoring the plate helmet effectively making the person blunt immune. One would expect it would first damage the plate helmet and THEN the gambeson would start taking integrity damage once the plate breaks in which case you can switch to proper blade and continue.

this isn't a bug and it's also not introduced by this pr
weapons will always damage the layer they have the highest resistance towards first when doing damage, that's why sharp weapons damage outer layers first and blunt damages things like gambesons first usually

PR #1363, opened 2026-04-16
Adds query amount to dbcore stat entry

Useroth:
Pretty much


zoktiik:
Oh now that's pretty cool. For anyone curious, this PR makes it so that you can monitor how many database queries are currently pending from the MC tab (as an admin). So if there's something that's really making the game's database handler chug, you can actually *see that* at a glance.

PR #1365, opened 2026-04-16
[DNM] Temporary disable of observer

Useroth:
For testing purposes regarding the recent lag and crashes.


coffeecoffeecoffee:
So, while there's supposedly the issue of observer ghosts keeping NPCs AI 'active' or something, this definitely hasn't solved the crashes...

zoktiik:
> So, while there's supposedly the issue of observer ghosts keeping NPCs AI 'active' or something, this definitely hasn't solved the crashes...

Well no, of course not. We haven't disabled adminghosts, and people who die can still ghost. But I think it's more or less an attempt to see if reducing the number of ghosts affects it, which it may or may not. Uncovering the source of "the stink" (the crash bug) is top-priority.

PR #1367, opened 2026-04-17
Low Pop only BYOS and VIRT-VICE Test

Guidesu:
## About The Pull Request
It will be expanded more upon after i do some test and gather feedback. The work on this PR is old and not updated and i learned much more things since i worked upon it, but i need people to test it. So later on i will write a more detailed changelog for everything. 

Don't EVEN think of TMing in the main server.

## Changelog

:cl:
add: Added more things
del: Removed old things
qol: made something easier to use
balance: rebalanced something
fix: fixed a few things
map: added/modified/removed map content
spellcheck: fixed a few typos
code: changed some code
refactor: refactored some code
admin: messed with admin stuff
server: something server ops should know
/:cl:


Jamdrawers:
Not sure how much of this is intended to go in and how much of this left for just testing purposes but it's still pretty crazy generous. You can start with a LOT of skills.

AuziaGitH:
the less than 4 starting skills being apprentice are fine imo. most of the skills you can grind to there in maybe 5-10 minutes. 

but I think the main issue with the feat system are their low & inconsistent cost. especially the combat ones. 
one of the worst thing this brings are the possibility to take BOTH duel wielder and natural armor. without sacrificing stat pack for virtuous. 

feats like well off (atomized nobility), wild born should atleast be 5 points cost
feats like natural armor, equestrian, that are inherently extremely beneficial in combat should always be 10 points to prevent people from taking more than 1 without sacrificing statpack.

code/modules/mob/living/carbon/human/life.dm
```diff
@@ -71,13 +71,13 @@
 	update_stamina()
 	update_energy()
 	
-	// Process all vices
+	// Process vices that need life ticks (performance optimization)
 	if(mind && length(vices))
 		for(var/datum/charflaw/vice in vices)
-			if(!vice.ephemeral)
+			if(!vice.ephemeral && vice.needs_life_tick)
 				vice.flaw_on_life(src)
```
zoktiik, line 78: flaw_on_life just does a basic `return` if no other behavior overrides it. Is calling a proc that instantly returns really all that performance heavy? If not, then checking a variable is just going to add on unnecessary load.

Jamdrawers:
> the less than 4 starting skills being apprentice are fine imo. most of the skills you can grind to there in maybe 5-10 minutes.

Gotta factor in the two or three other virtues that also granted skills

PR #1376, opened 2026-04-18
Removes Circuitus from Learnable Spells.

Kroto-Ker:
## About The Pull Request

Removes circuitus from learnable spell list.

Apologies for pushing this in the first place, you guys warned me it would cause issues and you were right!
I've learned my lesson.

## Testing Evidence

It compiled!

## Why It's Good For The Game

Requested en masse in feedback, the feature is too imbalanced and doesn't fit the game.

## Changelog

:cl:
balance: Removes circuitus from learnable spells.
/:cl:

### DIFF
--- a/code/modules/spells/spell_types/wizard/spell_list.dm
+++ b/code/modules/spells/spell_types/wizard/spell_list.dm
@@ -60,6 +60,5 @@ GLOBAL_LIST_INIT(learnable_spells, (list(/obj/effect/proc_holder/spell/invoked/p
 		/obj/effect/proc_holder/spell/invoked/silence,
 		/obj/effect/proc_holder/spell/self/recall,
 		/obj/effect/proc_holder/spell/self/findfamiliar,
-		/obj/effect/proc_holder/spell/invoked/incantation,
 		)
 ))


ChickyRooster:
Perhaps we can think about getting the Azure Peak system for magic instead, as it IS rather interesting.

Glad you at least realise this. Whilst the spell itself is fun and very interesting, it was certainly causing a lot of troubles.

Rin053:
are u kitten me

wikimody:
On one hand, circuits are much more powerful then other tools mages have acces to, and the system design pushes to make 5 words nukes instead of more interesting spells.

On the other hand, circuit make you feel like an actual mage, thinking and researching your spells, with incredible pay offs as you understand language of arcane more and more.

I think that while circuits need adjustment (particularly one tile nuke with I think is main thing non mages are mad about) and their design should be rethought, removing them from mages will remove a lot from the feel of the class, making it again much more like archer side grade instead of wielders of arcane force

wikimody:
I think circuits should be adjusted, not removed. The idea behind them was excellent, and I think for all the issues the execution has, none of them are unfixable

Jamdrawers:
It could be left as a court magos exclusive at least!

ceeetchell:
laaaaaame

PR #1391, opened 2026-04-19
Bugfix - Remove specific ways to not bleed

Rayforx:
## About The Pull Request
If you were cold enough, you stopped bleeding because of some relic SS13 code.
Instead, I made it into a feature that makes you bleed half as much if you are freezing (like a popsicle).

## Testing Evidence
image

## Why It's Good For The Game
Bugs are pretty bad, even more so when they are game-breaking.

## Changelog
:cl:
fix: Cold weather removes the ability to bleed.
balance: Makes extremely cold weather slow down bleeding by 50% instead of 100%.
/:cl:

### DIFF
--- a/code/modules/mob/living/blood.dm
+++ b/code/modules/mob/living/blood.dm
@@ -15,7 +15,7 @@
 		to_chat(src, span_warning("The blood soaks through my bandage."))
 
 /mob/living/carbon/monkey/handle_blood()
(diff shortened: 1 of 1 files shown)


zoktiik:
Changelog should say something like "fix: Fixed bleeding being stopped entirely if the character is freezing. Now it will slow bleeding by 50% instead."

Good PR. :+1:

PR #1400, opened 2026-04-20
adds secondary checks to most bedbreaker checks that allow particularly strong lovers to also break beds

zeroisthebiggay:
## About The Pull Request

read what the title says, it means anyone above twelve strength can switch to strong intent and do mostly everything the bedbreaker trait allows you to do
i also added a strong intent check to breaking pelvises so you don't accidentally guro someone even though people typically consider that genre of bone breaking to be safe horny

## Testing Evidence

i promise you i spawned in a guest account and fucked them with a debug message to verify it's working but i didn't test it out with every single interaction okay

image

## Why It's Good For The Game

i love fetish code

## Changelog

:cl:
add: anyone above twelve strength can use strong intent while having sex to act as if they have the bedbreaker trait
/:cl:

### DIFF
--- a/code/datums/sexcon/sex_actions/sex/anal_ride_sex.dm
+++ b/code/datums/sexcon/sex_actions/sex/anal_ride_sex.dm
@@ -33,8 +33,9 @@
 	user.sexcon.intercourse_noise(user)
 	user.sexcon.do_thrust_animate(target)
 
-	if(HAS_TRAIT(user, TRAIT_DEATHBYSNUSNU))
-		user.sexcon.try_pelvis_crush(target)
+	if(HAS_TRAIT(user, TRAIT_DEATHBYSNUSNU) || (user.STASTR > 12))
+		if(istype(user.rmb_intent, /datum/rmb_intent/strong))
+			user.sexcon.try_pelvis_crush(target)
 
 	target.sexcon.perform_sex_action(target, 2, 0, TRUE)
 	if(target.sexcon.check_active_ejaculation())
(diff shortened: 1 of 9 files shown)


josephpatrickgreen:
When I made this trait I did not intend the entire retinue to be able to shatter pelvises at will through sex. It was meant for werewolves and savages. Where I do think it would be funny for the knights to accidentally demolish their squires and I am not anti-fun. I'd raise the strength cap to 14 and also require a strong grab if you are dead set on it.

coffeecoffeecoffee:
oh
yeah 12 is... not very much at all
make it 15+?

zeroisthebiggay:
> oh yeah 12 is... not very much at all make it 15+?

it's greater than twelve so thirteen minimum
my logic here is that you'd need to at least be a subclass with plus one strength `and` a double strength statpack, or be a strong race on a strength neutral subclass with a double strength statpack

the amount of people who can reach fifteen strength is..........
extremely small and mostly limited to things like antags

zeroisthebiggay:
as a comprimise i could increase the pelvis breaking checks to fourteen or fifteen
but i'm adamant that the extra pain checks should stay at thirteen because it's fairly inconsequential i just want to plap people harder when i'm playing knight

HeroWithYay:
>I'd raise the strength cap to 14 and also require a strong grab if you are dead set on it.

Balance-jakking sex mechanics :sob:

Xander3359:
This makes the trait a lot less unique to have, but whatever I guess, it's just fetish content regardless.

PR #1411, opened 2026-04-22
Make deadite animals controllable with command undead

gatorceres:
## About The Pull Request

Make it so deadite animals (saigas and volfs) can be commanded, similar to simple skeleton NPCs, via command undead. They can currently be tamed by necromancers and manhandled to get them to other places, but don't react to orders, and volfs in particular tend to wander around until they blunder into enough kneestingers to re-kill themselves. 

## Testing Evidence

After making these changes I spawned in some deadite animals and skeletons and confirmed they can be tamed and commanded as normal. 

Deadite saigas have weird pathfinding on occasion, similar to what happens if your living saiga mount retaliates to something attacking it and gets hard to control, but it's not a serious hindrance to using them.  

## Why It's Good For The Game

More variety for the necromancer hordes. Volfs and saigas are admittedly harder opponents than the summons, but also require more effort to get. 

## Changelog

:cl:
add: Deadite saigas and volfs may be commanded by necromancers
/:cl:

### DIFF
--- a/code/modules/spells/roguetown/necromancer.dm
+++ b/code/modules/spells/roguetown/necromancer.dm
@@ -214,7 +214,7 @@
 /obj/effect/proc_holder/spell/invoked/tame_undead
 	name = "Tame Undead"
 	desc = "Oftentymes, husks and shamblers walk aimlessly - uncertain of their future. Befriends the undead \
-	Requires the target to be within four tiles. Works on undead animals, too."
+	Requires the target to be within four tiles. Works on undead animals, too, and they will heed your command."
 	overlay_state = "wolf_head_undead"
 	range = 4
 	warnie = "sydwarning"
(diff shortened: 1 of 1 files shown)


The-Sun-In-Splendour:
Good PR. I felt so robbed when I tamed a skeleton and found out I couldn't use it in battle.

gatorceres:
> Good PR. I felt so robbed when I tamed a skeleton and found out I couldn't use it in battle.

This PR doesn't change how command works with skeletons, just deadite animals. I was testing with simple mob skeletons and conjured skeletons to make sure I didn't break the existing control logic for them. 

Being able to command any tamed skeleton is also on my list, but I wasn't yet sure how to generalize the behavior to all NPC skeletons instead of just the conjured ones. I'll take another crack at it later, when I have more time to learn the existing code.

PR #1413, opened 2026-04-22
Increase in cooldown delay for player controlled summons

RadiantFlash7:
## About The Pull Request
This PR adds increased ranged cooldown delay for specific summoned mobs when they become player controlled.

## Why It's Good For The Game
Back when I made the whole summon gameplay system, binding mobs was fairly uncommon. Either for player culture reasons, or otherwise, Worrying about player controlled T3 and T4 mobs wasn't a big concern for a long while. The AI for the mobs is not particularly smart, so having a short cooldown is somewhat needed. However, in player hands it can become oppressive.
The changes now should help resolve the issue.

## Changelog

:cl:
balance: Player controlled Fiends, sylphs and watchers have 1.5-2x more ranged cooldown delay
/:cl:


Vizoplez:
can we fix summons having 0 FPS before trying to nerf them further

PR #1423, opened 2026-04-23
Deletes Circuitus.

Xander3359:
## About The Pull Request
Deletes Circuitus.

## Testing Evidence
Compiled on my machine, circuitus no longer ingame.

## Why It's Good For The Game
Quite frankly this thing is a balance nightmare. It's either some overpowered gigabullshit that oneshots everyone in a 3x3 combined with some stun bullshit to guarantee it. Or it's a worthless spell for 4 spell points that sucks and does nothing.

Just like normal ss13 circuits, you either know the meta copypaste oneshot string or you're a noob who does nothing.
Sure there is some fun experimentation and some cute things you can build by learning it and putting in the effort to experiment, but considering how admins consider this spell powergaming, it's best to just be rid of it and save the headache.

## Changelog

:cl:
del: Removes Circuitus
/:cl:


MickDick1:
If the admins punish you for using a feature, you remove the feature. It's good practice.

john-pawson:
Think a better solution may be just making it unobtainable or only work with the scroll system (So you need to put in the work to cast them, at the very least) rather than just outright deleting everything. 

It's a fun system that can make magic casting way more interactive and flexible, it IS broken as of right now, but between it being deleted and likely forgotten to be never readded or being left on a WAY more underpowered state until it can get reworked properly on a way that's more fun... latter seems better, doesn't it?

Xander3359:
> Think a better solution may be just making it unobtainable or only work with the scroll system (So you need to put in the work to cast them, at the very least) rather than just outright deleting everything.
> 
> It's a fun system that can make magic casting way more interactive and flexible, it IS broken as of right now, but between it being deleted and likely forgotten to be never readded or being left on a WAY more underpowered state until it can get reworked properly on a way that's more fun... latter seems better, doesn't it?

The issue is that it's seen as powergaming by the admins and I would rather not leave ban bait in the game.
It's unbalanced as shit and Krotoker has burnt out due to the negative feedback over circuitus that he's given up on adjusting it so it's fair.

Thus it's better overall to just be rid of it, it was fun while it lasted and an amazing concept but it unfortunately did not pan out well in practice.

Rin053:
> The issue is that it's seen as powergaming by the admins and I would rather not leave ban bait in the game. It's unbalanced as shit and Krotoker has burnt out due to the negative feedback over circuitus that he's given up on adjusting it so it's fair.
> 
> Thus it's better overall to just be rid of it, it was fun while it lasted and an amazing concept but it unfortunately did not pan out well in practice.

user was circuitus nuked for this post

wikimody:
As i said before, i think circuits are an amazing idea and one that's execution can be fixed from the issue they have. I dont think deleting them is a good idea

PR #1436, opened 2026-04-25
Gives Miraclist Zizites grave mark

HalfAwake691:
## About The Pull Request

Another skeleton summoner that needs the ability to control their summons, Many such cases. 

## Testing Evidence

Its a web edit sire 

## Why It's Good For The Game

I think necromancers having the ability to actually use their summons in a way that is not destructive to their teammates is a good idea. Every Necromancer class should get this.

## Changelog

Adds Gravemark and order minions to Miraclist Dunewell nomad


The-Sun-In-Splendour:
Why don't you just make gravemark bundled into the miracle tier that gives you raise undead?

PR #1440, opened 2026-04-25
Accent Previews + Saut-Al-Atash Accent Fix

yupup123:
## About The Pull Request

- Adds some gibberish preview text to show the player what an accent looks like when you select it.
- Fixes Saut al-Atash accent to actually do something. Desert Town enjoyers rejoice

## Testing Evidence

image

## Why It's Good For The Game

Lots of players end up joining only to find out the accent they picked is NOT what they expected. Now you can see kind of what it might look like.

## Changelog

:cl:
add: Accent Preview Text when Selecting Accents
fix: Fixes Saut-Al-Atash Accent

/:cl:


roguetown.dme
```diff
@@ -8,6 +8,7 @@
 // END_PREFERENCES
 // BEGIN_INCLUDE
 #include "_maps\_basemap.dm"
+#include "_maps\roguetest.dm"
```
Xander3359: what?

Xander3359 (changes requested):

roguetown.dme
yupup123: Oops. Was using that for testing. Removing it now.

Rin053:
kazen also used to have its own font accent I believe talking normally as well

Taliazure:
Can you make it so the desert accent replaces the words to the ones from the roles? Like duke being sultan and etc?

PR #1455, opened 2026-04-26
You get +3 spellpoints! And YOU get +3 spell points! Arcyne Potential is no longer restricted, but...

NotGoldenglow:
## About The Pull Request

This PR removes the restriction from Arcyne Potential.
Previously, if you had DODGE EXPERT, MEDIUM/HEAVY ARMOR, CRIT RESISTANCE or ANY OF THE ANTI-PAIN TRAITS? You could not receive the T1 Arcyne access given by this trait, only Prestidigitation. Now, anyone has access to the limited, flavored spell list... With a caveat.

**INVISIBILITY** and **LEAP** are now T2 spells, meaning only spellblades/mages+ can access them.
Why? Because they're two extremely powerful mobility spells, and I don't think they should be given out freely.

Also, if you were on the list of people who would previously not qualify (by not being an arcane user as-is), then you only get 2 points instead of 3. 
Why? So you can pick between wasting an entire virtue on a SINGLE "powerful" spell like darkvision or longstrider, OR two flavorful spells like mirror transform, campfire, etc...

## Testing Evidence

spell_tier = 1 -> 2
3 line change total
here's your local Valiant Warrior duke and his mighty armor training with spellpoints attached.
image

## Why It's Good For The Game

Flavor. Let people take mirror transform or message or whatever the hell else if they want to, they're giving up an entire trait for 1, maybe 2 spells at best.

## Changelog

:cl:
balance: arcyne potential is no longer limited by crit/dodge/pain trait/armor proficiency. if you previously did not qualify for this trait, you now do.
/:cl:

### DIFF
--- a/code/modules/spells/roguetown/acolyte/noc.dm
+++ b/code/modules/spells/roguetown/acolyte/noc.dm
@@ -48,7 +48,7 @@
 	range = 3
 	warnie = "sydwarning"
 	movement_interrupt = FALSE
-	spell_tier = 1
+	spell_tier = 2
 	invocation_type = "none"
 	sound = 'sound/misc/area.ogg' //This sound doesnt play for some reason. Fix me.
 	associated_skill = /datum/skill/magic/arcane
--- a/code/modules/spells/spell_types/wizard/buffs_debuffs/leap.dm
+++ b/code/modules/spells/spell_types/wizard/buffs_debuffs/leap.dm
@@ -10,7 +10,7 @@
 	no_early_release = TRUE
 	movement_interrupt = FALSE
 	gesture_required = TRUE // Mobility spell
-	spell_tier = 1
+	spell_tier = 2
 	invocations = list("Saltus!")
 	invocation_type = "whisper"
 	hide_charge_effect = TRUE
--- a/modular_azurepeak/virtues/combat.dm
+++ b/modular_azurepeak/virtues/combat.dm
@@ -2,16 +2,15 @@
 /datum/virtue/combat/magical_potential
 	name = "Arcyne Potential"
 	desc = "I am talented in the Arcyne arts, expanding my capacity for magic. I have become more intelligent from its studies. Other effects depends on what training I chose to focus on at a later age."
-	custom_text = "Classes that has a combat trait (Medium / Heavy Armor Training, Dodge Expert or Critical Resistance) get only prestidigitation. Everyone else get +3 spellpoints and T1 Arcyne Potential if they don't have any Arcyne."
+	custom_text = "You gain +3 spellpoints and T1 Arcyne Potential, which comes with a limited, mostly flavor and non-combat spell list. If you already have skills with magic, you gain +1 level and +3 spell points."
 	added_skills = list(list(/datum/skill/magic/arcane, 1, 6))
 
 /datum/virtue/combat/magical_potential/apply_to_human(mob/living/carbon/human/recipient)
 	if (!recipient.get_skill_level(/datum/skill/magic/arcane)) // we can do this because apply_to is always called first
 		if (!recipient.mind?.has_spell(/obj/effect/proc_holder/spell/targeted/touch/prestidigitation))
 			recipient.mind?.AddSpell(new /obj/effect/proc_holder/spell/targeted/touch/prestidigitation)
-		if (!HAS_TRAIT(recipient, TRAIT_MEDIUMARMOR) && !HAS_TRAIT(recipient, TRAIT_HEAVYARMOR) && !HAS_TRAIT(recipient, TRAIT_DODGEEXPERT) && !HAS_TRAIT(recipient, TRAIT_CRITICAL_RESISTANCE))
-			ADD_TRAIT(recipient, TRAIT_ARCYNE_T1, TRAIT_GENERIC)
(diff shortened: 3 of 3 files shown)


NotGoldenglow:
someone smarter than me can make sure i didn't break anything, i'm going to bed

ariaworld:
mending for free makes failed squire useless

UntoldTactics:
> mending for free makes failed squire useless

You need journeyman arcyne skill to make use of mending to fix up to full(with the repair changes TM'ed/merged), but that being said, you can create teaching books that teach up to journeyman arcyne (iirc)

The fact it'd give whoever wants it free mending potential tho is a bit of a concern.

anconfuzedrock:
failed squire is more useful for the polish than the repairs.

john-pawson:
Same applies to darkvision, and longstrider. Not really "flavor" spells. Doesn't help that they can change between those if they simply bother crafting a tome. That and message *may* affect balance greatly when it becomes a standard to get used by guards and/or wardens as a replacement for SCOMs.

anconfuzedrock:
message is t1 again??? yeah that's crazy good, crownstones for the whole garrison. not a good idea.

RoseBrando:
yeah if message weren't 1 cost i'd say just make it give 1 spellpoint, esp knowing the intent behind this pr is mostly for flavor stuff i.e. mirror transform _(trust me, i love mirror tf and i genuinely have decided to not play classes cause i cant access arcyne potential with them and thus cant access mirror tf)_

unfortunately right now message is 1 cost and that's very strong to give out even for a dedicated virtue slot, i think the best thing to do - and i have very little code making knowledge, only reading - is something like devotee where you could split the current  t1 spells into a sort of "t0" and "t1" similar to miracles, and at that point putting the mostly flavor or "useless" t1 spells like mirror tf, light, nondetection and maybe campfire into that, giving combat classes access to t0 and non-combat to t1. 
of course the alternative of making message not 1 spell-point and just giving combat classes 1 spell-point for the aforementioned spells is also there but that was apparently approved by a maintainer 🙃

good concept and i back the idea of it, just feel like theres a better way than removing the cap entirely

royalbeehavior-arch:
I think we're being a bit unserious if we believe that garrison members will all run arcyne potential instead of other things like natural armor, combat aware, and intellectual. Some might, but it comes at the very real opportunity cost of more useful combat items.

NotGoldenglow:
> Same applies to darkvision, and longstrider. Not really "flavor" spells. Doesn't help that they can change between those if they simply bother crafting a tome. That and message _may_ affect balance greatly when it becomes a standard to get used by guards and/or wardens as a replacement for SCOMs.

They can change between them once a day, after they've slept, so it's not like they can pingpong back and forth.
It's specifically still a virtue slot. If someone wants to fuck themselves out of the 100 other useful things that can be granted there, that's their choice. There's better dark vision in the form of night-eyed, and better mending in the form of failed squire.

If longstrider is a concern, I can further limit this list or drop the spell points given to 1 **if you do not have arcyne skill**, while maintaining the +3 for existing spellcasters who DO take this virtue for that reason.
You have a choice of flavor, and that's about it.

(Let's not forget that darkvision/longstrider are enchantments that can be provided in five minutes if they ask any mage, both in the form of the actual spell being applied OR item enchantments given out that'll last essentially, permanently.)

ophaq:
I was thinking of doing something similar but adding more spells that wouldn't be OP and adding exclusions to things like mending / etc that would be too OP for classes with DE / etc.

My suggestion is to remove mending and other spells from the current list of pickable ones for balance purposes if they have DE/etc. For example: Light, mirror magic, featherfall, darkvision would be fine for armor training/DE classes. Mending, longstrider, message and campfire? Probably not.

VIRTUALCRIME:
i do not think this is bad, majority of low level spells are relatively useless outside of optimizing things going

maker-of-nightmares:
All the issue spells listed are basically just the most minor of buffs you can get via other shit already in the game for no cost. Want Darkvision? Be a Zizoid or Nocian boom free darksight or Nightsight. Want Longstrider? literally a dozen ways including carrying one of the better sets of weapons in game(the polearm) something you are already inclined to do. Mending? Nerfed into the fucking ground so no funni combat mend unless you are so fucking awful at combat you let him get a 4 second repair break in which case you deserve it.

Message is literally the only problematic one and it's not even that big a deal if they don't have skill in arcane. Oh no they'll get 1 message out every few minutes! Egads thats... exactly the same as a Houndstone with none of the benefits... The only benefit it has over the houndstone is that you can tell your boss something you see to relay down the line from anywhere. This is not that potent a power since the vast majority of people who care to do that have a houndstone and are near an SCOM or Hermes which can do the same. You end up relying on someone else to be at their keyboard to see the message(or in a state to see it) and thus run the risk of no one hearing what you said. If it's so oppressive a better option would be that peasants and lower tier guards(like wardens and watchmen) don't know who the nobles are in their known person list. This means they have to physically walk up to them and add them to the list by talking with them. This also isn't out of line for bumfuck no ones to not exactly personally know the nobles well enough to beam a message into their head. Knowing the Duke is Sir Dicksalot doesn't mean you _know_ the Duke enough to be able to accurately message them.

PR #1458, opened 2026-04-26
Splint/Brigandine Sound... Buff?

KeberosNihilus:
## About The Pull Request

Changes the sound of the brigandine and split leggings to be less obnoxious. Technically a buff?

## Testing Evidence

So like it's a one-line change and I don't have video recording software sooooooo

## Why It's Good For The Game

Split/Brigandine leggings are just... Weirdly loud for how they're sprited? They're _tight_ leather pants with bits of metal strapped to your leg. It's not a coat of plates, or a set of steel plates rubbing against one another. They _really_ shouldn't make as much noise as they did.

I mean, they still _do_ make noise. Just the same as a breastplate, now. 

## Changelog

:cl:
sound: Changed Brigandine and Splint leggings from using the Brigandine Fullplate SFX to using the standard Plate SFX.
/:cl:

### DIFF
--- a/code/modules/clothing/rogueclothes/pants/chain.dm
+++ b/code/modules/clothing/rogueclothes/pants/chain.dm
@@ -47,7 +47,7 @@
 
 /obj/item/clothing/under/roguetown/splintlegs/Initialize(mapload)
 	. = ..()
-	AddComponent(/datum/component/item_equipped_movement_rustle, SFX_PLATE_COAT_STEP)
+	AddComponent(/datum/component/item_equipped_movement_rustle, SFX_PLATE_STEP)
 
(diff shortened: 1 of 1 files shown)


KeberosNihilus:
> Cant you just upload the sounds itself for a preview?

I didn't even think of that, honestly. Gimme a minute to dig up the oggs I guess.

KeberosNihilus:
So apparently Github doesn't support .oggs as attachments?

The filepath is sound -> foley -> footsteps armor -> coatplates (1), (2), (3) and trading that for plate (1), (2), (3)

Still a bit noisier than chain, but it's also a lot less obnoxious than Brigandine. Also makes more sense, imo, for the reasons listed above.

PR #1459, opened 2026-04-26
Adds Tailor's Silverface Imports Tab

KeberosNihilus:
## About The Pull Request

Adds an Exotic Light Armor tab to the tailor's silverface. Even have a fancy little comment in the code so people have a quick idea on how to add to it.

## Testing Evidence
proof03
proof01
proof02

## Why It's Good For The Game

Doesn't really affect the balance much? A lot of these items are nigh impossible to obtain either not having crafting recipes for whatever reason, or requiring stuff like master / legendary tailoring. I guess there's an argument it takes away from mercs, but like... We have _eight_ merc slots on a server that regularly sees 100+ players in a round. Plus it should help Light AC users have some actual cosmetic variance when there isn't a tailor around... And _not_ all look either like generic bandits or Grenzels.

Probably didn't add everything I should / could, but the way armors are sorted is... Unintuitive? At least to me. Plus with the foundation done, it should be fairly easy to expand in the future.

I'll admit I wasn't _super_ thorough adding stuff, but I was sure to not add things that seemed super unique (eg- "cut-throat's pants") or had powerful unique functions (a pair of Gronn gauntlets that look _really_ nice... but also have a 1.25x unarmed damage mod. Sadness) so hopefully it doesn't step too hard on certain classes' unique visual identities.

I mean, it _will_ at least a little, but I've heard a LOT of people beg for more armor variety so like... I feel like the increase in player fashion will outweigh the loss of unique identity. Needs of the many and all that.

## Changelog

:cl:
add: Added an Exotic Light Armor tab to the tailor's silverface.
add: Filled the Exotic Light Armor tab with a bunch of exotic light armors. Who could've seen that coming?
balance: Balanced(?) the exotic armors by making them cost slightly more than local light armors.
/:cl:

### DIFF
--- a/code/modules/cargo/packsrogue/merchant/armor/merch_armor_light.dm
+++ b/code/modules/cargo/packsrogue/merchant/armor/merch_armor_light.dm
@@ -67,3 +67,98 @@
 	contains = list(
 					/obj/item/clothing/head/roguetown/roguehood/reinforced,
 				)
+
+// Exotic import stuff goes here. Should probably be a little pricier than normal stuff. 2x average? Be sure to name the purchase option so it relates to the actual item, but also what slot it fills.
+
+/datum/supply_pack/rogue/light_armor/import
+	group = "Imported Armor (Light)"
+
+/datum/supply_pack/rogue/light_armor/import/otavangambeson
+	name = "Otavan Fencing Gambeson"
+	cost = 60 // Base sellprice of 30
+	contains = list (/obj/item/clothing/suit/roguetown/armor/gambeson/heavy/otavan)
+
+/datum/supply_pack/rogue/light_armor/import/otavanpants1
+	name = "Otavan Heavy Leather Trousers"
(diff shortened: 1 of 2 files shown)


ReynTheLord:
Some of these can't even be CRAFTED by the tailor, why would you just slap them in the silverface?

KeberosNihilus:
> Some of these can't even be CRAFTED by the tailor, why would you just slap them in the silverface?

That's part of why. The other part is I'm going to _make_ them craftable. Eventually. Like, this isn't really valid argument because we have imported _weapons._ And those are often either upgrades or sidegrades. These are - to my knowledge - mostly reskins of Hardened Leather Trousers and Padded Gambesons.

Also, as said in the OP, people have been begging for these on and off a LOT. Partly because there's never a tailor around in the first place.

Xander3359:
It's probably fine. Should be cheaper to craft (same as with blacksmith silverface)

PR #1460, opened 2026-04-27
More stuff for the bathmaster to buy at the bathmaster shop (and some map tweaks)

Jamdrawers:
## About The Pull Request

A few more map fixes, tweaks and lighting passes, decluttering some of the nightmaster's things in their stash, and a sweep of things added to the nightmaster's buy menu.

## Testing Evidence

image

## Why It's Good For The Game

Figure these things should've been here to begin with - also means I can declutter their stash a little and they can simply use some of their mighty wealth to buy some of their things instead.

## Changelog
:cl:
add: Added chastity cages, dildos and more to the nightmaster's buy-menu
/:cl:


maker-of-nightmares:
Did you make sure to ensure the area wasn't fucked with so they still get their passive dragon horde money? The file for it is too big for my internet to handle RN

code/modules/cargo/packsrogue/bathmatron/drugs.dm
```diff
@@ -88,3 +87,13 @@
...
+
+/datum/supply_pack/rogue/drugs/shhig
+	name = "Shhig brand zigs (pack of 6)"
+	cost = 40
+	contains = list(/obj/item/storage/fancy/shhig)
+
+/datum/supply_pack/rogue/drugs/suidust
+	name = "Dust of Disguise (Gender only)"
+	cost = 135
+	contains = list(/obj/item/alch/transisdust)
```
Xander3359, line 99: This doesn't actually do anything to your character model at all.
And I'm pretty sure you can just buy sui from the vendor in the whitevein for way cheaper than this.

Jamdrawers:
> Did you make sure to ensure the area wasn't fucked with so they still get their passive dragon horde money? The file for it is too big for my internet to handle RN

Yeah I figured out what was happening with passive money a while ago - it's quite goofy. Instead of checking for the areatype, it checks for specifically that particular type of bath-tile within a ~7 block radius of the brassface structure. To make visual variation, the different tiles surrounding it are actually reskinned instances of the original specific bath tile type.

code/modules/cargo/packsrogue/bathmatron/drugs.dm
Jamdrawers: It does work on your character model, actually! You just have to use it in-hand and say "yes" to "would you like to change your frame. This character had a masculine bodytype before using it 
image
I'll check the prices to try and equalise them though

Jamdrawers: Seems the price at the whitevein drugmachine is a random amount between 80 and 120 while it's 135 at the merchant. I'll nudge it down for the nightmaster since they should be getting discounts

PR #1461, opened 2026-04-27
ports hunger changes

Jamdrawers:
## About The Pull Request

See https://github.com/Azure-Peak/Azure-Peak/pull/6655

## Testing Evidence

Compiles, loads, seems fine

## Why It's Good For The Game

see https://github.com/Azure-Peak/Azure-Peak/pull/6655

## Changelog

:cl: WeNeedMorePhoron, ported by Jamdrawers
balance: Removed most stats and athletics skills influence on hunger rate and then halved the flat hunger rate gain. Report if hunger happens too often. Str above 10 increases hunger rate by 5% per point instead of 7.5%
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/energystamina.dm
+++ b/code/modules/mob/living/carbon/energystamina.dm
@@ -87,6 +87,7 @@
 		nutrition_amount *= 2
 	if (STASTR >= 11) // 5% increased nutrition loss for every STR above 11. the gainz don't come cheap
 		var/swole_malus = (10 - STASTR) * 0.05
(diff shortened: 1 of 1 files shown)


maker-of-nightmares:
Can we not make Athletics absolutely useless like they have?

kimesik:
> Can we not make Athletics absolutely useless like they have?

Please make Athletics absolutely useless.

In fact, please cut out Athletics entirely.

ariaworld:
good thing: athletics is actually useful not only for energy but also for stamina

PR #1466, opened 2026-04-27
Ports Leather Padded Hood from Scarlet Reach

Fn-OS:
## About The Pull Request

Ports a new clothing/armor item from Scarlet Reach, adding it to the loadout and crafting menu.

## Testing Evidence

It works! 
Screenshot (1786)
Screenshot (1783)
Screenshot (1784)

## Why It's Good For The Game

New drip and a new light armor item

## Changelog

Ports Leather Padded Hood and Studded Leather Hood
Adds new clothing item to loadout
Removes wearing the Leather Padded Hood on the neck slot (Headwear only.)

:cl:
add: Added Leather Padded Hood and Studded Leather hood, Studded Leather Hood can be found in loadout.
balanced: Leather Padded Hoods cannot be worn on the neck slot.
/:cl:


Guidesu:
Good First contribution keep up the good work!

Fn-OS:
Linters fixed

zeroisthebiggay:
i think the hood should be available in the tailor's silverface instead of in the loadout
taking it in loadout and then going to the smith to get it studded sounds strong

ophaq:
Finally, padded hoods. Also, yes. Add the studded variant to the smith silverface, perhaps. And unstudded to the tailor silverface. We don't need it in the loadout.

Fn-OS:
if I can figure out how to, sure.

john-pawson:
>Ports Leather Padded Hood and Studded Leather Hood
>Adds new clothing item to loadout
>Removes wearing the Leather Padded Hood on the neck slot (Headwear only.)
>
>🆑
>add: Added more things
>image: added/modified/removed some icons or images
>code: changed some code
>/:cl:

You need to format it like this:

🆑
add: Added leather padded hood, only wearable on the head slot, can be found on the loadout
add: Added crafting recipe for leather padded hoods
balance: Leather padded hoods can only be worn on the head slot
/:cl:

Or else the changelog will just say "Added more things"

code/modules/roguetown/roguecrafting/leather/leather_armor_recipes.dm
```diff
@@ -142,3 +142,8 @@
 	result = /obj/item/clothing/head/roguetown/helmet/leather
 	reqs = list(/obj/item/natural/hide/cured = 1)
 	craftdiff = 2
+/datum/crafting_recipe/roguetown/leather/studhood
+    name = "leather padded hood"
+    result = /obj/item/clothing/head/roguetown/helmet/leather/armorhood
+    reqs = list(/obj/item/natural/hide/cured = 2)
+    craftdiff = 2
```
john-pawson: Should likely have a higher craftdiff, considering how it has better protections than normal leather AND also protects your entire face + neck.

john-pawson:
Maaaybe consider changing the sprite a bit if possible? So the two hoods don't use the same sprite? It's not that big of a deal.

Other than those three things, great PR! Love it. We needed more non-plate helmets.

Fn-OS:
> Maaaybe consider changing the sprite a bit if possible? So the two hoods don't use the same sprite? It's not that big of a deal.
> 
> 
> 
> Other than those three things, great PR! Love it. We needed more non-plate helmets.

I am not ar ALLL a spriter so that'd be something to be done at another time, but it is something I'd like to do
I'll fix the PR and such, code will probably be fixed tomorrow since I'm headed to bed now.

code/modules/roguetown/roguecrafting/leather/leather_armor_recipes.dm
Fn-OS: How much higher do you think?  Would 4 work?

Jamdrawers:
These hoods seem to be balanced strangely - instead of having protection values similar to metal armour, leather armour or studded leather armour it's got a unique set that's like kind of the best of all of them.

Normally when we give armor protection values, instead of listing out the protections of each individual type in order, we define those somewhere else and just use a term that refers to that definition - it'd be more in line to instead list 'armor = ARMOR_LEATHER_STUDDED' - it'd probably also be balanced a little more normally that way.

Fn-OS:
> These hoods seem to be balanced strangely - instead of having protection values similar to metal armour, leather armour or studded leather armour it's got a unique set that's like kind of the best of all of them.
> 
> Normally when we give armor protection values, instead of listing out the protections of each individual type in order, we define those somewhere else and just use a term that refers to that definition - it'd be more in line to instead list 'armor = ARMOR_LEATHER_STUDDED' - it'd probably also be balanced a little more normally that way.

I'm a little bit iffy on making them exactly the same as studded leather armor because then it just becomes a reskin? I believe everything else has been fixed now though.
I replaced the loadout version of the hood with the studded leather hood so that people can't get it upgraded at the smith.

Fn-OS:
> Maaaybe consider changing the sprite a bit if possible? So the two hoods don't use the same sprite? It's not that big of a deal.
> 
> Other than those three things, great PR! Love it. We needed more non-plate helmets.

Problems should be fixed other than the sprite, which I will work on, though it might take some time. Should be ready to test/merge.

john-pawson:
> I'm a little bit iffy on making them exactly the same as studded leather armor because then it just becomes a reskin?

A reskin of what? We currently don't have any studded leather armor pieces on the head slot, and the hood blocks some vision in exchange for face protection*, which means that the hood would still be a unique piece of headwear compared to the normal leather helmets.

* : After checking, turns out, the hood actually doesn't block some of your vision! It probably should, considering how all face-covering headgear does so, even regular reinforced hoods, which are worse.

Fn-OS:
> > I'm a little bit iffy on making them exactly the same as studded leather armor because then it just becomes a reskin?
> 
> 
> 
> A reskin of what? We currently don't have any studded leather armor pieces on the head slot, and the hood blocks some vision in exchange for face protection*, which means that the hood would still be a unique piece of headwear compared to the normal leather helmets.
> 
> 
> 
> * : After checking, turns out, the hood actually doesn't block some of your vision! It probably should, considering how all face-covering headgear does so, even regular reinforced hoods, which are worse.

I completely forgot we dont have any studded headwear yeah I can change it
If I can figure out how to give it an FOV I'll do that too

Fn-OS:
> > I'm a little bit iffy on making them exactly the same as studded leather armor because then it just becomes a reskin?
> 
> 
> 
> A reskin of what? We currently don't have any studded leather armor pieces on the head slot, and the hood blocks some vision in exchange for face protection*, which means that the hood would still be a unique piece of headwear compared to the normal leather helmets.
> 
> 
> 
> * : After checking, turns out, the hood actually doesn't block some of your vision! It probably should, considering how all face-covering headgear does so, even regular reinforced hoods, which are worse.

Armor fixed, still not sure how to make it block vision.

john-pawson:
> Armor fixed, still not sure how to make it block vision.

Check hood code!

Fn-OS:
> > Armor fixed, still not sure how to make it block vision.
> 
> Check hood code!

OOoonly thing I'm seeing I'm seeing is block2add = FOV_BEHIND
Assuming that's it

Fn-OS:
Should be fixed!

Guidesu:
preliminary glance, everything looks okay, small enough to not need a TM, so yes. Good work.

john-pawson:
You're not even... gonna wait for the code review to get seen or applied before merging it? I'll open up a PR in a moment to do the changes I mentioned but, why the rush?

CaydenT:
OH HELL YES! I missed this thing! It's super cool looking!

john-pawson:
Reviewed errors were fixed on Jam's clothing cleanup PR

PR #1473, opened 2026-04-28
fixes wood dropping essance of the wilds instead of lumber

Elli-Skala:
## About The Pull Request

this broke months ago back when we did the AP parity pr. AP never had essence of lumber but made trees drop essance of wilds instead and it got override. making the only way to still get it via chopping tree stumps

## Testing Evidence
none yet left as draft

## Why It's Good For The Game

bug fix, having to cut 20 tree stumps for one for my t2 cog is agonizing and wild essance is supost to reward hunters

## Changelog

:cl:
fix: fix trees dropping essance of the wilds instead of essence lumber
/:cl:


Elli-Skala:
Couldnt figure out local host

PR #1479, opened 2026-04-28
Balancejak the Arcane Trickster

KeberosNihilus:
## About The Pull Request

Makes the Arcane Trickster not an (almost) objective upgrade to the thief.

## Testing Evidence

proof01

Using the Venardine race, if anyone's wondering why the stats don't line up with the changelog. Statpack was virtuous with no virtues taken.

## Why It's Good For The Game

Pretty straight forward, just lowers the stats of the Arcane Trickster a little so there's an actual tradeoff for being one instead of a regular thief. For some reason they had more stats than a thief on top of having magic? Now they don't!

For a more in-depth explanation why this matters: Arcane Trickster's access to magic can make one intent on being a nuisance _really_ hard, to borderline impossible, to pin down. Two int lets them easily hit the hardcap for spell scaling, as well as access to steel weapons for Conjure Weapon baseline, while one forces them to take intellectual and a statpack if they want that sweet 15. Especially if they want to max out invisibility's duration scaling.

Less willpower is obvious. Less stamina (and a common -stat in statpacks) so you have less room for error than a normal rogue. Play smart or suffer the consequences.

Tangentially makes them worse at combat with less stamina and worse feints, too. You got magic, be tricky! Or just keep doing what you're doing and Enchant Weapon: Force that cudgel of yours to break everyone's armor you maniac.

## Changelog

:cl:
balance: Arcane Trickster gets 1 less WIL and INT.
/:cl:

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/combat/mage.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/combat/mage.dm
@@ -246,9 +246,8 @@
 	traits_applied = list(TRAIT_ARCYNE_T2, TRAIT_DODGEEXPERT, TRAIT_LIGHT_STEP) //dodge expert has the potential for being a big pain on spellcasters,  so we take away their mage armor as a trade.
 	subclass_stats = list(
 		STATKEY_STR = -1,
-		STATKEY_INT = 2,
+		STATKEY_INT = 1,
 		STATKEY_PER = 1,
-		STATKEY_WIL = 1,
 		STATKEY_SPD = 2,
 	)
 


Xander3359:
I'd rather see the rogue classes buffed rather than seeing trickster nerfed in the lamest way possible.

This also puts them at a weight statspread of +4 which is quite awful.

KeberosNihilus:
In fairness, they _are_ also one of the few mages with a combat skill (Expert Dodger) on top of having a trio of inbuilt spells.

That said, no matter what happens, I do think - at least - they need to go down to 11 INT for the reasons listed in the OP.

Jamdrawers:
> I'd rather see the rogue classes buffed rather than seeing trickster nerfed in the lamest way possible.

Gotta be careful about Powercreep. That's kinda like... the definition of powercreep!

anconfuzedrock:
you do need int and wil to avoid passing out from trying to cast spells. I think -1 con or maybe lower knives skill would be better for keeping them playable, but weaker.

PR #1482, opened 2026-04-29
Portable alchemy equipment [Bounty] [Port]

john-pawson:
## About The Pull Request

Partial port of: https://github.com/Azure-Peak/Azure-Peak/pull/4530

Ports portable cauldrons and alchemy stations! Can be bought at the merchant for 45 each or be made by the artificer at their table, the alchemy station works the same as a normal one but foldable, meanwhile the cauldron has a maximum capacity of 60 drams and fuel for just 2 minutes, which means any potions you brew will be limited to those 60 drams. 

I also removed the additive reagent from the results of recipes, since due to the maximum capacity limit it meant that the normal reagent would fit but the additive wouldn't, making strong recipes impossible to make. (Maybe we could turn this semi-bug into a feature? Not let people make strong potions on the portable cauldron? But considering how they already lose a majority of the potion by just making it there... I'm not too sure)

Anyways I only did this since I saw nothing justifying the presence of the additive other than the "the mix glows" message when the brew reacts, if there's an actual reason these recipes use additive instead of the already reacted reagents, do let me know so I can revert it and then work something out.

On a minor note: I changed folding table code so I can use it with subtypes, since the original PR just copy-pasted the folding table code twice and then changed some lines.

## Testing Evidence

image
image
image

## Why It's Good For The Game

Lets things such as barber surgeons, certain mages or simple alchemy enthusiasts to do alchemy while gathering the ingredients or on the run with much more convenience at a cost, since the alchemy station and normal cauldron only really needs a log, 2 cloth, 2 sticks and 9 stones, I doubt this'll be much of a balancing concern, since making a cauldron on-the-go isn't that hard to do on the first place.

## Changelog

:cl:
add: Added portable alchemy stations and cauldrons! The cauldron can only hold 60 drams, so you won't be able to make full potion batches with it.
add: The portable alchemy station and the portable cauldron can be bought at the merchant or made by an artificer.
code: Folding table code slightly changed so subtypes can be made of it more easily.
code: Strong cauldron recipes no longer use half normal potion half addittive when being made, now they simply give the full amount of strong potion.
/:cl:


maker-of-nightmares:
I could see a class like Plague Bearer using these to empower their gang in the field with Strong Red. Maybe raise the price/difficulty of aquiring a bit so it's not so readily abused? 45 afterall is only slightly more than most classes spawn with

john-pawson:
> I could see a class like Plague Bearer using these to empower their gang in the field with Strong Red. Maybe raise the price/difficulty of aquiring a bit so it's not so readily abused?

Sounds like its working as intended...? Don't know how this would exactly count as "abused".

Means the plague bearer has to risk going into town and the merchant's or artificer's to get the cauldron, not only that but, the inventory space that the cauldron takes on the first place is the same that you can get by carrying around two bottles, which means you either: 

sacrifice 4 slots for the cauldron, 2 slots of ingredients per batch, and have some water + fuel source nearby to get 60 units of strong red.

...or carry 2 bottles filled with red beforehand, which is 50 drams per 2 slots.

maker-of-nightmares:
Ingredients can be stored in a Summoners Bag. That bag can hold literally dozens of potions worth of materials thus making the only difficult issue sourcing water(which is pretty plentiful). So it's about 16 brews worth of space for 1 belt slot and 4 bag slots.

This also doesn't account for backpacks which can store MUCH more and doesn't account that this can also supply raiders with additional potions they can carry in the field than their backpack can theoretically hold especially if they have 2 summoners pouches and a backpack full of potion supplies.

john-pawson:
> Ingredients can be stored in a Summoners Bag. That bag can hold literally dozens of potions worth of materials thus making the only difficult issue sourcing water(which is pretty plentiful). So it's about 16 brews worth of space for 1 belt slot and 4 bag slots.
> 
> This also doesn't account for backpacks which can store MUCH more and doesn't account that this can also supply raiders with additional potions they can carry in the field than their backpack can theoretically hold especially if they have 2 summoners pouches and a backpack full of potion supplies.

Still means they'll be losing 40% of the potions made because they decided to use a portable cauldron rather than 5 stones and 3 sticks.

RiceGateau:
Part of me really wants to say remove it from artificer and put it in general crafting. Mainly due to how artificer really has had a low pick rate historically. Might be different now due to guns being added but they are pretty rare still IMO.

I really don't think it is too busted of an addition as majority of potion makers rarely leave their crafting area anyhow except to gather ingredients.

PR #1484, opened 2026-04-29
Adds a Crossbow Virtue

EvenInDeathIStillServe:
## About The Pull Request

adds a crossbow virtue that works just like the bowman virtue

## Testing Evidence

image
image

## Why It's Good For The Game

every weapon has a virtue except crossebows

## Changelog

:cl:
add: Adds a crossbowman virtue that gives you +1 crossbow skill, along with a stashed crossbow and bolts.
/:cl:

### DIFF
--- a/modular_azurepeak/virtues/combat.dm
+++ b/modular_azurepeak/virtues/combat.dm
@@ -122,6 +122,20 @@
 	else
 		added_skills = list(list(/datum/skill/combat/bows, 1, 6))
 
+/datum/virtue/combat/crossbowman
+	name = "Crossbow Levy"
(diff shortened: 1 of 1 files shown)


Jamdrawers:
Always felt it was kinda intentional that crossbows are more rare/less accessible than bows.

ariaworld:
honestly crossbows are good if you're behind cover, every time I use a crossbow I end up getting hit because you need to stay still to reload it, bows are good because you can move

Xander3359:
Considering a crossbow is 30 mammons from the goldface, and grinding ranged XP levels is super easy. This is fine.

PR #1485, opened 2026-04-29
[fetishslop] Knot Riding and other fetish things

IronDragoon:
## About The Pull Request

- Adds powerbottom appropriate sexcon actions relating to knots. Currently if you ride someone else with a knot, you cannot be knotted unless the top does the penetrating. This adds a 'forcing knot' toggle in appropriate situations which the bottom can use which will result in them being knotted upon the top cumming. 

- Also adds knot grinding subtle option, and the bottom is able to knot grind if they forced it inside themselves. They can also unknot themselves if they forced it

- Also I added a few extra messages which are visible to those nearby when someone is cum onto

- Cum now satiates hunger slightly from oral sex actions and anywhere else I found it was possible (floor cleaning, tongue bathing)

- More detailed messages around someone actively dripping cum. This can also now be lessened/cleaned entirely through rimming/cunnilingus and will satiate hunger/stain the face of the person doing it as well as cleaning up the receiver

- Adds subtle oral options

- Adds 'Cumdrunk' effect after consuming cum. This will provide a small mood buff for 10 minutes for nymphomaniacs.

- Readds the 'non-knotted lupians' count to the baotha part of the end round influences page. I dunno why it was removed but considering the god influence system does essentially nothing and is just for fun, I wanted it back. I also added a 'Impregnations' to this, for the same reason (fun)

- Tech stuff, I refactored some sexcon code to add the variable 'btm', which is the receptive participant; explicitly passed for inverted-role actions (riding, blowjob) where self.target may not point at the rider/sucker. AFAIK this should not cause any issues, and testing seems to confirm this. 

- Because of how the pain/injury stuff is handled, this type of action is not going to injure the bottom, but this feels fine to me because, power bottom. 

## Testing Evidence

I've tested about as much as I can on localhost, everything seems to perform as expected, as well as fixing several bugs with help from others in the comments. Seems fairly robust now

<details>
  <summary>Images contain SEX</summary>
image
image
image
image
image

</details>

## Why It's Good For The Game
~~I am a switch and want to power bottom~~
Just feels like something that should be in game, if you're able to ride someone you should be able to knot yourself too. It's just neat. People complain a lot about 'fetishslop' but realistically this sort of thing should not affect anyone that is not 'into it' and only allow those who are to have a better experience. I've made the code as modular and sensible as I am able so it should have minimal overhead (in some cases where I was able to simplify things, even a positive one). 

## Changelog

:cl:
add: Added ability for bottoms to force themselves onto a top's knot
add: Ability for bottoms to grind a knot in themselves if they are the one forcing it
add: Subtle knot grinding
add: Some gooner local visible messages about people taking loads on their body/face
add: Can now 'clean' a partner who is actively 'dripping' through rimming/cunnilingus actions.
add: Can now consume cum/femcum to gain a small amount of nutrition through oral sex. This should also be implemented in all the other areas I thought of including cleaning mentioned above, as well as cleaning floor stuff
add: More detailed examine text around active cum dripping
add: Baotha round end influence now tracks non-lupians knotted as well as impregnations
refactor: refactored sexcon a little to respect the bottom/top as opposed to just the 'instigator' 
/:cl:


code/datums/sexcon/sexcon_knotting.dm
```diff
@@ -30,81 +30,106 @@
...
-				to_chat(target, span_userdanger("You have been double-knotted!"))
+				to_chat(btm, span_userdanger("You have been double-knotted!"))
 			else
-				to_chat(target, span_userdanger("You have been triple-knotted!"))
+				to_chat(btm, span_userdanger("You have been triple-knotted!"))
 		if(we_got_baothad)
-			to_chat(target, span_userdanger("Baotha magick infuses within, you can't think straight!"))
-	if(!target.has_status_effect(/datum/status_effect/knot_tied)) // only apply status if we don't have it already
-		target.apply_status_effect(/datum/status_effect/knot_tied)
+			to_chat(btm, span_userdanger("Baotha magick infuses within, you can't think straight!"))
```
Gug1234:
```suggestion
			to_chat(btm, span_userdanger("Baothan magick infuses within, you can't think straight!"))
```

IronDragoon: Thanks, resolved in latest push!=

code/datums/sexcon/sex_actions/oral/blowjob.dm
```diff
@@ -4,6 +4,8 @@
 	category = SEX_CATEGORY_PENETRATE
 	user_sex_part = SEX_PART_JAWS
 	target_sex_part = SEX_PART_COCK
+	knot_on_finish = TRUE
+	subtle_supported = TRUE
```
tmyqlfpir, line 8: Subtle and knotting [are not supported on the UI](https://github.com/user-attachments/assets/2788210d-07c8-490e-a881-d7c32b4ca51d).

Either update the UI to support actions that both have subtle and knotting functionality, or remove subtle for this action.

IronDragoon: good catch, I am mid testing right now, but yep I will adjust this

tmyqlfpir:
Testing this locally with two clients [I could not get the character on top to be knotted](https://github.com/user-attachments/assets/91ac0920-f02d-4150-8a5c-af15f10a4b87).

IronDragoon:
> Testing this locally with two clients [I could not get the character on top to be knotted](https://github.com/user-attachments/assets/91ac0920-f02d-4150-8a5c-af15f10a4b87).

can you let me know what types of dick the characters had and the situation? It seems to have worked in almost everything i've tested, but will keep testing

tmyqlfpir:
> can you let me know what types of dick the characters had and the situation? It seems to have worked in almost everything i've tested, but will keep testing

Investigating this further this is caused by the top's sexcon having target set to null.

This happens when the top has not opened the sexcon menu over the current mob riding them, and so when it calls `cum_into()` target is set to null as it was called by the bottom.

The easiest (and hackiest) solution would be to add an target override parameter to `cum_into()` so when it is called by the bottom, it will still try to follow through with attempting to knot.

```
/datum/sex_controller/proc/cum_into(oral = FALSE, mob/living/carbon/human/splashed_user = null, datum/sex_action/knot_action = null, knot_swap_roles = FALSE, mob/living/carbon/human/knot_btm = null, orifice = SEX_PART_NULL, mob/living/carbon/human/target_override = null)
	var/mob/living/carbon/human/cum_target = target_override ? target_override : target
```
Then use `cum_target` for the rest of the proc instead of `target`.

IronDragoon:
> > can you let me know what types of dick the characters had and the situation? It seems to have worked in almost everything i've tested, but will keep testing
> 
> Investigating this further this is caused by the top's sexcon having target set to null.
> 
> This happens when the top has not opened the sexcon menu over the current mob riding them, and so when it calls `cum_into()` target is set to null as it was called by the bottom.
> 
> The easiest (and hackiest) solution would be to add an target override parameter to `cum_into()` so when it is called by the bottom, it will still try to follow through with attempting to knot.
> 
> ```
> /datum/sex_controller/proc/cum_into(oral = FALSE, mob/living/carbon/human/splashed_user = null, datum/sex_action/knot_action = null, knot_swap_roles = FALSE, mob/living/carbon/human/knot_btm = null, orifice = SEX_PART_NULL, mob/living/carbon/human/target_override = null)
> 	var/mob/living/carbon/human/cum_target = target_override ? target_override : target
> ```
> 
> Then use `cum_target` for the rest of the proc instead of `target`.

Ahhhh, gotcha. Yep, will fix, thank you!

IronDragoon:
> > can you let me know what types of dick the characters had and the situation? It seems to have worked in almost everything i've tested, but will keep testing
> 
> Investigating this further this is caused by the top's sexcon having target set to null.
> 
> This happens when the top has not opened the sexcon menu over the current mob riding them, and so when it calls `cum_into()` target is set to null as it was called by the bottom.
> 
> The easiest (and hackiest) solution would be to add an target override parameter to `cum_into()` so when it is called by the bottom, it will still try to follow through with attempting to knot.
> 
> ```
> /datum/sex_controller/proc/cum_into(oral = FALSE, mob/living/carbon/human/splashed_user = null, datum/sex_action/knot_action = null, knot_swap_roles = FALSE, mob/living/carbon/human/knot_btm = null, orifice = SEX_PART_NULL, mob/living/carbon/human/target_override = null)
> 	var/mob/living/carbon/human/cum_target = target_override ? target_override : target
> ```
> 
> Then use `cum_target` for the rest of the proc instead of `target`.

realised theres a smarter fix, I can use knot_btm to bypass the guard as it already was doing what I wanted I'd just forgot to gate it properly in that proc. Thanks for finding this bug!

tmyqlfpir:
> realised theres a smarter fix, I can use knot_btm to bypass the guard as it already was doing what I wanted I'd just forgot to gate it properly in that proc. Thanks for finding this bug!

Great idea. There is still target being used for `after_intimate_climax()` as sexcon was always written to assume `cum_into()` is called by the top. That may need the same argument to overwrite.

IronDragoon:
> > realised theres a smarter fix, I can use knot_btm to bypass the guard as it already was doing what I wanted I'd just forgot to gate it properly in that proc. Thanks for finding this bug!
> 
> Great idea. There is still target being used for `after_intimate_climax()` as sexcon was always written to assume `cum_into()` is called by the top. That may need the same argument to overwrite.

Yep, I will sort that too. Sorry this is kind of why my code sucks, I keep thinking of things to add as i'm doing it...

And yeah looking at that I will need to rewrite it slightly to use an 'effective target'. Shouldn't be too difficult

code/datums/sexcon/sexcon.dm
```diff
@@ -337,55 +340,119 @@
...
 		if(!oral)
 			var/obj/item/organ/testicles/testes = user.getorganslot(ORGAN_SLOT_TESTICLES)
 			if(testes?.ball_size > DEFAULT_TESTICLES_SIZE)
-				splashed_user.apply_status_effect(/datum/status_effect/creampie_leak/long)
+				splashed_user.apply_status_effect(/datum/status_effect/creampie_leak/long, orifice)
 			else
-				splashed_user.apply_status_effect(/datum/status_effect/creampie_leak)
+				splashed_user.apply_status_effect(/datum/status_effect/creampie_leak, orifice)
 	if(target.has_flaw(/datum/charflaw/addiction/lovefiend))
 		target.sate_addiction(/datum/charflaw/addiction/lovefiend)
```
tmyqlfpir: Commit f33cb638d6869378ca717fdf8ccdeafdae616f7e introduced these two lines when it is possible target is set to null. This is pretty bad and I'm not sure why nobody noticed this yet. Ideally if you use the above solution and take an target override, it will also fix this bug.

IronDragoon: yup, noted I'll make sure this will be covered

tmyqlfpir:
The same fix will also need to be applied to `cunnilingus.dm`'s `target.sexcon.cum_into()` by setting `splashed_user = user` argument.

IronDragoon:
> The same fix will also need to be applied to `cunnilingus.dm`'s `target.sexcon.cum_into()` by setting `splashed_user = user` argument.

yep, I was working on that one when you mentioned the previous issue!

IronDragoon:
(Okay I think that's cleanup of all the bare target references now)

IronDragoon:
With this final adjustment I'm ready for this to be TMed now, I don't have anything else I want to add specifically within this PR.

ophaq:
It's peak sire. I have been wanting the knot stats for a while. The knot riding and other stuff is a great bonus too.

n3tis:
Why just knottings from non-Lupians? Other canine species exist too... also, werewolves!

PR #1486, opened 2026-04-29
Adds sneaking hiding spots

Xander3359:
## About The Pull Request
Ports of https://github.com/Azure-Peak/Azure-Peak/pull/6622

## Testing Evidence
I was able to hide in all the things that were added as hiding spots.
floor crawling and wall leaning alpha works as expected

## Why It's Good For The Game
Makes sneaking cooler, see the original PR for more details.

## Changelog

:cl: NalimaTheWitch, Xander
add: Hiding spots. Go examine bushes or tables to find out how to use them.
add: Crawling and wall-leaning sneaking: crawl or wall-lean with sneak mode on to become partially transparent based on Sneak skill.
add: You can spot people in hiding spots by standing next to them and examining the object, or by using 'look around' (fixed eye button).
qol: Far-looking with sneak mode on no longer prints a visible message to others.
qol: Sneaking movement speed now scales from Sneak skill.
/:cl:

### DIFF
--- a/code/game/objects/structures/beds_chairs/roguechair.dm
+++ b/code/game/objects/structures/beds_chairs/roguechair.dm
@@ -9,8 +9,13 @@
 	sleepy = 0.5
 //	pixel_y = 10
 	layer = OBJ_LAYER
+	hidingspot = TRUE
+	/// So we can find them with fixed eye search
+	var/mob/living/hiddenguy = null
 
-
+/obj/structure/chair/bench/examine(mob/user)
+	. = ..()
+	. += span_info("Some structures can be used as hiding places. Toggle the 'SNEAK' button on your HUD, then click the structure to hide in it. You can stop hiding by clicking the structure again, or by moving out of it.")
 
 /obj/structure/chair/smallbench
 	name = "small bench"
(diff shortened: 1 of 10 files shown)


Jamdrawers:
I will kinda miss the more organic strategy of physically inching your character's model behind the actual sprite of the object. Imagine if right clicking with sneak intent would, instead of make your character dissappear completely, place it on the layer beneath that object for the time...

NotGoldenglow:
Presently, you cannot look afar OR examine people while sneaking AND wall-leaning. It 'asks you to move away from the wall first'. Is this intended?

yzsentia:
I never really leave feedback, but I play a char that is primarily a sneak build, so I'll leave my thoughts:
I know its a port of APs sneak, I actually like the core of it more than the original, it feels like it gives you more options for hiding and makes indoor espionage have more strategy.
Here's what I don't like:
Not being able to inspect or look ahead while leaning against walls is such a severe handicap that it hurts the efficacy of sneaking substantially. I do not mind if you cant see very much while hiding under a table or bush for instance, but regular wall leaning in stealth is a much more mobile option, and should allow the user to still see their surroundings.
A by-product of this is the fact that you'll lean against a wall, try to look ahead, "I need to get off the wall first", so you press a few movement buttons to get off the wall, except NOPE!, you still can't see anything! so you have to physically leave the tile in a direction not against a wall.

Honestly, tweaking just that one feature would go a long way to making sneaking feel more natural, and is my only pressing concern. I hope its made a priority, because wow :0 I love sneaking around, but I wouldn't change it in any way that would make it more obtuse to the player.

code/game/objects/structures/beds_chairs/roguechair.dm
```diff
@@ -48,6 +53,40 @@
 		layer = OBJ_LAYER
 		plane = GAME_PLANE
 
+/obj/structure/chair/bench/attack_hand(mob/user)
```
john-pawson, line 56: Missing . = ..() on attack_hand means you can no longer unbuckle people from benches, beds, etc.

Vizoplez:
https://github.com/user-attachments/assets/104bd0f0-e5a1-4e26-9f1d-5bf39f785ff5

you broke my PR grrrr (jokes aside, you do need to fix a few issues; you should know what they are already but-
1) Leaning off of walls sets your Alpha back to default, breaking the point of this entirely (as you need to toggle sneak to become stealthy again)
2) Laying down has no cooldown for stealth, and allows you to bypass forced alpha raises/etc.

PR #1487, opened 2026-04-29
Burrying someone now refunds their job slots.

Xander3359:
## About The Pull Request
Makes it so that when you bury someone their job is refunded to be able to be picked again.

## Why It's Good For The Game
I thought it was like this before.
If you bury someone and then grave mark them, it sends them back to lobby. It should also reopen the job slots. 
It's dumb that you can only refund jobs by far travelling them, when being buried is the same type of RR.

## Changelog

:cl:
add: Grave marking a corpse will now refund their job as well
/:cl:

### DIFF
--- a/code/datums/mind.dm
+++ b/code/datums/mind.dm
@@ -241,10 +241,8 @@ GLOBAL_LIST_EMPTY(personal_objective_minds)
 	if(M.known_people && istype(H))
 		M.known_people -= H.real_name
 
-
 /datum/mind/proc/unknow_all_people()
 	known_people = list()
-
 
 /datum/mind/proc/display_known_people(mob/user)
 	if(!user)
--- a/code/modules/mob/living/carbon/spirit/spirit.dm
+++ b/code/modules/mob/living/carbon/spirit/spirit.dm
@@ -230,6 +230,29 @@
 					qdel(human_corpse.mouth)
 					human_corpse.update_inv_mouth()
 					break
+
+	var/datum/job/mob_job
+	mob_job = SSjob.GetJob(corpse.mind.assigned_role)
+	if(mob_job)
+		mob_job.current_positions = max(0, mob_job.current_positions - 1)
+		var/target_job = SSrole_class_handler.get_advclass_by_name(user.advjob)
+		if(target_job)
+			SSrole_class_handler.adjust_class_amount(target_job, -1)
+	corpse.mind.unknow_all_people()
+	for(var/datum/mind/MF in get_minds())
+		corpse.mind.become_unknown_to(MF)
+	for(var/datum/bounty/removing_bounty in GLOB.head_bounties)
+		if(removing_bounty.target == corpse.real_name)
+			GLOB.head_bounties -= removing_bounty
+	GLOB.chosen_names -= corpse.real_name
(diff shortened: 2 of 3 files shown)


Azlier:
It's a good idea in principle, but it _always_ leads to some jackass spawning in, assassinating and burying the duke, and immediately far traveling to rejoin as duke. Similar things have happened already. I don't like people.

coffeecoffeecoffee:
This is not a good idea in principle. Being able to remove people and roles for the round is an essential game mechanic for the progression of the round or for really any round drama to take place beyond "lol let's all talk things out, kumbaya!" - and that's not to even speak of how it enables people to respawn as the same role over and over again.

We need _more_ round removal, not less.

john-pawson:
Do need to point out that dead people not refunding slots is *intentional*, and people getting buried and/or returning to the lobby already reopens slots, that's why the "Open slots on death" variable exists. 

(Upon checking it looks like there's an early return upon being buried that ignores the open slots on death check, so it's actually bugged apparently, may want to go ahead and simply move that check into burial code and/or fix the one from "Journey to the underworld")

Also it looks like you're removing them from everything? Just making burial the same as far travel? Even known list? Why would people stop knowing someone if they were buried? To stop scrying? Bounty removed so you can't graverob someone for money? You even put a special check for lords. You can just- bury the duke and get a new one now? Burying bandits now gives them new slots... burying a gnoll reopens a slot for a new one...

Hmeister-fake:
> Also it looks like you're removing them from everything? Just making burial the same as far travel? Even known list? Why would people stop knowing someone if they were buried? To stop scrying? Bounty removed so you can't graverob someone for money? You even put a special check for lords. You can just- bury the duke and get a new one now? Burying bandits now gives them new slots... burying a gnoll reopens a slot for a new one...

Most thought out Xander PR to date btw.

JackSage:
Maybe if it was only for certain roles, but like the duke dying it seems stupid.

The duke is forced DNR because a PR a while back where the idea is that if the duke dies and can't be replaced the duchy has to scramble to find a replacement duke which adds tension and RP, heirs actually having to lead instead of being ERP bait

But now the duke dies, is buried because DNR and either before a new duke is crowned or after someone joins as the duke slot that was opened due to the duke's death.

I think leadership/important rolls should be ineligible. Bishop? No, Acolyte? Sure. Marshal? No. MAA? Sure.

Xander3359:
Okay so from what I can tell.
You can literally far travel corpses...
This means you are incentivized to NEVER bury corpses because it griefs job slots.
If you encounter a dead warden and you choose to bury them, when far travel (their corpse) is an option you are literally griefing.

Hmeister-fake:
> Okay so from what I can tell. You can literally far travel corpses... This means you are incentivized to NEVER bury corpses because it griefs job slots. If you encounter a dead warden and you choose to bury them, when far travel (their corpse) is an option you are literally griefing.

Okay so that's a bug. If someone far-travels a Gnoll to open a Gnoll slot that's bug abuse.

coffeecoffeecoffee:
> > Okay so from what I can tell. You can literally far travel corpses... This means you are incentivized to NEVER bury corpses because it griefs job slots. If you encounter a dead warden and you choose to bury them, when far travel (their corpse) is an option you are literally griefing.
> 
> Okay so that's a bug. If someone far-travels a Gnoll to open a Gnoll slot that's bug abuse.

It's almost like this PR opens up extremely easily exploitable bugs.

PR #1488, opened 2026-04-29
TDI begone attempt

MarioTheShiba:
## About The Pull Request

hi.

progress bar is lagging shit.

viewer scan moved out of update() which was being hammered every tick. update() is now just a clamp + icon_state set
refresh_viewers() runs once on New() to seed the tracked clients list.
re-runs only when the user moves, via COMSIG_MOVABLE_MOVED. so a player walking mid craft still gets the bar shown to newly-visible viewers, but standing still does, and costs nothing.
tracked_clients check uses in instead of |=, skipping the linear find-and-append on EVERY ALREADY TRACKED CLIENT.
signal unregistered in Destroy() so we don't leak the hook.

## Testing Evidence

bismillah i hope it works.

## Why It's Good For The Game

at 150 average players with observers stacking, this should - within theory, drop progressbar/update from #1 self-cpu to a non-entry on the profiler. the viewers() calls only fire on actual movement now, not 4172 times in a screenshot i saw which caused me to put a bullet in my head.

## Changelog

:cl:
fix: unclogs the progressbar.
/:cl:

### DIFF
--- a/code/datums/progressbar.dm
+++ b/code/datums/progressbar.dm
@@ -30,14 +30,20 @@
 	bar.alpha = 0
 	animate(bar, pixel_y = 32 + (PROGRESSBAR_HEIGHT * (listindex - 1)), alpha = 255, time = PROGRESSBAR_ANIMATION_TIME, easing = SINE_EASING)
 
-/datum/progressbar/proc/update(progress)
-	if(!tracked_clients)
-		tracked_clients = list()
-	for(var/mob/M in viewers(user, null))
-		if(M.client)
-			M.client.images |= bar
-			tracked_clients |= M.client
+	tracked_clients = list()
+	refresh_viewers()
+	RegisterSignal(user, COMSIG_MOVABLE_MOVED, PROC_REF(refresh_viewers))
 
+/datum/progressbar/proc/refresh_viewers()
+	SIGNAL_HANDLER
+	for(var/mob/M in viewers(user, null))
+		var/client/C = M.client
+		if(!C || (C in tracked_clients))
+			continue
+		C.images |= bar
+		tracked_clients += C
+
+/datum/progressbar/proc/update(progress)
 	progress = CLAMP(progress, 0, goal)
 	last_progress = progress
 	bar.icon_state = "prog_bar_[round(((progress / goal) * 100), 5)]"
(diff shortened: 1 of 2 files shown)


gatorceres:
Performance enhancements. I kneel.

MarioTheShiba:
> Performance enhancements. I kneel.

the tradeoff to this is if someone is trying to escape chains off screen - you won't see it if you walk on screen.

i could put in a checkup for long do_afters, but honestly this would legit cut down on TDI so fucking hard i doubt anyone would give a shit.

gatorceres:
> > Performance enhancements. I kneel.
> 
> the tradeoff to this is if someone is trying to escape chains off screen - you won't see it if you walk on screen.
> 
> i could put in a checkup for long do_afters, but honestly this would legit cut down on TDI so fucking hard i doubt anyone would give a shit.

I think that is an acceptable tradeoff for now since the tidi severely affects everyone, and that bug will only really affect people who are chained up, not moving, and unguarded, which probably mostly happens when the Inquisition leaves somebody alone in their dungeon.

MarioTheShiba:
I also fixed LOOC whilst I was at it.

changed to get_hearers_in_view(7, src)

now instead of like 150 LOS scans, it does one. yes, i am him sometimes.

MarioTheShiba:
atomized #1490

code/datums/progressbar.dm
```diff
@@ -30,14 +30,20 @@
 	bar.alpha = 0
 	animate(bar, pixel_y = 32 + (PROGRESSBAR_HEIGHT * (listindex - 1)), alpha = 255, time = PROGRESSBAR_ANIMATION_TIME, easing = SINE_EASING)
 
-/datum/progressbar/proc/update(progress)
-	if(!tracked_clients)
-		tracked_clients = list()
+	tracked_clients = list()
+	refresh_viewers()
+	RegisterSignal(user, COMSIG_MOVABLE_MOVED, PROC_REF(refresh_viewers))
```
noelle-lavenza, line 35: you probably want some kind of proximity checking, too, because what happens if someone walks into view range of a stationary target? they won't be added. that's why it checked in update().

noelle-lavenza (changes requested):

code/datums/progressbar.dm
MarioTheShiba: it doesn't showcase progress bars. its a cool way of actually having people escape ropes, but i see ur point. its more of a 'hey, lets no upgrade this every tick because it uber lags!' type shit.

noelle-lavenza:
i don't think this is a good tradeoff, i would suggest something like replacing `viewers(src, null)` with `get_hearers_in_view(world.view, src, RECURSIVE_CONTENTS_CLIENT_MOBS))` as a first pass.

MarioTheShiba:
> i don't think this is a good tradeoff, i would suggest something like replacing `viewers(src, null)` with `get_hearers_in_view(world.view, src, RECURSIVE_CONTENTS_CLIENT_MOBS))` as a first pass.

pls suggest change and ill do it!

Kroto-Ker:
Alternative was merged.

PR #1495, opened 2026-04-30
Natural Armor Flight Fix for Harpies

ophaq:
## About The Pull Request
- 'Fixes' natural armor preventing harpies from taking flight.

## Testing Evidence
harpynatarmortest

## Why It's Good For The Game
Currently you can take this virtue as a harpy, and then suffer permanent no-flight, despite it being a part of your actual body. Harpies are already pretty fragile, and since you use up a virtue to make yourself less squishy (in comparison to a +CON stat boost), you end up maiming yourself entirely for half of the purpose of being a harpy... to fly. Nobody noticed before, perhaps. But I think this will be fine.

## Changelog
:cl:
fix: Natural armor virtue now no longer prevents harpy flight from happening.
/:cl:

### DIFF
--- a/code/modules/surgery/organs/feature_organs/wings.dm
+++ b/code/modules/surgery/organs/feature_organs/wings.dm
@@ -98,6 +98,12 @@
 
 /obj/effect/proc_holder/spell/self/harpy_flight/cast(mob/living/carbon/human/user)
 	var/harpy_AC = user.highest_ac_worn()
+	// Natural/innate skin armor (e.g. Natural Armor virtue) occupies the shirt slot but is part of the body - don't let it block flight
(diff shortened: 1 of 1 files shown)


zeroisthebiggay:
this is technically the only armor that harpies are allowed to wear
harpies are very intentionally forbidden from flying while even wearing light armor

anconfuzedrock:
you can absolutely take +con AND nat armor, virtuous isn't necessary.

Xander3359:
This is intended. Flying is one of the strongest mechanics in the entire game and it's heavily restricted for a good reason.
Trust me you do not want armored flyers enabled.

PR #1502, opened 2026-04-30
Make Tiefling TRAIT_HELLSPAWN (actually do something)

Rin053:
## About The Pull Request

Refactors TRAIT_HELLSPAWN -> TRAIT_FIRERESISTANCE to be used globally.

Gives tieflings their long awaited DND damage resistance trait and makes the hellspawn trait ACTUALLY do something (it did nothing before).

Adds burn flags to the code (specifically BURN_FLAG_FIRE) for adding inherent resistances to traits.

Makes it so those with TRAIT_FIRERESISTANCE take 50% reduced damage from ALL sources flagged as fire(does not count all burn damage, only the damage flagged as FIRE)

Tieflings who are in covens as vampires (or are converted to vampires) do not get the benefits of TRAIT_FIRERESISTANCE.

This also makes it so the time it takes for them to heat up be halved with the temperature system.

## Testing Evidence

Testing using normal fireball from the spellbook aiming at the head.

Non hellspawn trait ->
dreamseeker_hIJ0Ht576X

Hellspawn trait ->
dreamseeker_YcYJOxgxdr

Roughly the same (duration of the burning different) as expected ~50% damage reduction

## Why It's Good For The Game

The hellspawn trait did nothing and was flavortext. This gives them their natural fire resistance being the infernal things they are.

## Changelog

:cl: SSix
add: Added 50% damage resistance to all fire sources from TRAIT_FIRERESISTANCE.
add: Tieflings are now actually fire and heat resistant.
refactor: Trait_HELLSPAWN is now TRAIT_FIRERESISTANCE
/:cl:


Hmeister-fake:
They should probably also have innate hot temprature insulation in this case, so they don't get heatstroke as fast.

Jamdrawers:
It miiiiight be easier to just tie fire resistance to a new trait called 'fire resistance' instead, so that in future things that aren't HELLSPAWN! can be fire resistant.

Rin053:
> They should probably also have innate hot temprature insulation in this case, so they don't get heatstroke as fast.

done

> It miiiiight be easier to just tie fire resistance to a new trait called 'fire resistance' instead, so that in future things that aren't HELLSPAWN! can be fire resistant.

done

VIRTUALCRIME:
fucking finally

PR #1508, opened 2026-04-30
LACRIMA - Lux for ZIZO

zeroisthebiggay:
## About The Pull Request

adds Lacrima, a tier three ZIZO miracle, and a version that doesn't use devotion for necromancers to use
Lacrima has you plunging your hand into a victim's chest, breaking their ribs, and then ripping out their impure lux
obviously, the whole fucking experience is pretty traumatic for the victim, encouraging you to do it on your enemies pretty much exclusively

doesn't work on simpleminded creatures, doesn't work on the undead or people with their lux already torn

also changes the necromancer's crystal to require impure lux instead
why was it pure lux again

to compensate for this, Lacrima has a second intent that allows you to pervert pure lux back into impure lux either to grief the church and or to fuel your dark crystal

## Testing Evidence

image

## Why It's Good For The Game

gives ZIZO miraclists something actually unique and cool, and adds some depth to the necromancer gameloop that encourages them to go out and harvest the lux from unsuspecting passerbys quickly and easily instead of having to go through a song and dance of capturing them alive to carefully surgery out their lux

## Changelog

:cl:
add: Lacrima, a tier three ZIZO miracle for necromancers that tears out the lux from a victim
/:cl:

### DIFF
--- a/code/datums/gods/patrons/inhumen/zizo.dm
+++ b/code/datums/gods/patrons/inhumen/zizo.dm
@@ -14,6 +14,7 @@
 					/obj/effect/proc_holder/spell/invoked/raise_undead_formation/miracle= CLERIC_T2,
 					/obj/effect/proc_holder/spell/invoked/tame_undead/miracle			= CLERIC_T3,
 					/obj/effect/proc_holder/spell/invoked/rituos/miracle 				= CLERIC_T3,
+					/obj/effect/proc_holder/spell/targeted/touch/lacrima				= CLERIC_T3,
 	)
 	confess_lines = list(
 		"PRAISE ZIZO!",
--- a/code/game/objects/items/rogueitems/necro_relics.dm
+++ b/code/game/objects/items/rogueitems/necro_relics.dm
@@ -25,7 +25,7 @@
 	..()
 	set_light(2, 2, 1, l_color = "#551c1c")
 
-/obj/item/necro_relics/necro_crystal/proc/recharge(obj/item/reagent_containers/lux/L, mob/user)
+/obj/item/necro_relics/necro_crystal/proc/recharge(obj/item/reagent_containers/lux_impure/L, mob/user)
 	if(current_charges >= max_charges)
 		to_chat(user, span_notice("The crystal is already brimming with power."))
 		return FALSE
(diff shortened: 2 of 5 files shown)


Venithefox:
now THIS. this is EPIC. and a good change to encourage necromancers to build up an army of player skeletons instead of exclusively the NPC guards.

also being a zizite heretic spy and assassinating people by ripping their lux out would be sovl

maker-of-nightmares:
> also being a zizite heretic spy and assassinating people by ripping their lux out would be sovl

Given the decently long windup showcased(I don't care to code dive RN to find the exact one and the testing evidence shows at least 5/10 seconds) this wouldn't be that easy a way to assassinate someone. You might be able to do this to wounded people on a battle field to bolster your own forces though.

IgiariValkyr:
I believe the reason the necromancer's staff requiring pure lux is the same reason you need pure lux to revive a soul. It has no focus otherwise. Yes, necromancy "has" no focus, but to be able to reach into the underworld and bring those souls back should require such. It is not easy to pierce the veil between the planes of existence. That and this spell making it so you rip out lux left and right means pure lux will be plentiful on the bad guy side.

zeroisthebiggay:
> I believe the reason the necromancer's staff requiring pure lux is the same reason you need pure lux to revive a soul. It has no focus otherwise. Yes, necromancy "has" no focus, but to be able to reach into the underworld and bring those souls back should require such. It is not easy to pierce the veil between the planes of existence. That and this spell making it so you rip out lux left and right means pure lux will be plentiful on the bad guy side.

a skeleton is not worth the effort of having to hold down two guys and rip out their whole lux
even after this change to make it faster and not require surgery skeletons aren't worth it

The-Sun-In-Splendour:
> I believe the reason the necromancer's staff requiring pure lux is the same reason you need pure lux to revive a soul. It has no focus otherwise. Yes, necromancy "has" no focus, but to be able to reach into the underworld and bring those souls back should require such. It is not easy to pierce the veil between the planes of existence. That and this spell making it so you rip out lux left and right means pure lux will be plentiful on the bad guy side.

Just because "in idea" raising a skeleton should NOT be easy, it should not come into balance by making skeletons require TWO LIVING PLAYERS for one person. Maybe if skeletons weren't so disgustingly weak.
The idea of necromancy being difficult is already reflected in the game: Necromancers are rare (2 max out of 10 wretches, and rarely will there be more than 1)

PR #1509, opened 2026-05-01
The Pathological Bottom Vice

zeroisthebiggay:
## About The Pull Request

adds the compliant vice
it turns on your compliance, and you can't turn it off

## Testing Evidence

image

## Why It's Good For The Game

someone asked, and I for one encourage empowering the zapebaits

## Changelog

:cl:
adds: compliance vice, take a guess what it does
/:cl:

### DIFF
--- a/code/__DEFINES/traits.dm
+++ b/code/__DEFINES/traits.dm
@@ -110,6 +110,7 @@
 #define TRAIT_CHASTITY_SPIKED "Genital Spikes" // Causes discomfort during arousal.
 #define TRAIT_CHASTITY_LOCKED "Locked Chastity Device" // Prevents removal of the chastity device.
 #define TRAIT_EXTREME_TEMPERATURE_IMMUNE "Extreme Temperature Immunity" //immunitty to heatstroke and frostbite without damage reduction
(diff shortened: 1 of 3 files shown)


zeroisthebiggay:
apparently you can click on the compliant status to also toggle it i gotta test if this covers that too
it covers using the verb and middeclicking the combat mode button though

zeroisthebiggay:
> apparently you can click on the compliant status to also toggle it i gotta test if this covers that too it covers using the verb and middeclicking the combat mode button though

yeah it works fine
this pr is good in its current state and fine to add

EmilyFontaine:
Are there any role restrictions here? I like it a lot, but it might be very frustrating to see dedicated combat classes with this vice.

zeroisthebiggay:
> Are there any role restrictions here? I like it a lot, but it might be very frustrating to see dedicated combat classes with this vice.

you choose the vices you take
if someone is playing as warden and taking this vice to be zaped easier then they were going to get zaped anyways

JackSage:
I don't really see the need or reason to add role restrictions since you can already just turn compliance mode on and run it 24/7 by default. This just forces it so you can't turn it off. Bad actor garrison members can already run compliance 24/7 manually.

PR #1529, opened 2026-05-03
Vampire Presence number tweaks (big nerfs)

NotGoldenglow:
## About The Pull Request

If vampires have 1 million haters, I'm one of them. If vampires have 1 hater, it's me. If vampires has 0 haters, I'm dead because they fragged the entire town.

## Testing Evidence

6 lines, mostly number values for durations/cooldowns.

## Why It's Good For The Game

You really shouldn't have three different abilities that completely remove someone from a fight for three seconds at a time (and also completely pacify them and make them unable to control themselves and thus defend themselves) for insanely cheap costs and on single minute cooldowns. The general consensus is that it's abusable, and so we want to remedy that.

## Changelog

:cl:
balance: vampire presence ability costs have been increased by 50%, reduced in duration from between 30 to 50%, and have had cooldowns doubled.
/:cl:

### DIFF
--- a/code/modules/vampire_neu/covens/coven_powers/presence.dm
+++ b/code/modules/vampire_neu/covens/coven_powers/presence.dm
@@ -18,10 +18,10 @@
 	research_cost = 0
 	check_flags = COVEN_CHECK_CAPABLE | COVEN_CHECK_SPEAK
 	target_type = TARGET_HUMAN
-	vitae_cost = 100
+	vitae_cost = 150
 	range = 4
 	multi_activate = TRUE
-	cooldown_length = 60 SECONDS
+	cooldown_length = 2 MINUTES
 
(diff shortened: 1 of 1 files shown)


DatFrostys:
I don't know about the third, but the two first abilities don't pacify you?? I've seen people legit just successfuly fight back against the force walk by running in another direction and/or hitting the vampire using it. The cost change is totally justified, these abilities should cost a lot to cast but increasing the cooldown and reducing the time it last seems unecessary.

maker-of-nightmares:
I think you should test if the cost and cooldown increase are enough to make it not abuseable before touching the time period of effect. Touching everything at once is a bit much to test where it'll be comfortable and will just result in someone PRing it back to normal or up.

PR #1531, opened 2026-05-03
Removes a lot of janky and mostly nonfunctional 'slapcraft' recipes

Jamdrawers:
## About The Pull Request

There's a lot of code for special crafting where you touch a log with another log to build something. Some of this is nice, a lot of this either doesn't work or is annoying - every time you touched a log with another log you'd be asked if you want to make a boat, which actually requires 4 of a DIFFERENT type of log along with like three ash and five fiber, which it didn't tell you. That sort of thing is not a job for slapcrafting, that sort of thing is what the craft menu is for.

It also cuts a number of 'peasant millitia' recipes that crafted metal tools out of logs, ropes, and huge quantities of whetstones. A lot of them just didn't make any sense

## Testing Evidence

Compiles, works

## Why It's Good For The Game

As described above

## Changelog
:cl:
del: Cuts a large number of slapcraft recipes.
/:cl:


maker-of-nightmares:
So to be clear the militia weapons(which are intended to be made cheaply) are no longer craftable or just not the slapcraft method?

PR #1532, opened 2026-05-03
Lots of crafting tweaks, mostly just cleaning up

Jamdrawers:
## About The Pull Request

All weapon and armor crafting recipe entry included a line detailing what material it needed. Instead of listing "required material: steel bar" 60 times under each of the 60 steel recipes, lists "required material: steel bar" under the baseline steel recipe which then inherits the trait automatically.

Also, changes the recipe for Chain Gloves so that they produce two pairs per bar, giving them a reason to smith them when plate gauntlets had the same material cost.

Changes the recipe for a Full Chain Coif to cost two bars, giving one a reason to smith a regular coif or a regular chain mantle, to which the full coif was objectively superior and had the same material cost.

Deletes the 'brayette' item - nearly completely identical to the chain skirt, which is also completely mechanically useless

Increases the skill requirements on magical transmutation so that only Expert Alchemists can convert stone to coal or coal to iron or more. I was tempted to just delete the thing entirely but leaving it to a narrow pool of roles is more merciful. 
Who could have thought that the ability to literally spin mud into gold would be bad for the economy

Compresses the entries for Carpet crafting and equalises the craft difficulties

Changes the filepaths on the recently added outfit sets so that they're properly subtypes of existing clothes instead of sitting outside the menu

Crafting Pillows takes Sewing skill

The two different kinds of 'hanging chain' have the same material cost

Wooden fences (like the railings, not the pallisades) are made of sticks instead of logs

Fixes a typo found in the 'iron saber' to make it function properly as a subtype of 'sabre', and gives it a crafting recipe

A Heater Shield require more than 'novice' carpentry if you're using carpentry and not blacksmithing/engineering.

Also makes chain gloves smelt into Irondust, 2-3 of which depending on your Smelting skill can be now crafted to make iron ore, which can be then smelted back into iron/steel. Pretty certain you can't cheese this for better returns or infinite iron since I this doesn't give more dust than it would cost to make the gloves

## Testing Evidence

Compiles, loads fine
image

image
image
image

## Why It's Good For The Game

Mostly makes things way cleaner and quicker for future development.
Gives reasons to smith chain gloves and chain coifs/mantles
Cuts down a little on free money from transmutation 
Makes things a bit more intuitive

## Changelog

:cl:
balance: chain gloves are cheaper to smith at two pairs per bar. Full chain coifs are more expensive to smith at two bars per full coif.
Alchemical Transmutation Recipes such as stone-to-coal and coal-to-iron have a higher skill requirement
iron dust can be crafted into iron ore with smelting skill
refactor: cleans up and compresses code for lots of crafting recipes
/:cl:


gayshark:
Itzk here, my man, your update removed all blessed armor with the billion ingots, can only make blessed rings.

Jamdrawers:
Should be fixed now

code/modules/roguetown/roguejobs/blacksmith/anvil_recipes/armor.dm
```diff
@@ -992,105 +756,87 @@
...
 	additional_items = list(/obj/item/clothing/suit/roguetown/armor/plate/fluted/ornate, /obj/item/natural/hide/cured, /obj/item/natural/hide/cured)
 	created_item = /obj/item/clothing/suit/roguetown/armor/plate/full/fluted/ornate
 
 /datum/anvil_recipe/armor/blessedsilver/psyfullplatealt
 	name = "Psydonic Full-Plate, Hauberked (+Psydonic Hauberk, +2 Blessed Silver, +2 Cured Leather)"
-	req_bar = /obj/item/ingot/silverblessed
 	additional_items = list(/obj/item/clothing/suit/roguetown/armor/chainmail/hauberk/ornate, /obj/item/ingot/silverblessed, /obj/item/ingot/silverblessed, /obj/item/natural/hide/cured, /obj/item/natural/hide/cured)
 	created_item = /obj/item/clothing/suit/roguetown/armor/plate/full/fluted/ornate
 
 // BLESSED SILVER BULLION
```
john-pawson, line 799: To be speciefic, all armors under this part with the "/inq" at the end should use blessed silver bullions for their required and additional ingots, and any above without it should use normal blessed sliver, the way you've written it, there's two versions of the same recipes that BOTH require bullions, but only the top ones use normal ones as additional items (And the cuirass below this comment uses a normal ingot when it should use a bullion)

gayshark:
For some reason, the gold plates only make one when it clearly says two? Look into that?

Kroto-Ker:
I NEED my scadu TM'd.

PR #1539, opened 2026-05-04
Lots of back-end area tweaks, moving the 'speakeasy', other tweaks

Jamdrawers:
## About The Pull Request

Another map update. Wide but shallow. Lots and lots of reordering and cleaning up of different areatypes, including deleting a lot of SS13 areatypes (and also commenting out some old and unused code that used those areatypes)

Most noticeable changes would be moving the 'secret' underground 'speakeasy' down a z-level, swapping it with the 'hobo village' beneath, as well as nudging the inquisitorial manor left a few squares.

Cuts down on the number of yeoman flagpoles since it's generally not relevant information outside of hightown (or indeed in a dungeon on the other side of the map)

-Deadite NPCs might spawn in the cursed mire. Careful down there
-A handful of areas have some more unique music 
-The wizard-tower-inferno dungeon plays the correct music now

## Testing Evidence

image
image

## Why It's Good For The Game

The file cleanup is filecleanup. Less clutter is good, and I understand area types enough that I understand how to make one when I need one.

The 'speakeasy' being two z-levels down will gently reduce everyone by the tailor, church steps and watch barracks from overhearing you have sex and breaking the door down just to gawk, though I do think it was always cooler that it was suspended above the hobo pits.

Inquisitorial Manor should be a little less easy to break into. Inquisition should be breaking in to other places more than they're defending against being broken into.

## Changelog

:cl:
map: moved the 'secret' underground 'speakeasy' down a z-level, swapping it with the 'hobo village' beneath.
refactor: reorganised a lot of areatypes, and deleted a lot of unused areatypes
/:cl:


IronDragoon:
Completely disagree with removing flag poles outside of town. That is the entire point of them, to show you who IS in the town so you know if it's worth traveling there or not

halbdog:
Echoing above comment - I get cutting flags in the bog, but saying they are "not relevant information outside of hightown" is crazy when the best use of the feature is knowing what stuff is in hightown while in lowtown. What flags does this remove?

PR #1546, opened 2026-05-04
[Performance] Ghost latency reduction

IronDragoon:
## About The Pull Request

_I want to preface this with saying this is far from my area of expertise as a developer. But with recent issues of super high (over %50) TDI at peak times, I wanted to try and make some changes to hopefully make the game run more smoothly. I'll be making each change in a seperate PR, so issues can be more easily hotfixed/adjusted._ 

This change should make so that player ghosts are no longer featured in hostile target gathering logic, which will hopefully improve latency. I also made the call to make players be returned to lobby upon far travel. Spectate is still fully there, this just means you need to choose to spectate as opposed to it being the default, as currently we end up with a lot of ghosts around far travel points. It may not make MUCH difference, but it might help!

Also adds some admin debug verbs to help assess the impact of this change.

## Testing Evidence

Runs and seems to function as intended. As always, impossible to test on live without TM

## Why It's Good For The Game

TDI bad

## Changelog

:cl:
fix: Hostile mobs should now ignore ghost mobs in target logic, and players will need to choose to spectate as opposed to it being the default on death/fartravel. This will hopefully reduce TDI somewhat (but other changes will be needed too)
/:cl:


The-Sun-In-Splendour:
> I also made the call to make players be returned to lobby upon death/far travel.

I need some clarification? Am I not allowed to ghost and observe around if I'm dead and revival is taking a while? Will dying send me to lobby instantly? Wouldn't that make DNR the default, then?

IronDragoon:
> > I also made the call to make players be returned to lobby upon death/far travel.
> 
> I need some clarification? Am I not allowed to ghost and observe around if I'm dead and revival is taking a while? Will dying send me to lobby instantly? Wouldn't that make DNR the default, then?

Hello,

Apologies this was worded badly as I wasn't thinking at the time. No, death works as it does now, apologies!

PR #1547, opened 2026-05-04
[Performance Test, Test Merge Only!] Temporarily Disable Spectator Mode & Filter AI Wake-Up 

hugmeorelse:
## About The Pull Request

**Do not merge.**

This is purely for testing and observation as of this moment.

This PR is different from #1546. In fact, some aspects of that PR covers some areas that this one does not in regards to CanAttack. However, I haven't really reviewed in full.

For one, in this PR, we temporarily disable spectate to begin with. After a discussion with a maintainer, this PR along with the temperature/weather PR (I asked Radiant if they could make turning it on and off a configurable option) both seem to coincidence around the time when time dilation ballooned significantly, i.e. late March. Although it may not be either of those PRs necessarily.

Why are we temporarily disabling spectate? This is to gauge performance differences properly. A significant number of players did notice an increase in performance when observe was disabled. However, not all of its features were fully disabled. This is to see the full effects of the spectator removal on performance.

We will then see if there is any appreciable effects of ghost protection as well separately later on performance. **There is some wonkiness to it as noted by Ketrai in the original PR.**

This adds filtering for only mob/living when it comes to reviewing over eligible clients and in regards to what triggers them waking up. This is to see if it prevents ghosts and aghosts from causing lag spikes as demonstrated from video examples shown by a staff member.

**Some notes:**
~~for(var/mob/living/M in grid.client_contents) may be an entirely too big of a list at times. I don't know how DM works under the hood when filtered through a loop like this. Suggestion was made to make a special category for non-ghost mobs? Either ways, could be reverted back if it's a concern.~~

After further review, I believe that spatial grid search is already being called to begin with anyways. Furthermore, from what I also understand, it only checks for any mobs with a client. So, the list should not be that big. So, it should be fine. Again, someone can double-check me. But I believe this should be good to go.

Ghost harddel is a major issue as well in regards to performance. Not really touched here.

## Testing Evidence
compile

## Why It's Good For The Game

Testing.

## Changelog

**Not to be merged!**

### DIFF
--- a/code/_onclick/observer.dm
+++ b/code/_onclick/observer.dm
@@ -54,8 +54,8 @@
 
 // Oh by the way this didn't work with old click code which is why clicking shit didn't spam you
 /atom/proc/attack_ghost(mob/dead/observer/user)
-	if(is_hidden_from_ghosts(src, user))
-		return TRUE
+	// if(is_hidden_from_ghosts(src, user))
+	// 	return TRUE
 	if(SEND_SIGNAL(src, COMSIG_ATOM_ATTACK_GHOST, user) & COMPONENT_NO_ATTACK_HAND)
 		return TRUE
 	if(user.client)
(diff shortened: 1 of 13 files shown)


hugmeorelse:
Doesn't seem to be the main candidate. Performance gains is visible, but it's not the main driver. Can be unTMed for now.

PR #1551, opened 2026-05-05
Bedbreaker Improvements

WolfOrion:
## About The Pull Request
Adds a check for both players' extreme preference before attempting a fracture.

Adds less extreme alternative effects for those with the preference disabled, as well as effects for targeting the face/mouth.

## Testing Evidence
The check passing text was just for test purposes.

image

image

image

image

## Why It's Good For The Game

A flat 10% chance to just fracture your partner with every single force level if you forget you have STRONG intent on can completely derail a scene so you can bandage and heal before bleeding out, and it's happened to me by accident, twice, so I want it to be locked behind extreme preferences. At the same time, bedbreaker still doing _something_ would be nice, so I've given it a new status effect, as well as added effects for if someone is targeting the mouth instead of just your groin.

## Changelog

:cl:
add: Added a new status effect, as well as face-specific effects
qol: made bedbreaker check for both force level and extreme preferences
balance: rebalanced something
image: added a new sprite for the quivering status effect
/:cl:


Jamdrawers:
A 'sore' debuff that makes you walk funny should have always the result instead of litterally shattering your actual bones, was a long time coming!

PR #1555, opened 2026-05-05
[AI] Small tweaks for potential performance gains

IronDragoon:
## About The Pull Request

The mobs subsystem is our primary source of server load right now, often exceeding even 70ms of block time (over 140% of the tick budget). 

This PR contains two relatively minor changes which SHOULDN'T have any noticeable impact in terms of gameplay, though I will preface this PR with saying I am NOT an engine coder and this may just be me entirely misunderstanding some of the logic here. Maints please take a close look and if you think i'm being stupid, I probably am. I just figured i'd have a crack at some things which seem to be fixable.

Firstly we now throttle target acquisition logic for hostile mobs who do not have a target. This lets us not need to rely on large numbers of 'idle' mobs calling the _hearers()_ proc EVERY tick, which is relatively intensive. Instead mobs who are idle will now do a cheap client-list scan most of the time, then performing a hearers check less often. If a mob is attacked or loses its target the timer resets so it should still react immediately. I cleared up a few logical changes like a mob losing a target and not immediately finding it, I also didn't realize how retaliate mobs were different but now do, so this should not affect gameplay noticeably, but hopefully help to start reduce load on the mob controller

Secondly simple_animals will now skip the life() proc for 2/3 of ticks IF idle. They were using this despite it not (afaik) being needed, Now it only fires alongside the status effect proc every 3 ticks. This should just help also reduce processing that is functionally entirely unneeded which making no change to gameplay.

## Testing Evidence

Tested on local, where mob behavior seems unchanged.  

## Why It's Good For The Game

TDI bad

## Changelog

:cl:
fix: boring AI proc stuff, should help with TDI
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/hostile.dm
+++ b/code/modules/mob/living/simple_animal/hostile/hostile.dm
@@ -60,6 +60,7 @@
 	var/next_full_seek
 	var/lazy_seek_interval = 4
 	var/full_seek_interval = 20
+	var/use_lazy_target_scan = TRUE
 
 	cmode = 1
 	setparrytime = 30
@@ -102,7 +103,7 @@
 	if(target)
 		possible_targets = ListTargets()
 	else
-		if(search_objects)
+		if(search_objects || !use_lazy_target_scan)
 			possible_targets = ListTargets()
 		else if(world.time >= next_seek)
 			next_seek = world.time + lazy_seek_interval
--- a/code/modules/mob/living/simple_animal/hostile/hostile.dm
+++ b/code/modules/mob/living/simple_animal/hostile/hostile.dm
@@ -57,6 +57,9 @@
 	var/retreat_health
 
 	var/next_seek
+	var/next_full_seek
+	var/lazy_seek_interval = 4
+	var/full_seek_interval = 20
 
 	cmode = 1
 	setparrytime = 30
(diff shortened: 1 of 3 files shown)


noelle-lavenza:
there's probably a much easier way to reduce how much time is spent on this:
```dm
/mob/living/simple_animal/hostile/proc/ListTargets() //Step 1, find out what we can see
	if(!search_objects)
		. = hearers(vision_range, targets_from) - src //Remove self, so we don't suicide

		var/static/hostile_machines = typecacheof(list())

		for(var/HM in typecache_filter_list(range(vision_range, targets_from), hostile_machines))
			if(can_see(targets_from, HM, vision_range))
				. += HM
	else
		. = oview(vision_range, targets_from)
```
there are no hostile machines defined, so the `typecache_filter_list()` is a very very expensive no-op. either remove that block or add a `length(hostile_machines)` check around the for loop.

IronDragoon:
> there's probably a much easier way to reduce how much time is spent on this:
> 
> ```dm
> /mob/living/simple_animal/hostile/proc/ListTargets() //Step 1, find out what we can see
> 	if(!search_objects)
> 		. = hearers(vision_range, targets_from) - src //Remove self, so we don't suicide
> 
> 		var/static/hostile_machines = typecacheof(list())
> 
> 		for(var/HM in typecache_filter_list(range(vision_range, targets_from), hostile_machines))
> 			if(can_see(targets_from, HM, vision_range))
> 				. += HM
> 	else
> 		. = oview(vision_range, targets_from)
> ```
> 
> there are no hostile machines defined, so the `typecache_filter_list()` is a very very expensive no-op. either remove that block or add a `length(hostile_machines)` check around the for loop.

agreed, i'll adjust

PR #1556, opened 2026-05-05
Eoran bow and some bow refactoring

Jamdrawers:
## About The Pull Request

I was commissioned to sprite a little Eoran-themed bow a while ago. Here it is, finally, as an option for Eoran templars! Might make little heart-shaped arrows later on down the line.

Also while I was at it, I noticed that there's a bow called 'bow/short' which is used in like 2 obscure places, while the '/bow' which is called 'simple shortbow' and is used in a million places actually has the exact same stats as the recurve bow other than dealing 1 less damage when whacking someone with it in melee

I applied the 'shortbow' stats to the base bow, so now it will function a little different to the recurve.

## Testing Evidence

Normobow2
harpbow1
image

## Why It's Good For The Game

I think they look cute.

## Changelog

:cl:
add: Eoran Harp Bow as a ranged option for the Eoran Templar!
balance: shortbows now function as shortbows
refactor: adjusted how bows are laid out
/:cl:


ophaq:
Can I play the bow like a harp?

Jamdrawers:
> Can I play the bow like a harp?

...If you can figure out how to code that and then code that, then yes!

IgiariValkyr:
Harp bows should notch three arrows cause three strings. :)

Inveldy:
I COMMISSIONED THIS BY THE WAY YOU CAN THANK ME FOR THIS

Broill:
One thing, I think the stat that should be changed shouldn't be str for int but str for per, since bows rely on per more than anything, something like +2 would probably help them.

wikimody:
Can you change the recurse bow to work like the shortbow instead? Changing the normal bow affects every sinlge bowman class, and is straight up nerf in most situations (they do less damage). Recursive menwhile is somthing you need to go out of your way to get, and i cant recal a class that has it roundstart, so trade off becomes intentional

Jamdrawers:
> Can you change the recurse bow to work like the shortbow instead? Changing the normal bow affects every sinlge bowman class, and is straight up nerf in most situations (they do less damage). Recursive menwhile is somthing you need to go out of your way to get, and i cant recal a class that has it roundstart, so trade off becomes intentional

The recurve bow is adjusted and should now work the way the recurve bow has always worked.

Lots of roles start off with a recurve bow; Wardens start with a subtype of a recurve bow, MAA skirmisher gets a recurve bow. Plaguebearers, veterans, knights, squires, desert riders...

wikimody:
> Lots of roles start off with a recurve bow Wardens start with a subtype of a recurve bow, MAA skirmisher gets a recurve bow. Plaguebearers, veterans, knights, squires, desert riders...

I feel like nerfing the damage of everyone starting with short bow shoudl still be more promiment in the PR, but fair enough, from quick check most advanced combat classes start with recursive

Jamdrawers:
> One thing, I think the stat that should be changed shouldn't be str for int but str for per, since bows rely on per more than anything, something like +2 would probably help them.

I assume you mean speed instead of int

I knocked a point off of con and end for perception - same weighted total at the end, but less hardy more snipey

PR #1572, opened 2026-05-07
Dying with critical weakness forces an emote

Xander3359:
## About The Pull Request
Makes dying with critical weakness force a death gurgle emote.

## Testing Evidence
image
Chopped head with the sword and it played the emote when I died 👍 

## Why It's Good For The Game
It's super weird that when you die with critical weakness it's just completely silent.
This makes it where you at least force out an emote upon death.

## Changelog

:cl:
add: Dying with critical weakness forces out an emote so that people can recognize you are no longer alive.
/:cl:


The-Sun-In-Splendour:
I've noticed that when I crit a goblin's head it will do the deathgasp, and when I crit its chest afterwards it deathgasps again even though it's dead.

PR #1579, opened 2026-05-08
Sewrepair fix

Jamdrawers:
## About The Pull Request

Fixes some things that got set to be repairable with sewing that weren't supposed to be

## Testing Evidence

Went down the list in the editor and all the things that should be listed 'sewrepair 1' were '1' and '0' were '0'

image

## Why It's Good For The Game

Fix

## Changelog

:cl:
fix: fixes a bunch of things being repairable with sewing that weren't supposed to be
/:cl:

### DIFF
--- a/code/modules/clothing/rogueclothes/armor/_armor.dm
+++ b/code/modules/clothing/rogueclothes/armor/_armor.dm
@@ -27,3 +27,4 @@
 	flags_inv = HIDEBOOB|HIDECROTCH
 	grid_width = 64
 	grid_height = 64
+	sewrepair = FALSE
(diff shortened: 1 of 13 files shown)


Jamdrawers:
> Please double-check everything. This broke smelting on some items completelly. Chain gloves for example

That was due to an unrelated change that made chain gloves cheaper to smith by making two pairs for one bar, so that there might be a reason to smith chain gloves instead of plate gloves (being cheaper). That said, latest change should make them smelt into iron dust which can be crafted into iron ore (which can then be smelted into steel bars again)

PR #1582, opened 2026-05-08
gives necromancer wretch the "Raise Deadite" spell

The-Sun-In-Splendour:
## About The Pull Request
what it says on the tin

## Testing Evidence
works on my localhost

## Why It's Good For The Game
Personally I think it doesn't make sense that the necromancer is only able to raise long-dead corpses (skeletons) and not the people he's just killed as deadites.

It hurts a bit that I have to use arcyne potential just to feel like I can afford something that I should thematically have by default already, when I really wish I could use my virtue to drink murky water and raw meat so it fits more my character. 3 points isn't a lot but on a class that only has 16 spellpoints it ends up hurting.

Also lets the necromancer let people who've died to him actually play the game (by biting and scratching people as a deadite). Not only does it fit the zizite/necromancer fantasy but deadites are pretty weak and usually end up getting killed/restrained for rotcuring, which means you don't have to do the awkward "I drop this corpse in a public spot so they can be revived because I don't want to full RR this guy". Win win for both sides.

## Changelog

:cl:
balance: necromancer wretch gets raise deadite
/:cl:

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/wretch/necromancer.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/wretch/necromancer.dm
@@ -66,6 +66,7 @@
 		H.mind.AddSpell(new /obj/effect/proc_holder/spell/invoked/tame_undead)
 		H.mind.AddSpell(new /obj/effect/proc_holder/spell/invoked/raise_undead_formation/necromancer)
 		H.mind.AddSpell(new /obj/effect/proc_holder/spell/invoked/raise_undead_guard)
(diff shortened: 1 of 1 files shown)


kimesik:
The spell is kinda janky but the change itself sounds good to me.

Xander3359:
bro you have 16 points on this class.

The-Sun-In-Splendour:
> bro you have 16 points on this class.

3 out of 16 is 18.75%
Fireball is 6 points. thats over 33%.
16 points is not as much as you think it is.

PR #1590, opened 2026-05-10
Improves Hemophage Vice

DelineFortune:
## About The Pull Request

Adds a new vampire bite debuff: **The Kiss**. (hemophage-only, for now)
Victims affected while being fed on receive a short ecstasy debuff (the debuff lowers -2 CON INT WIL temporarily for 30 seconds)
Adds **Vital Essence**, hemophage-only healing/stamina snowflake reagent they can get from drinking blood. (As hemophages cannot obtain buffs from food or even drink water from any sources to restore blood)
The effectivity is 50% of mana/normal red potions, wound (not hp) healing is a bit weaker than strong red.

## Testing Evidence

image

## Why It's Good For The Game

At the moment, the bite does not provide any effect (the trait you get says otherwise), while the vice itself is already fairly dangerous due to the loss of bonuses you can get from food and blood restoration from drinking water. 
On top of that, attempting to feed on someone can easily get you killed by other players.
Feeding saturation is also extremely slow, causing bearers of the curse to spend a significant portion of the round suffering from hunger and thirst.

## Changelog

:cl:
add: Hemophage debuffs their targets on bite, but their bite also restores their health+stamina now
/:cl:


The-Sun-In-Splendour:
Isn't hemophage and vampirism literally broken in that it prevents any and all healing? Wther by sleeping, bloodheal, or red vials?

ophaq:
Oh good, I won't have to make it so hemophages can actually drink blood from bloodletting now when it gets TM'd. Although, I should probably make it so they bleed dirty blood at some point so they're not drinking their own blood with this... unless they don't have blood to begin with?

DatFrostys:
> Isn't hemophage and vampirism literally broken in that it prevents any and all healing? Wther by sleeping, bloodheal, or red vials?

Nooo it's a bug that may happen when you have fire resistance and get hit by fire/holy fire damage. Only way I know to fix it is remove the fire resistance, set yourself on fire and extinguish yourself

IronDragoon:
> Oh good, I won't have to make it so hemophages can actually drink blood from bloodletting now when it gets TM'd. Although, I should probably make it so they bleed dirty blood at some point so they're not drinking their own blood with this... unless they don't have blood to begin with?

can we drink from bloodletting? 

I was thinking about this PR, maybe we can add a craftable bottle/container of some kind that can be held to a wound to hold SOME blood for use later? My main problem with hemophage is that you have no way to keep 'supplied', OTHER than carting around some tied up mob, which feels a little silly (and is just annoying). Maybe if we add a reagent you can buy from merchants or is made with alchemy, when poured into a bottle it 'lines' it with a material that will stop blood curdling? And could add a 'rot' time to it, like with food? So we can at least, with some work initially, have a way to have SOME reliable supply when doing long 'adventures' or whatever else.

I would happily code/sprite a new bottle for this...

If NOT, at least making us take longer to get hungry as currently it's so quick we have to just keep a blood bag around at ALL times

HalfAwake691:
At this point should Hemophage be buffed more and just made into a virtue? Because having the ability to drink someone's blood to restore a shit ton of mana bar seems like not a vice to me.

PR #1608, opened 2026-05-12
Ports Amulets from AP. 

JustMeTheIInd:
## About The Pull Request
Ports like five amulets from AP. Because they are just a better codebase apparently
-Golden Z-cross
-Bronze amulet of Graggar
-Amulet of Matthios 
-Amulet of Baotha
-Amulet of Graggar
-silver amulet of Ten

It makes the three Ascendant amulets, and a amulet of the ten which was *kinda* ported selectable via the loadout menu. 

The PR also gives the Prelate a silver amulet of the Ten instead of the imbued amulet of Astrata, and also gives the martyr the amulet of the SEE. (Because there are two versions now! the amulet of the See, and of the Ten, the former is not gettable via normal means iirc and the latter is selectable in the menu. Why? Lore. Yeah. I care about that, I know, I know. I am a loser. Woe.

It also adds several on_mob icons, like the astrata amulet, or the dendor one. Because we didn't have those. But we did have the one for Matthios, despite it not being an actual item. 

The Bronze amulet and the Golden Z-cross are not selectable, because the Z-cross is made out of **gold** which makes it kinda valuable. And the bronze amulet has like a 25 sellprice so it's also out of loadout. Womp-womp.

## Testing Evidence

Zrzut ekranu 2026-05-12 201346
Zrzut ekranu 2026-05-12 201621
Zrzut ekranu 2026-05-12 201532

## Why It's Good For The Game

More customization options. I don't know why you would willingly wear the amulet that says "I am okay with murder and cannibalism" while out in the town... but I mean you have the option now. Go, play out your role. 

## Changelog

:cl:
add: Amulet of Matthios
add: Amulet of Graggar
add: Amulet of Baotha
add: Silver amulet of Ten
add: Amulet of Ten
add: Golden Z-cross
add: Bronze amulet of Graggar.


JustMeTheIInd:
Oh right. I also made like a lil change because it wouldn't compile if I didn't remove 'ms' from one value in like, a sheet file for whatever reason? I discarded the change before pushing it, though, but if it fails it'll be prolly because of that.

PR #1615, opened 2026-05-13
Artificer Improvements & More!

AllCorruptingWyrm:
## About The Pull Request

I spoke to several people about this and I personally agree with them so I've made this PR after reaching out to artificer players and some guildies too.

The following PR has been made after several requests to make Artificer more bearable and encourage interaction between artificers and the Mages Guild in an effort to make crafts within the Guild a not-so-miserable experience. Furthermore, this pr as a whole touches (and even adds) several recipes and MORE eventually once ideas start rolling in here. The PR also allows those with engineering skills to make items that relate to ensure firearms transition from being a gimmick to something uniquq and distinct.

This PR does NOT give anyone a firearms skill; It purely makes certain items craftable.

## Testing Evidence

When its no longer a draft pr.

## Why It's Good For The Game

Firearms, in my opinion, will be one way to differentiate ourselves from other servers; they were here during RW1 and they should be here as well and not feel so gimped after rebase, where firearms had little to no care.. Along with that, with imps (and other mobs from the Mage loop) dropping essentia on death, this allows artificers to interact with mages and alchemists for essentias so they can make their stuff.

## Changelog

:cl:
add: all essentia's to be made by artificers
add: bronze ingot recipe (might need to be removed, idk)
add: EMPTY lead ball pouch recipe
add:  powder flask recipe
add: ramrod recipe
changed: Pyro bolt recipes
changed: replace fyritus with fire essentia in several recipes
changed: Arcane meld (a power source for artificer armor) makeable by  1 pure essentia and 4 manabloom dusts
TODO add: mageloop mobs dropping essentias
TODO add: Craftable spectacles
remove: Pick requirement from grapples
remove: Adjusts the artificer plate to be moderately less expensive
/:cl:

If I missed anything in the change log, mention it but this is a heavy draft.

### DIFF
--- a/code/modules/roguetown/roguecrafting/engineering.dm
+++ b/code/modules/roguetown/roguecrafting/engineering.dm
@@ -42,6 +42,60 @@
 	skillcraft = /datum/skill/craft/engineering
 	craftdiff = 2
 
+/datum/crafting_recipe/roguetown/engineering/ore2bronze
+	name = "Ore to Bronze"
+	result = /obj/item/ingot/bronze
+	reqs = list(/obj/item/rogueore/copper = 1) // 1 to 1 conversion rate
+	verbage_simple = "engineer"
+	verbage = "engineers"
+	skillcraft = /datum/skill/craft/engineering
+	craftdiff = 4 // Bread and butter for artificers
+
+/datum/crafting_recipe/roguetown/engineering/ore2firedust
+	name = "Ore to Fire Essentia"
+	result = /obj/item/alch/firedust
+	reqs = list(/obj/item/rogueore/copper = 1) // 1 to 1 conversion rate
+	verbage_simple = "engineer"
+	verbage = "engineers"
+	skillcraft = /datum/skill/craft/engineering
+	craftdiff = 4 // Bread and butter for artificers
+
+/datum/crafting_recipe/roguetown/engineering/ore2airdust
+	name = "Ore to Air Essentia"
+	result = /obj/item/alch/airdust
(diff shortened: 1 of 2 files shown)


maker-of-nightmares:
There is already a recipe for bronze using 1 tin and 3 copper. It requires no fancy materials to craft a bronze smelter either which artificers all have by default in the basement. The recipe for the bronze smelter is similar to the forge requiring only stone and coal.

Having the engineering recipe exist is already a bit of a conflict on that and having it be 4 times less expensive and easier to boot is even bigger because after all why would anyone use the existing recipe if that exists?

Chief301:
The bronze smelter has a lot of non-functional, deprecated recipes - IE sticks not turning into needles. It also has something like a TEN PERCENT CHANCE TO BLOW UP IN YOUR FACE, destroying the smelter, the item you were trying to smelt, and anything in a two tile radius. The only way to stabilize it is with an essence of lumber which, even with the drops being fixed, is STILL a pain in the ass to get.

The bronze ingot recipe in question lets artificers convert bronze to essentia, and vice versa.

anconfuzedrock:
Artificer is plate is so ludicrously powerful that I hesitate to ever mention it because gamers might get access to it. No point hiding it now, It's fucking full plate armor that doesn't require armor skill, so dodge experts, mages, and berserkers can use it. Don't make it cheaper. Don't make the melds easier to get either. The rest isn't important.

Jamdrawers:
I think it'd be bad and booooring to replace using smelters with a simple menu crafting recipe for making bronze.

RadiantFlash7:
Stale and abandoned.

PR #1620, opened 2026-05-13
Noc Acolyte Fixes

ophaq:
## About The Pull Request
Adjusted arcyne affinity to be better and less jank. 
- 50 minute CD -> 40 minute CD.
- Devotion adjustment 200 -> 250.
- Doesn't go on cooldown if you hit cancel.
- Removes the category from the chosen list once you cast it again. Actually lets you pick a new category instead of not being able to once you cast it again. This means you can get more spells to your arsenal.
- You can't get all the buff spells though. Also, once you have selected all 3 categories (if you choose to), the spell will remove itself from your spell bar to make more room.

I also fixed the ritual's text when it ends to not look like it cut off mid sentence.

Noc acolytes / missionaries qualify for the weapon enchanting refresh now. It was really annoying they would get the spell, but it wouldn't detect them as a magic user to auto-refresh the enchant for their weapon.

## Testing Evidence
arcyneaffnity

## Why It's Good For The Game
It's more fun to play and feels better. Fixes are good, too.

## Changelog
:cl:
fix: Arcyne affinity is less jank. Lower CD, slightly higher devotion cost. Lets you pick multiple categories over time now.
fix: Noc acolytes & missionaries qualify for the weapon enchant spell's auto refresh.
fix: Noc ritual text when it ends. It used to just look like it cut off mid sentence.
/:cl:

### DIFF
--- a/code/datums/status_effects/rogue/roguebuff.dm
+++ b/code/datums/status_effects/rogue/roguebuff.dm
@@ -1257,7 +1257,7 @@
 
 /datum/status_effect/buff/moonlightdance/on_remove()
 	. = ..()
-	to_chat(owner, span_warning("Noc's silver leaves my"))
+	to_chat(owner, span_warning("Noc's silver light leaves my sight."))
 	REMOVE_TRAIT(owner, TRAIT_DARKVISION, MAGIC_TRAIT)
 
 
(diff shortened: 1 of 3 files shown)


Venithefox:
there is currently a bug with the spell that makes it completely impossible to use arcyne affinity more than once. tested on noccite heretic spy and lunacy embracer if you could fix it.

wait no nevermind this I'm stupid and didn't read it properly. you already fixed it in this. you're a legend ophaq

PR #1636, opened 2026-05-15
nukes staves from orbit

zeroisthebiggay:
## About The Pull Request

dumb ass skill used by only `one` weapon class that's literally just a blunt spear and isn't different enough from polearms or is common enough to warrant a whole ass skill, scarcely given out to anyone so they can actually use it

kinda tampers with the identity of some classes because it means they can drop their staff and pick up a spear, but who gaf spears are mid anyways

all quarterstaves now use polearm skill, and the classes that used staves have had their polearms skill bumped up to their original staves skill value to compensate - most classes had their polearms skill on par with their staves skill anyway lmao

only significant balance consequence of this is that heresiarch has expert polearms now but who gaf

## Testing Evidence

image

## Why It's Good For The Game

dumb bloat skill gone, players are now more inclined to use staves as nonlethal spear alternatives

## Changelog

:cl:
del: staves skill, now everyone who used it previously has had their polearms skill upped to compensate
balance: quarterstaves are now polearms
/:cl:

### DIFF
--- a/code/datums/migrants/migrant_waves/grenzel_noble_roles.dm
+++ b/code/datums/migrants/migrant_waves/grenzel_noble_roles.dm
@@ -179,7 +179,6 @@
 	subclass_skills = list(
 		/datum/skill/combat/wrestling = SKILL_LEVEL_EXPERT,
 		/datum/skill/combat/unarmed = SKILL_LEVEL_JOURNEYMAN,
-		/datum/skill/combat/staves = SKILL_LEVEL_JOURNEYMAN,
 		/datum/skill/combat/polearms = SKILL_LEVEL_JOURNEYMAN,
 		/datum/skill/misc/reading = SKILL_LEVEL_LEGENDARY,
 		/datum/skill/craft/alchemy = SKILL_LEVEL_JOURNEYMAN,
--- a/code/datums/migrants/migrant_waves/otavan_noble_roles.dm
+++ b/code/datums/migrants/migrant_waves/otavan_noble_roles.dm
@@ -228,8 +228,7 @@
 		STATKEY_WIL = 2,
 	)
 	subclass_skills = list(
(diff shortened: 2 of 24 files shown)


Jamdrawers:
It'd be nice to make it so that when you use a weapon you can use the highest of a number of applicable skills, like if you swung a wooden staff it could use the highest of either your polearms *or* your staves skill, so that roles-that-aren't-meant-to-be-swinging-halberds don't have reason to buy a naginata every round, but a polearm user can still make use of a stick easily enough.

Lots of cases like that, where like a Bardiche couldn't take advantage of someone's Axe skill or a pole-mace couldn't make use of someone's polearm skill or like a pitchfork or scythe couldn't make use of someone's polearm skill because they're only keyed to farming.

Would need some proper coding though, alas.

Jamdrawers:
That said, imo the roles that had staves skill shouldn't have much polearm skill (and shouldn't have had much stave skill) to begin with for the most part, you don't need high skill to get 90% parry against mobs and, as well as being a walking stick and a status symbol that's really what they're there for.

zeroisthebiggay:
> That said, imo the roles that had staves skill shouldn't have much polearm skill (and shouldn't have had much stave skill) to begin with for the most part, you don't need high skill to get 90% parry against mobs and, as well as being a walking stick and a status symbol that's really what they're there for.

if you want a walking stick then the wdefense of the staff should be lowered instead
as it is now things like wooden staves which are actually used as walking sticks ingame are considered polearms but quarterstaves are considered staves

zeroisthebiggay:
https://github.com/Rotwood-Vale/Ratwood-2.0/pull/842

i just realized that this pr also exists and is even testmerged but it's so old
also my pr somehow touches one more line i can't be assed to find out which

PR #1637, opened 2026-05-16
Athleticism & Knights

petwir1:
## About The Pull Request

Upgrades the Journeyman Athletics on Foot Knight, Heavy Knight, and Mounted Knight to Expert. 

## Testing Evidence

Psydon strike me down if this doesn't work, this is a very very simple change.

## Why It's Good For The Game

Probably due to some powercreep long ago, these two knight subclasses have ended up with lower athleticism than 75% of the garrison. These are supposed to be elite warriors and guardians of the Crown... why would they not be trained to a higher or equivalent standard.

In-game, it'll just make it so that the constant running around and fighting that knights have to do will not be as draining. This will make them a bit better of fighters, but it shouldn't really make a massive impact.

## Changelog

:cl:
balance: gave heavy knight, foot knight, and mounted knight expert athletics
/:cl:

### DIFF
--- a/code/modules/jobs/job_types/roguetown/garrison/knight/knight_foot.dm
+++ b/code/modules/jobs/job_types/roguetown/garrison/knight/knight_foot.dm
@@ -22,7 +22,7 @@
 		/datum/skill/combat/unarmed = SKILL_LEVEL_JOURNEYMAN,
 		/datum/skill/misc/climbing = SKILL_LEVEL_JOURNEYMAN,
 		/datum/skill/misc/reading = SKILL_LEVEL_APPRENTICE,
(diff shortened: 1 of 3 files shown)


zeroisthebiggay:
> I decided not to change the mounted knight to encourage mount usage.

it's the least popular knight by a mile throw them a bone
i am extremely biased

petwir1:
> > I decided not to change the mounted knight to encourage mount usage.
> 
> it's the least popular knight by a mile throw them a bone i am extremely biased

I generally agree with that, it won't make a huge difference and mounted knights almost never get played.

Randumb-Phantomb:
This is a tiny change regardless, there's only two knights, so like... Why not give them this.

PR #1639, opened 2026-05-16
Makes rontz ring (and scomstone)(and the crown) unusable (while restrained and/or incapacitated)

JustMeTheIInd:
## About The Pull Request

This PR makes it so that if a user is either incapacitated (stuned, paralysed, asleep (I think..)) or restrained by (chain, in hand) they can no longer use the rontz ring/scomstone/crown.

## Testing Evidence

Zrzut ekranu 2026-05-16 005331

## Why It's Good For The Game

Consequences... good? Yes I do indeed believe that if you fucked up big-time as antag you should not actually be permitted to just PSY-BLAST every last one of your non-erping buddies your exact coordinates, how many people are guarding you, what they are armed with, and also what they had for breakfast that morning. You fucked up. You are chained, you are at the mercy of your captor.

Don't want to not be able to call your buddies for backup? Do want to be able to talk with them? Play better. Don't get caught. 

## Changelog

:cl:
balance: Rontz Ring is now no longer usable when the user is restrained or incapacitated
/:cl:

### DIFF
--- a/code/modules/roguetown/roguemachine/scomm/rontzring.dm
+++ b/code/modules/roguetown/roguemachine/scomm/rontzring.dm
@@ -85,6 +85,9 @@
 	if(user.restrained() || user.incapacitated())
 		to_chat(user, span_warning("I cannot use this while restrained or incapacitated!"))
 		return
+	if(!get_location_accessible(user, BODY_ZONE_PRECISE_MOUTH, grabs = TRUE))
+		to_chat(user, span_warning("My mouth is covered!"))
+		return
 	user.changeNext_move(CLICK_CD_INTENTCAP)
 	var/input_text = input(user, "Enter your message:", "Message")
 	if(input_text)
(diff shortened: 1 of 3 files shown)


vincentthemongrel-commits:
Finally

Hmeister-real:
This should be applied to Scom, Hound and Kingstones too for consistencies sakes

JustMeTheIInd:
> This should be applied to Scom, Hound and Kingstones too for consistencies sakes

I thought that the other stones already weren't usable when restrained, but I will double-check and add it if it isn't

Xander3359:
undraft when merge ready

JustMeTheIInd:
Ought to work now!

Foxentail:
So. Im all for not letting bandits use the rings when chained up and restrained. HOWEVER.

After playing several rounds with this there's a few things I think should be changed.  First of all make it so we can still speak into it if masked. Im pretty sure the crownstone allows you to, why shouldnt bandits? Also with this change you cant even use it while wearing a padded coif without taking it off. Smh.

Second. Bandits should be able to mute the ring even if chained. Garrison is already abusing the shit out of this by carrying around bandits or using their dead bodies to listen to the ring. I also think the ring should crumble to dust on death, like how cursed collars fall off when you die.

PR #1643, opened 2026-05-16
[Performance] Overlay churn reduction

IronDragoon:
## About The Pull Request

Here’s a concise, accurate summary of the changes:

Overlay subsystem (overlays.dm):
Skips unnecessary overlay work: If you try to add, remove, or clear overlays and the result would be the same as before (no overlays added/removed, or already empty), it does nothing—avoiding wasted processing and overlay rebuilds.

Overlay normalization is more efficient: Only normalizes overlay lists when needed, reducing list mutation and allocation.

Vis overlay subsystem (vis_overlays.dm):
Rotation logic is faster: When an atom changes direction, it now loops through overlays once, rotating only what’s needed, instead of building temporary lists for set operations. Skips work if there are no overlays to process.

The third change adds overlay state caching to the inventory HUD overlay logic. Now, overlays are only rebuilt and applied if the held item, overlay color, or slot actually changes, instead of every update. This reduces unnecessary overlay churn and improves server performance without affecting gameplay or visuals.

Net effect: These changes reduce server CPU usage from overlays by avoiding redundant overlay list rebuilds and unnecessary overlay state changes, especially in high-churn scenarios. No game logic or visual behavior is changed—only performance is improved.

## Testing Evidence

Builds and runs, I did a bunch of UI churn to try and check as many overlay states as possible and noticed nothing broken.

## Why It's Good For The Game

TDI bad

## Changelog

:cl:
fix: bunch of small changes to overlays to hopefully reduce unneeded refreshes which contribute greatly to server overhead
/:cl:


code/controllers/subsystem/overlays.dm
```diff
@@ -41,28 +41,35 @@ SUBSYSTEM_DEF(overlays)
...
+			for(var/raw_overlay in build_overlays)
+				if(!raw_overlay)
+					continue
+				if(istext(raw_overlay))
+					// This is too expensive to run normally but running it during CI is a good test
+					// if(PERFORM_ALL_TESTS(focus_only/invalid_overlays))
+					// 	if(!icon_exists(icon, raw_overlay))
+					// 		var/icon_file = "[icon]" || "Unknown Generated Icon"
+					// 		stack_trace("Invalid overlay: Icon object '[icon_file]' [REF(icon)] used in '[src]' [type] is missing icon state [raw_overlay].")
+					// 		continue
```
Xander3359: just delete commented out useless code.

Xander3359 (changes requested):

PR #1650, opened 2026-05-17
makes the zizo froggemund not a snowflakey skin

zeroisthebiggay:
## About The Pull Request

when you get dark armaments, you get to choose between the froggemund which has neck protection and the barbute which has a visor

## Testing Evidence

image

image

## Why It's Good For The Game

it's fun

## Changelog

:cl:
add: when using the dark armaments ritual, you get to choose between the zizo froggemund which has neck protection and the zizo barbute which has a visor yippee yay
/:cl:


Wadelicious:
Will Ancient Champion be able to choose between the two? I don't see code to make it have the ability to do.

zeroisthebiggay:
> Will Ancient Champion be able to choose between the two? I don't see code to make it have the ability to do.

i can add that

Wadelicious:
> > Will Ancient Champion be able to choose between the two? I don't see code to make it have the ability to do.
> 
> i can add that

yeah that'll be like cool because the AC's current weakness is definitely the neck and i think all AC mains would love you for that and give you nice goodnight kisses on the forehead :)

PR #1651, opened 2026-05-17
Fixes Sconce North Wall Crafting

Trexdude:
## About The Pull Request
Previously, trying to craft a sconce on a wall north of the player would result in the sconce floating in front of the wall where the player was standing, rather than putting the icon of the sconce on the wall itself. This PR makes it so crafting the sconce causes it to sit on the wall itself, as (probably) originally intended.

## Testing Evidence
Screenshot 2026-05-17 112839
Screenshot 2026-05-18 104859
Screenshot 2026-05-18 104912

## Why It's Good For The Game
Hopefully opens up more possibilities for sconce crafting, now that one of the four directions they can be crafted in isn't bugged.

## Changelog
🆑 
fix: crafting sconces on a north wall now properly puts the sconce on the wall itself
/:cl:

### DIFF
--- a/code/game/objects/lighting/rogue_fires.dm
+++ b/code/game/objects/lighting/rogue_fires.dm
@@ -366,6 +366,10 @@
 	QDEL_NULL(torchy)
 	on = FALSE
 	set_light(0)
+	pixel_x = 0
+	pixel_y = 0
+	if(dirin == SOUTH)
+		pixel_y = 32
 	update_icon()
 
 	..(dirin, user)


Xander3359:
Can you show screenshot in the game after crafting it?

Trexdude:
> Can you show screenshot in the game after crafting it?

Of course! Here's the sconces as crafted currently, both with and without torches:

Screenshot 2026-05-18 104859
Screenshot 2026-05-18 104912

PR #1662, opened 2026-05-18
differentiates addiction vice debuffs

zeroisthebiggay:
## About The Pull Request

this gives every addiction vice its own associated debuff and differences in stat debuffs, meaning stacking addiction vices will reduce your stats separately and be able to be cleared separately

image

## Testing Evidence

https://github.com/user-attachments/assets/6cabf1a7-f030-4fc5-967b-876bdc7e8303

here's me purging myself of a bunch of addiction vices, excuse the fact that you don't get to see alcoholic be removed - it takes a wihle for the game to consider you drunk enough so it cleared shortly after the video ended
also I didn't clear sadist in the video but I assure you it works

## Why It's Good For The Game

makes people who stack addictions actually reap the consequences of doing so

## Changelog

:cl:
add: addiction vices each have their own separate debuffs, meaning you can stack and clear them individually
balance: across the board made addictions take longer to give you negative effects by about fifty percent
/:cl:


halbdog:
I really like splitting up the debuffs so they can stack, and how they all give different stat debuffs. But the change to the addiction timers is pretty radical, especially for the 3 substance addictions. They're easy and cheap to sate, but will now only proc twice in a round, while the harder to sate masochist + sadist flaws get a shorter cooldown? 
Still a good PR that I support, though. I just enjoy hardship.

PR #1678, opened 2026-05-19
Padded + Studded Hood to Tailor Silverface | Slot Adjustments | Mask FoV Bugfix

ophaq:
## About The Pull Request
- Adds the padded and studded hood to the tailor silverface.
- Adds studded hood to the merchant goldface.
- Allows the hoods to be worn on the neck slot (not just the head).

Bonus:
- Fixes a FoV issue with certain masks like the jester, oni and kitsune ones not updating the FoV at times when they were removed. Now it force updates your FoV to the smaller cone when taking it off instead of ruining your FoV like you had a mask still on.

## Testing Evidence
hoodsgoldface

## Why It's Good For The Game
Currently you could only buy the padded hood from the merchant. The tailor was lacking them. I also made them wearable on neck since it was like that on SR. If necessary, I can bump their prices up more but they're pretty expensive at the moment anyway.

Also, the FoV bug has been bugging me for a while so I got that fixed.

## Changelog
:cl:
add: Adds the padded and studded hood to the tailor silverface.
add: Adds studded hood to the merchant goldface.
add: Allows the hoods to be worn on the neck slot (not just the head).
fix: Fixes a FoV issue with certain masks like the jester, oni and kitsune ones not updating the FoV at times when they were removed.
/:cl:

### DIFF
--- a/code/modules/cargo/packsrogue/merchant/armor/merch_armor_iron.dm
+++ b/code/modules/cargo/packsrogue/merchant/armor/merch_armor_iron.dm
@@ -31,6 +31,11 @@
 	cost = 25
 	contains = list(/obj/item/clothing/neck/roguetown/bevor/iron)
 
+/datum/supply_pack/rogue/armor_iron/studded_leather_hood
+	name = "Studded Leather Hood"
+	cost = 45
+	contains = list(/obj/item/clothing/head/roguetown/helmet/leather/armorhood/advanced)
+
 /datum/supply_pack/rogue/armor_iron/breastplate_iron
 	name = "Breastplate"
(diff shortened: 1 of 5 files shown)


zeroisthebiggay:
> Allows the hoods to be worn on the face

you shouldn't do that
neck slot is fine but don't let people wear full head armor on their face slot

ophaq:
> > Allows the hoods to be worn on the face
> 
> you shouldn't do that neck slot is fine but don't let people wear full head armor on their face slot

Fair point. I only did it cause the other type of hood could but on second thought, I'll take it off the face.

PR #1682, opened 2026-05-20
RG Commission #01 P1 Redflash

RiceGateau:
## About The Pull Request

Ports: https://github.com/Azure-Peak/Azure-Peak/pull/4880

## Testing Evidence

image

## Why It's Good For The Game

Flash can be agitating to some users so being able to toggle it is a good feature.

## Changelog

Beware, changelog is likely not accurate as this is a split PR so I had to revert changes.

:cl:
add: Added more things
code: changed some code
/:cl:


Xander3359 (changes requested):
Instead of being an `if(flash) -> show_flash` this really should just be its own proc which will automatically do the pref check followed by the flash.

Genevesh:
Rather than removing the flash (literally your character experiencing pain which might affect them in combat), make it DARKEN the screen instead of flashing red?

It's genuinely debilitating sometimes to not be able to see anything for a split second, and it often matters in a fight and can cost you a free hit or two. This should absolutely not be possible to completely opt out of being briefly blinded.

Xander3359:
> Rather than removing the flash (literally your character experiencing pain which might affect them in combat), make it DARKEN the screen instead of flashing red?
> 
> It's genuinely debilitating sometimes to not be able to see anything for a split second, and it often matters in a fight and can cost you a free hit or two. This should absolutely not be possible to completely opt out of being briefly blinded.

I do somewhat agree on this. Being able to remove the red pain overlay is a bit immersion breaking? Not sure if that's the right term but I am quite unsure on this tbh.

RiceGateau:
> I do somewhat agree on this. Being able to remove the red pain overlay is a bit immersion breaking? Not sure if that's the right term but I am quite unsure on this tbh.

I mean, the main purpose is to give the client the option of removing it, either due to preference or health concerns as the flash is bright and fast. I believe it is not a balance concern as not having the flash on is debilitating compared to another client who does. This is how the flash acts as a warning sign for impeding crit or if you were just injured. If you toggle it off, you will have to rely on audio queues or the heart on the left HUD.

There is nearly no benefit for having it off as the time to see during a fight by not getting flashed is negated by having to glance to your left. Not to mention the flash time is already pretty fast so you aren't being blinded for a long enough duration to affect combat a considerable amount.

zoktiik:
being handled in #2102

RadiantFlash7:
> being handled in #2102

Thank you. I'll close this PR while we refer to the other one.

PR #1685, opened 2026-05-20
Nature's Reclamation event

Jamdrawers:
## About The Pull Request

Hightown is all covered in grass and leaves. Perhaps a spell from a mad and powerful druid?

## Testing Evidence
image
image
image
image

## Why It's Good For The Game

Would be fun for a one-round event, who knows


Maromar15910:
This is actually pretty, I'd even sue for the Dendor storyteller to have a chance to do something like this

Jamdrawers:
> This is actually pretty, I'd even sue for the Dendor storyteller to have a chance to do something like this

The hand-placed stuff would be quite a hassle to maintain, and mapgen runs *before* storytellers can be picked/voted, though I guess I could imagine someone who can code good making a special mapgen that spawns after the storyteller gets picked.

ophaq:
Druids will be seething as every towner fights back to chop down every tree and bush in sight.

maker-of-nightmares:
> > This is actually pretty, I'd even sue for the Dendor storyteller to have a chance to do something like this
> 
> The hand-placed stuff would be quite a hassle to maintain, and mapgen runs _before_ storytellers can be picked/voted, though I guess I could imagine someone who can code good making a special mapgen that spawns after the storyteller gets picked.

Provided nothing about the map majorly changes you can just have the structures be built after the storyteller is selected. It's an absolute pain in the ass to do though and the easier way would just have the story teller vote happen before round end so it can load the map which has the grass instead of Rockhill normally

MarioTheShiba:
WELCOME TO THE JUNGLE

WE GOT…i forgot how the song goes.

maker-of-nightmares:
> WELCOME TO THE JUNGLE
> 
> WE GOT…i forgot how the song goes.

... everything you want, honey, we know the names
We are the people that can find whatever you may need
If you got the money, honey, we got your disease

PR #1687, opened 2026-05-20
Massively simplify hostile mob ListTargets

noelle-lavenza:
## About The Pull Request
Removes dead code in ListTargets that was making hostile mobs far more expensive than necessary.

Small note for maints: ~~instead of `hearers() - src` and `oview()` it now uses `viewers() - src` and `view() - src`. This should make object-targeting mobs able to target things on their own tile, and make sure mobs (by default) take visibility into account. If that's not desirable I can change it back to hearers().~~ It now uses `hearers() - src` and `view() - src`. I'm not sure if that `view()` needs to be `dview()` or not.

also removes a useless optimization and associated bookkeeping. skipping newplayers is pointless because those should never be on the map, and observers will fail either the invisibility or isliving checks.

## Testing Evidence
A volf and skeleton both targeted and attacked me normally. They didn't notice me while I was sneaking with max sneak skill, until I stopped sneaking and I was attacked. Woo.

## Why It's Good For The Game

I mentioned it on #1555. This removes dead code and simplifies the checks a lot. This PR means we will likely be able to check much more frequently with less lag, or at least less lag at the current check frequency, so there will be more responsiveness.

### DIFF
--- a/code/modules/mob/living/simple_animal/friendly/snake.dm
+++ b/code/modules/mob/living/simple_animal/friendly/snake.dm
@@ -2,6 +2,7 @@
 	var/poison_per_bite = 0
 	var/poison_type = /datum/reagent/toxin
 
+// todo: envenomated component that registerse to COMSIG_HOSTILE_ATTACKINGTARGET?
 /mob/living/simple_animal/hostile/retaliate/poison/AttackingTarget()
 	. = ..()
 	if(. && isliving(target))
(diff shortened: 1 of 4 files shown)


code/modules/mob/living/simple_animal/friendly/snake.dm
```diff
@@ -2,6 +2,7 @@
 	var/poison_per_bite = 0
 	var/poison_type = /datum/reagent/toxin
 
+// todo: envenomated component that registerse to COMSIG_HOSTILE_ATTACKINGTARGET?
```
Xander3359, line 5: todo?

code/modules/mob/living/simple_animal/hostile/hostile.dm
```diff
@@ -169,16 +169,13 @@
...
-	if(!search_objects)
-		. = hearers(vision_range, targets_from) - src //Remove self, so we don't suicide
-
-		var/static/hostile_machines = typecacheof(list())
-
-		for(var/HM in typecache_filter_list(range(vision_range, targets_from), hostile_machines))
-			if(can_see(targets_from, HM, vision_range))
-				. += HM
+	if(search_objects)
+		. = view(vision_range, targets_from) // todo: does this need to be dview to ensure they have darkvision like with hearers()?
```
Xander3359, line 205: todo? :)

Xander3359 (changes requested):
Still being worked on?

code/modules/mob/living/simple_animal/hostile/hostile.dm
noelle-lavenza: the todos are for future work in other PRs, if it were for this PR i would've put it in the PR body

PR #1693, opened 2026-05-21
makes baotha marked better interact with other vices

zeroisthebiggay:
## About The Pull Request

right now baotha marked gives you the nymphomaniac vice but also it also fucking breaks you if you take it alongside nymphomaniac
this makes it so baotha marked gives you its own thing so it can stack, allowing you to either be double nymphomanic or be a baotha marked without being a nymphomaniac, and simply having unnatural impulses that need to be sated

## Testing Evidence

image

## Why It's Good For The Game

bug fixes and improvements

## Changelog

:cl:
add: baotha marked applies a unique addiction vice if you aren't baothan yourself, sated same as nymphomaniac but without letting you do things like identify other nymphomaniacs with it
del: baotha mark giving nymphomaniac
fix: baotha mark being taken with nymphomaniac breaking you
/:cl:


balthazargreen63-design:
God pr 
We can take both without the need to jork it 24/7

zeroisthebiggay:
> God pr We can take both without the need to jork it 24/7

until then just take nymphomaniac in your first vice slot instead of baotha marked
baotha marked checks for if you already have nymphomaniac before it gives you it and if it's in your first vice slot it sees that you don't have nymphomaniac so it gives you another before the game actually gives you the nymphomaniac from your second vice slot
this gives you two instances of it which for some reason makes it not sateable

PR #1705, opened 2026-05-22
Wretched Toiler again;

AllCorruptingWyrm:
## About The Pull Request

i'm going to copy paste what was said here as to why it was good and why it should happen:

I would actually interject and say support classes for wretches are quite useful. It enables wretches, especially when paired up with others, to do things that they normally cannot. For example, fortify the fort at the Bog, so it no longer becomes a death trap for the antagonists. You can actually have a focal point now that the wretches can use as a sort of forward operating base.

I think you're really quite missing out by disabling it by not realizing that there's some real potential here, especially with some of the players that we do have.

I strongly encourage you reconsider, as this class did have a positive impact on other servers. Furthermore, this is a roleplay server after all, not a TDM, so not every antagonist has to have insane combat power in order to create conflict or support it.

After all, Sawbones, for instance, have great utility and importance for Bandits. You wouldn't remove it, simply because it's not narrowly focused solely on combat.

## Testing Evidence

standby.

## Why It's Good For The Game

See above

## Changelog

:cl:
add: Ports Wretched Toilers from Scarlet Reach. Rejoice, minions! For you are WRETCHED! And you must TOIL!!
/:cl:

### DIFF
--- a/code/game/objects/items/rogueitems/gems.dm
+++ b/code/game/objects/items/rogueitems/gems.dm
@@ -82,6 +82,12 @@
 		slapcraft_recipes = slapcraft_recipe_list,\
 	)
 
+/obj/item/roguegem/houndgem
+	name = "houndstone gem"
+	icon_state = "topaz_cut"
+	sellprice = 0
+	desc = "A component to make a Houndstone, it looks like a topaz."
+
 /obj/item/roguegem/yellow
 	name = "toper"
 	icon_state = "topaz_cut"
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/wretch/wretchedtoiler.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/wretch/wretchedtoiler.dm
@@ -1,0 +1,119 @@
+/datum/advclass/wretch/vigilante
+	name = "Masked Lunatic"
+	tutorial = "You were a disenfranchised pauper, sickened by the rampant corruption of the garrison - or perhaps, just a crazed vagrant in a costume? Whether those brutalized 'thieves' were justified in their acts is up to YOU to decide, not them! You specialize in utilizing your various gadgets and thrown projectiles to dote out JUSTICE, however you see it fit."
+	allowed_sexes = list(MALE, FEMALE)
+	allowed_races = RACES_ALL_KINDS
+	outfit = /datum/outfit/job/roguetown/wretch/vigilante
+	cmode_music = 'sound/music/combatmaniac.ogg'
+	class_select_category = CLASS_CAT_ROGUE
+	category_tags = list(CTAG_WRETCH)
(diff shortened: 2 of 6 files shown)


Venithefox:
THE RETURN OF THE KING

also isn't the point of test merging to see how something interacts with the game? just TM it if we're so uncertain

AllCorruptingWyrm:
also i literally played with it live

It has mage spells and a choice between them??? So enjoy.

IronDragoon:
I think we should absolutely TM this. Having a wretch that is geared to supporting rather than straight combat surely can only be good imo

Venithefox:
Bump

TamedAchilles:
So a good point was raised a while ago that honestly was fair -

I don't think in the long run this should be slotted into the Wretch slots for say, but in either the Towner or Sidefolk slots or even its own new job slot category. The reasoning behind this is is that as a support role, you really wouldn't want to take away from the actual slots that the Wretch already get since they only have 10 at the moment. This promotes the idea of what the Wretched Toiler should be while not outright diminishing the Wretch population by intermingling Support roles in what're suppose to be antagonist roles.

I've liked the idea though of a Toiler or Wretched Towner for quite some time, providing assistance to those who follow the Ascendents. It provides more RP on the side of the evil do'ers while also, if played right, giving the Garrison and the Town something to nearly fear as there'd be basically a second village/town built around Ascendent worship and their followers. Built by the hands of those who want to promote those ideologies every round as opposed to just starting with a fancy pre-built town.

Frick Astrata. Gosh darn Tyrant making everyone cozy and shit.

RasbeRissa:
As the person who ported it previously, I just want this test merged already man it's been like three months since my PR was opened.

AllCorruptingWyrm:
There will be discussion about it

Venithefox:
LETS, FUCKING, GOOOOO

vincentthemongrel-commits:
Wretches are just becoming Brigands v2. Team work? Not exactly bad out of context, but saying that wretches will be able to reinforce at the bog fort says enough.

wikimody:
I woudl ask for small change, to be exact to instead of giving the straight up call to slaughter, a very specyfic god miracle with non garogorites shoudnt be able to use, give them a reskin of said miracle with similar function. It has precedence since call to slaughter itself is a reskin of call to arms ravox has. It just feels a bit weird for a random toiler to be multigod caster

wikimody:
Also sidnote, undocumented houndstone crafting recepture change. You might have messed up a commit from difrent PR.
Its the Lets try it comit

LostLittleMeowMeow:
this is awesome. a few more soulful additions like this and we will have saved ratwood

UntoldTactics:
This class is cool but I've seen a lot of people gaming it with how ridiculous vicious mockery is + a few miracles and magic spells. A good combo is eyebite and going abyssor patron for the drowning exhaust thing.

May need to do some fucky stuff to make them combat neutered; the people playing it half the time aren't really being support. They're being powergamers.

PR #1709, opened 2026-05-23
Pontifex Pontifixes (And giving it the Vizier treatment)

KeberosNihilus:
## About The Pull Request

Pontifex, at some point in the distant past, got nerfed despite it being only pretty okay so far as unarmed classes go. I'd even argue that, as it is on live, it's worse than _Arcane Trickster_, an Adventurer class. This brings it... Not quite up to par with Hierophant, but at least to a point I'd call it a solid pick again.

Also just fixes some common complaints I've heard about it, cleaned up the grammar here and there, and made its unique spell(s) have more clear descriptions about their requirements.

## Testing Evidence

proof01

(I know I should probably have more but like... It's really tedious loading up every subclass, especially when I had to fix some spaghetti)

## Why It's Good For The Game

The Bugfixes: I really shouldn't have to explain why this is a good thing. Easier to read, less obfuscation for noobs, etc etc.

The Balance Stuff: Pontifex, on live, isn't... Great? Your punch dagger having only 20 strikes before it vanishes is certainly... _A_ decision of all time. Really, I _should_ make it a variant of Conjure Weapon but having be a literally unremovable weapon is a novel feature so... It just lasts a LOT of hits. It shouldn't just vanish mid-fight anymore, at least. Also, you get 4 spellpoints with your T1 spells. Really tight budget even for cantrips, but it's similar to what they used to have. Due to the Paths giving them spells they didn't originally have access to, the SP is low.

Pontifex as it is, and was, is a weird one in that it's a dedicated unarmed build but doesn't have the stats normally needed to make it work and instead has really useful magic. Less good on their own but amazing as bodyguards, rogues, and in teamfights. So I decided to lean into that with their Paths.

Path of War: Air Blade is criminally underused as it's pretty much exclusive to the Adventurer Spellsword, despite being a side/downgrade to Arcane Bolt. And it's really hard to go wrong with Enchant Weapon, though you can't apply it to your conjured weapon... It also has a 1.3 damage mod intent so it doesn't really need it. Goes amazing well with a pair of knucks, however.

Path of Control: Ensnare was moved here and you get Greater Forcewall. This lets Pontifexes who choose this path dictate the flow of battle. This is just returns you to the original intent of Pontifex where you're an annoying shit who stops people from running and/or chasing you.

Path of Shadows: The name says it all, but it embraces the theme of you using the secrets of Noc to your advantage. Lesser Knock because sneakthief, invis because shadows, and Blindess... Because shadows. Are you noticing a theme? If you use it right, this is probably the strongest Path out of the three though with your Apprentice (or Journeyman, with AP) Arcane, Blindness shouldn't be _too_ oppressive. That and your Blind is on a whopping 25 second cooldown, which should further make the other Paths more appealing.

Path of Survival: Because you think Pontifex is good as it is, but could just use a _little_ more in terms of skills to shine. Mostly a sovl pick, but still valid due to the +1 athletics and +2 swimming/medicine. Diagnosis is there for the rare Pontifex bodyguard main.

## Changelog

:cl:
add: Arcyne Blindness - A variant of Blindness that scales on Arcane skill. Longer CD, requires a Naledian psycross to use.
fix: Adjusted some text regarding the Pontifex. Makes some of what they need clearer and it should read better.
balance: Pontifex now has 4 Spellpoints but still only T1 spells.
balance: Pontifex no longer has Ensnare by default.
balance: Pontifex now has 4 'Paths' (subclasses) to choose from.
add: Path of War - Air Blade (sidegrade to Arcane Bolt) and Enchant Weapon. Enchant Weapon can't be used on your Push Dagger, but is deadly if applied to a pair of knuckles.
add: Path of Control - Ensnare and Greater Forcewall. Heavy focus on CC and battlefield control.
add: Path of Shadows - Lesser Knock, Invisibility, and Arcane Blindness. Heavy focus on trickery and shadows.
add: Path of Survival - Increases your Medicine, Cooking, Athletics, and Swimming skills. Also gives you Secular Diagnosis. Sovl and lets you bodyguard or run around like a punch-dagger wielding maniac better.
/:cl:

### DIFF
--- a/code/modules/jobs/job_types/roguetown/mercenaries/classes/warscholar.dm
+++ b/code/modules/jobs/job_types/roguetown/mercenaries/classes/warscholar.dm
@@ -93,7 +93,7 @@
 
 /datum/advclass/mercenary/warscholar/pontifex
 	name = "Naledi Pontifex"
-	tutorial = "You are a Naledi Pontifex, a warrior trained into a hybridized style of movement-controlling magic and hand-to-hand combat. Though your abilities in magical fields are lacking, you are far more dangerous than other magi in a straight fight. You manifest your calm, practiced skill into a killing intent that takes the shape of an arcyne blade."
+	tutorial = "You are a Naledi Pontifex, a warrior trained into a hybridized style of movement-controlling magic and hand-to-hand combat. Your chosen Path determines your specialization, though you'll never match another mage in pure magical power. Instead, you manifest an imitation of a shard of PSYDON's blade and rely on trickery and battlefield control."
 	outfit = /datum/outfit/job/roguetown/mercenary/warscholar_pontifex
 	subclass_languages = list(/datum/language/celestial)
 	traits_applied = list(TRAIT_DODGEEXPERT, TRAIT_CIVILIZEDBARBARIAN, TRAIT_ARCYNE_T1)
@@ -117,7 +117,7 @@
 		/datum/skill/misc/stealing = SKILL_LEVEL_JOURNEYMAN,
 		/datum/skill/misc/lockpicking = SKILL_LEVEL_JOURNEYMAN,
 	)
-	subclass_spellpoints = 0 // Override inheritance lol
+	subclass_spellpoints = 4 // Override inheritance
 
 /datum/outfit/job/roguetown/mercenary/warscholar_pontifex
 	var/detailcolor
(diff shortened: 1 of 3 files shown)


Jamdrawers:
> It's really tedious loading up every subclass

To speed up this sort of thing you can change the map to 'Roguetest (no other z)' and reboot the world, should load in a few seconds

KeberosNihilus:
I... Need to keep that in mind for the future so setup is quicker. But it's more I get a really weird lagspike that tries to crash the game when I far travel. It happens on live too, not just a localhost. Nobody else has the issue and clearing my cache hasn't worked. Just makes trying several loadouts/classes in a row really tedious.

DelineFortune:
image

>>the creator of the spell says not to give it to mages
>>gives it to mages anyway

KeberosNihilus:
> the creator of the spell says not to give it to mages
> gives it to mages anyway

1) It primarily says to not be _available_ to everyone.
2) The phrasing implies it wasn't intended to be _permanently_ unavailable. Presumably awaiting a 'TODO' that never got followed up on.

In any case, I acknowledge that the Path containing is potentially the strongest one there. However, for the sake of clarity, allow me to spell out how, exactly, blindness works as well as my change.

Blindness' strength is determined by the associated skill. Normally this is 'Miracle,' and the classes that have access to Blindness range from 2 - 5 miracle skill. Maximum of 6 with devotee. The length is [SKILL] * 4 seconds, with a minimum of 4.

For most Nocites (Acolytes/Missionaries) this means the duration is going to be 4 * 4, or 16 seconds. It has a cooldown of 15 seconds.

However, Arcane Blindness has been changed to Arcane and is tied to a Mercenary exclusive class that is capped at 2 Arcane. 3 with Arcane Potential, which doesn't even give them Spellpoints as they have a combat trait. This means, for the Pontifex, the duration for the VAST majority of players is going to be 2 * 4, or 8 seconds. On top of this, the cooldown is extended from 15 to _25_ seconds.

Additionally, it requires you to have your Naledian Psycross. If you lose it, these are nearly impossible to replace. That's it. You simply _cannot_ cast this spell anymore. You are _hard_ countered by a pickpocket.

On top of this, taking the Path of Shadow excludes you from taking CC options, extra combat power, or general skill improvements. You're also getting two Cantrips (Invisibility / Lesser Knock) instead of T2/T3 spells.

What does this mean? The Path of Shadows is _really_ good for 1v1 fights, lackluster in teamfights, middling for bodyguard work (the most common work mercs find), and fairly useful for a more rogue/spy playstyle.

The-Sun-In-Splendour:
> I... Need to keep that in mind for the future so setup is quicker. But it's more I get a really weird lagspike that tries to crash the game when I far travel. It happens on live too, not just a localhost. Nobody else has the issue and clearing my cache hasn't worked. Just makes trying several loadouts/classes in a row really tedious.

just write "lobby" in your command bar and press enter twice. Your body will stay there but your slot will re-open. makes testing reallyf ast.

Anyways please do not give them blindness. Noc blindness is the most cancerous spell in this game. It can instantly win fights on its own. If you really want to give them some sort of blindness consider making a subtype of eyebite that's less overtly evil.

PR #1713, opened 2026-05-23
Spurt! Spurt! | Multiple Loads Per Creampie

EvenInDeathIStillServe:
## About The Pull Request

makes it so that, based on your cum volume, you get more spurts out (up to 3) when finishing internally. applies to oral, vaginal and anal.

each spurt rolls for impregnation like a separate creampie. the extra spurts do not consume a cum charge.

also made your load storage also be affected by ball size. also made it so that the minimum load number is 3 with normal sized balls instead of 2, like it used to be. Small balls are 2.

## Testing Evidence

image

image

## Why It's Good For The Game

big cum should be big

## Changelog

:cl:
add: Creampies will have extra spurts based on semen volume.
balance: Minimum load count restored to 3 with normal size balls. Load count is now modified by ball size, too.
/:cl:

### DIFF
--- a/code/datums/sexcon/sex_actions/force/force_blowjob.dm
+++ b/code/datums/sexcon/sex_actions/force/force_blowjob.dm
@@ -48,7 +48,9 @@
 	user.sexcon.perform_sex_action(user, 2, 4, TRUE)
 	if(user.sexcon.check_active_ejaculation())
 		user.visible_message(span_love("[user] cums into [target]'s throat!"))
-		user.sexcon.cum_into(oral = TRUE, splashed_user = target) // give facial status effect for the target, considering this was rough throat sex
+		for(var/i = 1; i <= user.sexcon.get_load_bursts(); i++)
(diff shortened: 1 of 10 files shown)


IronDragoon:
peak, continue doing the Lord's work

Inveldy:
Does this increase the speed you fill a container?

EvenInDeathIStillServe:
> Does this increase the speed you fill a container?

no, babe. this is only for internal orgasms

Inveldy:
> > Does this increase the speed you fill a container?
> 
> no, babe. this is only for internal orgasms

Could it? That's another desired feature...

UntoldTactics:
It'd be quite heinous if knotting added like a 1.5x/2x modifier to total count and did a 'spurt' every 5-15 seconds up to that count. Realistic function 🤣

EvenInDeathIStillServe:
> Could it? That's another desired feature...

you already have modifiers for cum volume when filling containers

ronfino:
SPURT, SPURT, SPURT..................................

PR #1726, opened 2026-05-25
A few possible skeleton tweaks

Jamdrawers:
## About The Pull Request

I'm not settled on all of these since some might be a bit power-creep-y but some might be good. Throwing things at the wall

1) Buffs wrestling skill across the board from 2 to 3. Keeps it slightly closer in line to the other roles and slightly increases the possibility of taking a living prisoner, but they still have the low constitution and strength so it won't be a big change.
2) Gives roles slightly wider skill variety by giving them a lower skill level with the weapons they don't normally have. They have a baseline competence with whatever they can pick up off the ground.
3) Builder skeleton gets Squire Repair, a hammer, and the iron/copper axe and pick are now ancient-alloy ~~(will have to update this if/when the ancient alloy refactor gets in)~~. They also get Ceramics because why not
~~4) Bulwark starts with Heavy armour instead of medium amour and heavy proficiency. They could already get heavy armour via the ritual, and heavy armor proficiency seemed to be a big chunk of the power budget~~
~~5) Instead of wooden shields they get ancient alloy shields. Maybe it's too strong, or maybe we'd just be too sad to lose out on the shield customisation~~
6) bit more int across the board so it isn't literally 1. Still very low. Might stop random beggars and barbarians from right-click feinting you so much. It's kind of a fun idea but mechanically idk if it really contributes positively.
7) Porting the big-death-explosion from AP for the sapper so they can sacrifice themselves to blow a hole through a wall helm's deep style
8) spear option for the legionaire
9) Refactoring the Decrepit hammer/decrepit pick to be subtypes of new Ancient hammers and Ancient picks which builder skeletons now get

## Testing Evidence

Tricky to test things as a skeleton

## Why It's Good For The Game

Described above really

## Changelog
:cl:
balance: player skeletons WRESTLING skill increased, Weapon Skill Variety increased, Squire Repair and a hammer for skeleton sapper, starting heavy armor for bulwark, int increased...
/:cl:


code/modules/jobs/job_types/roguetown/other/lich_skeleton.dm
```diff
@@ -166,17 +168,18 @@ LICH SKELETONS
 	H.STASPD = 5
 	H.STACON = 11
```
Vizoplez, line 171: Ideally we could -1 STR from these guys to give +2 CON instead, they don't need 13 STR for the weapons they use are typically 1-handed and do not scale that well with STR barring the grand mace, which I have already said my grievances about.

Ideally we'd substitute the grand mace for something else, and the longsword is cracked due to it being able to decap through plate armor in like 5 chops so we don't have to really address that part of reducing the STR(they can use it with 12 anyway).

Only concern now is that they might abandon their shield entirely and go get a minotaur axe to go fragging with the funny limb removal device, but at journeyman it prooobably isn't an issue.

Vizoplez: I don't see any immediately glaring issues otherwise, these guys have pretty low CON across the board and only moderate STR levels, a +1 to wrestling won't turn them into grapple fragmachines.

Needs a full on hyperwar to be tested though, so hopefully one happens when this is TM'd

The-Sun-In-Splendour:
> Builder skeleton gets Squire Repair, a hammer, and the iron/copper axe and pick are now ancient-alloy (will have to update this if/when the ancient alloy refactor gets in). They also get Ceramics because why not

skeletons already get squire repair because they get self sustenance

> Bulwark starts with Heavy armour instead of medium amour and heavy proficiency. They could already get heavy armour via the ritual, and heavy armor proficiency seemed to be a big chunk of the power budget
Instead of wooden shields they get ancient alloy shields. Maybe it's too strong, or maybe we'd just be too sad to lose out on the shield customisation

this is already done by the decrepit refactor pr. Bulwarks wil get ancient plate armor and ancient shields if it gets merged. Legionaires still keep regular wooden shields, though.

> bit more int across the board so it isn't literally 1. Still very low. Might stop random beggars and barbarians from right-click feinting you so much.

soul annihilation. skeletons being stupid is soul

Jamdrawers:
Really ought to get that refactor in

Jamdrawers:
> > Builder skeleton gets Squire Repair, a hammer, and the iron/copper axe and pick are now ancient-alloy (will have to update this if/when the ancient alloy refactor gets in). They also get Ceramics because why not
> 
> skeletons already get squire repair because they get self sustenance

Where do they get the self-sustenance from, in the code? I know it's a thing for wretches but skeleton roles are in their weird separate zones.

maker-of-nightmares:
Could just give the sapper smithing skills. It's not like they aren't already engineering and some of those skills directly correlate to cool buildings. Like the Lich Doom Tower could have been even cooler if we all spawned with smithing/engineering skills as sappers

zeroisthebiggay:
if refactor gets in I plan on porting chunks of this PR https://github.com/Azure-Peak/Azure-Peak/pull/7186 so keep in mind that things like skeletons having their weapon skills shoved around and legionnaires getting cool shields is coming soon after that

zeroisthebiggay:
> Could just give the sapper smithing skills. It's not like they aren't already engineering and some of those skills directly correlate to cool buildings. Like the Lich Doom Tower could have been even cooler if we all spawned with smithing/engineering skills as sappers

done in decrepit refactor!!!!! it'll get merged eventually!!!!!

PR #1702, opened 2026-05-22
Soap adjustments

SneakySlinky:
## About The Pull Request

Does a few things, all largely related to soap.

Introduces a check to make sure the mover is sprinting before causing the slip itself, and the duration of the knockdown has been reduced from 8 seconds to 1.

Cleans up some old redundant code of herbal soap, child of regular soap, and deletes old vestigial code for a dummied out soap that had an improper item path.

...Also adds a period to the examine description for herbal soap. 

~~Edit: Apparently also nerfs Xylixian miracles (and ritual) related to slipping by extension.
I do not weep for Xylixians. Get better material!!~~

Edit 2: *Apparently*, I was mistaken, the Xylixian miracles related to slipping are untouched, actually, so slipsquare and grand slipsquare still make turfs inflict a 10s knockdown, and Ritual of Pratfall still retains its ability to force a knockdown (albeit 2 seconds) on any who step onto the laying down Xylixian in question (I guess this is fine?).
<sup>Woe...</sup>

## Testing Evidence
dreamseeker_dJwoq6RG86
dreamseeker_zaL86oihqU

## Why It's Good For The Game

Soap, while infrequently used, has a nasty punishment of inflicting an **Eight Second** hard knockdown, during which you cannot stand at all, and all you needed to do to inflict this upon someone is have them enter the tile it's on. No check, nothing. You could even grab ahold of someone, drop soap, and pull them into the tile to do so. Needless to say, this invites a very... *slippery* slope in terms of strategizing for combat, where people would just take soap for the sheer advantage it offers in making someone effectively crippled for eight full seconds, suffering reduced dodge and parry chances on the ground.

## Changelog

:cl:
del: Deleted old code referencing an improper component initialization for a dummy item path
del: Removed redundant code from soap/bath item that was copied from soap itself
balance: Adding a sprinting check to soap to slip people
balance: Also by extension nerfs Xylixian miracles + rituals and the associated slip-granting sources
/:cl:

PR #1731, opened 2026-05-26
Fixes Xylix Slip Spells

ophaq:
## About The Pull Request
- Reverts running check from soap adjustment PR.
- Shortens slip time on Xylix spells so it does not stun you forever on the floor like soap did.

## Testing Evidence
slipproof

## Why It's Good For The Game
Reverts running check from soap adjustment PR instead of my previous fix I had here before where I separated slip components. Now I added NO_SLIP_WHEN_WALKING to soap itself. Works as intended where you run and slip but can still walk over it.

Also, I understand it was annoying having to deal with a long stun from the soap's effect after slipping. It's been made to about a second now so that you can get up pretty quickly after slipping from the xylix miracle slipsquares.

## Changelog
:cl:
fix: Fixes Xylix slip miracles not actually slipping properly.
balance: Shortens slip time on Xylix miracles so it does not stun you for a long time on the floor like soap did.
/:cl:

### DIFF
--- a/code/datums/components/slippery.dm
+++ b/code/datums/components/slippery.dm
@@ -16,14 +16,8 @@
 
 /datum/component/slippery/proc/Slip(datum/source, atom/movable/AM)
 	var/mob/victim = AM
-	if(istype(victim) && !victim.is_flying() && !victim.jumping && !victim.zfalling && isnull(victim.throwing) && victim.m_intent == MOVE_INTENT_RUN)
-		if(victim.slip(knockdown_time, parent, lube_flags, paralyze_time, force_drop_items) && callback)
-			callback.Invoke(victim)
-
-/datum/component/slippery/no_run/Slip(datum/source, atom/movable/AM)
-	var/mob/victim = AM
 	if(istype(victim) && !victim.is_flying() && !victim.jumping && !victim.zfalling && isnull(victim.throwing))
-		if(victim.slip(min(knockdown_time, 10), parent, lube_flags, 0, force_drop_items) && callback)
+		if(victim.slip(knockdown_time, parent, lube_flags, paralyze_time, force_drop_items) && callback)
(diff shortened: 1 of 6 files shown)


noelle-lavenza (changes requested):
i don't think this is the correct solution, `handle_slip()` already checks `lube_flags` for `NO_SLIP_WHEN_WALKING`. #1702 was also an incorrect change and should be reverted. it specifies `TURF_WET_LUBE` which does not have `NO_SLIP_WHEN_WALKING`.

basically just revert the running check added in #1702.

ophaq:
> i don't think this is the correct solution, `handle_slip()` already checks `lube_flags` for `NO_SLIP_WHEN_WALKING`. #1702 was also an incorrect change and should be reverted. it specifies `TURF_WET_LUBE` which does not have `NO_SLIP_WHEN_WALKING`.
> 
> basically just revert the running check added in #1702.

Okay call me dumb if I misinterpreted what you said but I...
- Undid my previous changes and cross compared to 1702 for that running check. So now this PR is basically a revert of that PR. Yes, soap once more slips again when walking over it. Is there a different approach to this so we can keep soap from slipping non-running people or is this acceptable for now?
- I reduced the lube floor's knockdown duration.

noelle-lavenza:
> Is there a different approach to this so we can keep soap from slipping non-running people

for soap you can add `NO_SLIP_WHEN_WALKING` to the AddComponent call, like `AddComponent(/datum/component/slippery, 10, NO_SLIP_WHEN_WALKING)`

noelle-lavenza (approved):

SneakySlinky:
I never thought to just change it to `NO_SLIP_WHEN_WALKING` because I'm a dumbass, though largely cause turf related stuff like lube and whatnot is just... not used like anywhere

PR #1736, opened 2026-05-27
Nerf Unholy Grasp (Graggar/Gnoll net)

Shell-Knight:
## About The Pull Request
This is very much a mald PR. And my first PR, so its possible I didnt even do this right. But I think a needed one because this has genuinely stayed like this for way too long.

## Testing Evidence
It works (I think)

## Why It's Good For The Game
Nobody likes getting hit by a single no-cost projectile and then effectively put out of the fight because they have a crippling 5 movespeed debuff.

## Changelog

:cl:
balance: Nerfed Unholy Grasp (Graggar Net)
/:cl:

### DIFF
--- a/modular_azurepeak/code/modules/spells/pantheon/inhumen/graggar.dm
+++ b/modular_azurepeak/code/modules/spells/pantheon/inhumen/graggar.dm
@@ -40,7 +40,7 @@
 	releasedrain = 30
 	chargedrain = 0
 	chargetime = 15
-	recharge_time = 10 SECONDS
+	recharge_time = 20 SECONDS //no more spamming please.
 
 /obj/projectile/magic/unholy_grasp
 	name = "viceral organ net"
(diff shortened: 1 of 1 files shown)


Shell-Knight (commented):
Did I mess up the CL message?

coffeecoffeecoffee:
It doesn't just need a nerf, it needs to actually work like a net does - something you can take a moment to remove if you can get the chance.

The-Sun-In-Splendour:
make it 10 seconds debuff. 5 seconds is genuinely insane. Make it 30 seconds cd instead

2whatever2:
Should probably fix the bug of it not using viscera up if it hasn't been already aswell

coffeecoffeecoffee:
Please god all anybody wants is it for it to work like an actual net and for it to go away if you take a bit to remove it
That's all it needs.

Kroto-Ker:
I like the suggestion of having the spell function as an actual net, making it removable, instead of reducing the status effect duration.

AlucardHellsingChapel:
I am not even sure if this pr changes anything seeing that the 30 sec it had before did not work and you just changed it to 5

PR #1748, opened 2026-05-28
The Balancejakkiest Balancejak: Tremendous single-digit adjustment to broke_free

coffeecoffeecoffee:
## About The Pull Request
After experiencing the changes to grapple, originally I thought to revert https://github.com/Rotwood-Vale/Ratwood-2.0/pull/1297, but after playing around with a bit, this should simply reduce the cooldown to 1 second from the whopping 5 that people would be completely immune to grapples before. Keep in mind that I'm pretty sure broke_free applies to everybody, not the specific grappler, so anybody in a group trying to restrain you cannot grapple you for [however many] seconds. This change means that you still get an opportunity to move out of range or engage in counterplay, but escaping a grapple doesn't mean your opponents have a significant 5+ second period where there's nothing they can do except skullcrack you or murder you if they want to do anything about the situation.

PR commits are a little messy because I'm git illiterate and was wrestling with gitfork. I hope it's okay. Sorry.

## Testing Evidence
Before:

dreamseeker_NlTpFQ61nv

After:

dreamseeker_p3c342JR22

Note that in both instances, the times in actual practice should be noticeably longer due to TIDI.

## Why It's Good For The Game
Grappling serves an important purpose: it's the entire garrison's way to not have to murder you to get you to comply when they have you outmatched. I've been discussing how grappling has felt the past couple days with people and how it feels as though you literally cannot subdue criminals alive anymore without murdering them first and taking them to church, and how it feels like the gameplay has been forcing us to break all the bones in a person's body to get them to submit (as nobody yields, ever, and likely nothing can be done to change that.)

Grappling had already gotten to a point where it is rather weak, breaking out of a grapple was reasonably easy and it puts you in a position to do quite a lot in your own defense - counter-grappling becomes much easier, the grappler literally cannot escape swift pick dagger, it gives you an incredibly powerful free feint opportunity, etc. - and adding 5 seconds of break_free has made it so that when the grappler (inevitably, given the weak state of grappling) fails their first grab...

They have a long window where they can really do nothing but kill you. This sucks.

It turns out that this small change has had fairly significant ramifactions on people's ability to try and take others alive. Giving garrison members, or abducting bandits, a 5 second window (actually noticeably longer, due to time dilation) where they are encouraged to do nothing but wail on your head with blunt weapons or else sit around with their thumbs up their ass encourages hyperlethality, who would've guessed!

_Why 1 second instead of just removing it?_ I actually think the implementation of broke_free helps alleviate one problem and it's instant-grappling by groups and lack of opportunity to escape. By retaining a second of cooldown, it still gives you a moment to break away, maneuver or grapple in return. Just because I think 5 seconds is creating hyperlethal behavior and further encourages the necessity of skullcracking doesn't mean we should return to groups being able to perma-grapple people. All of the reasons in the original PR that implemented broke_free still apply and are good points.

## Changelog

:cl:
balance: Changed grapple cooldown from a whopping 5 seconds to 1 second
/:cl:


zeroisthebiggay:
does this fix the bug with phantom grabs

coffeecoffeecoffee:
> does this fix the bug with phantom grabs

no i have no idea how to fix that i'm really sorry i _have_ stared at the code to try and find it but i do _not_ have the know-how to tackle an actual bug like this i am really sorry i am _not_ a skilled coder

coffeecoffeecoffee:
WHAT.
I didn't mean to do this fuck god damn it um uhh???

coffeecoffeecoffee:
i'm just gonna remake this git is black magic

PR #1752, opened 2026-05-29
Building QoLs, Mining QoLs & Pick Bumping + Auto Woodcutting

ophaq:
## About The Pull Request
- Implemented a partial port and expanded a bit upon AP's pick bumping. Basically, you can just walk into a rock wall to initiate auto dig on it. Requires apprentice mining.
- Added in auto woodcutting.
- Added toggles for both these features in the new fancy toggles menu. Pick bumping is bundled in with auto digging.
- Adds auto stone chiseling. It repeats chiseling for each stone on the turf! Chiseling has always been incredibly click intensive.
- Boulders now inherit the same do_after rock walls get when mining them.
- Removed mason set and carpenter set for commoner. Now it the craftsman set. It gives you the traits that let you get the most out of wood and chiseling like builder/architect has. No more handsaw or chisel since you already start out with those. Slightly nicer handaxe and hammer in trade.

## Testing Evidence
doafterwoodcutandpickbump

## Why It's Good For The Game
Clicking over 100 times every round on trees as a builder gets really old. This helps with the fingers, just as autodig did with mining. Pick bumping is just a further QoL expansion to the mining one. Feel free to tell me your gripes about why this is bad, because I really don't see why it is. There's even an option to toggle them both off now if you want 'immersion' or if you enjoy your finger hurting and mouse getting destroyed after hundreds of hours clicking trees.

Also there was some bloat in the commoner class sets and I figured it would be better to just combine carpentry and masonry into one set since most players are going to be doing both anyway if they're building and it makes this set more appealing than the other two which didn't add much. You already get a handsaw and chisel in your starting gear so these two sets were not really useful unless you wanted spares. Now you have an actually handy one that establishes your intent to be building oriented.

## Changelog
:cl:
qol: Implemented a partial port and expanded a bit upon AP's pick bumping. Basically, you can just walk into a rock wall to initiate auto dig on it. Requires apprentice mining.
qol: Added in auto woodcutting.
add: Added toggles for both these features in the new fancy toggles menu. Pick bumping is bundled in with auto digging.
qol: Chiseling stones will now repeat for each stone on the turf.
qol: Boulders now inherit the same do_after rock walls get when mining them.
add: Removed mason set and carpenter set for commoner. Now it the craftsman set. It gives you the traits that let you get the most out of wood and chiseling and a slightly nicer axe + hammer.
/:cl:

### DIFF
--- a/code/game/objects/items/rogueitems/natural/stones.dm
+++ b/code/game/objects/items/rogueitems/natural/stones.dm
@@ -477,12 +477,12 @@ GLOBAL_LIST_INIT(stone_personality_descs, list(
 	if(istype(W, /obj/item/rogueweapon/pick))
 		if(!isliving(user))
 			return ..()
-		var/mob/living/LivingUser = user
-		if(!LivingUser.client || !LivingUser.client.prefs?.autopicking)
+		var/mob/living/living_user = user
+		if(!living_user.client || !living_user.client.prefs?.autopicking)
 			return ..()
 		user.doing = FALSE
 		while(!QDELETED(src) && user.Adjacent(src))
-			if((LivingUser.energy <= 0))
+			if((living_user.energy <= 0))
 				break
 			if(!do_after(user, CLICK_CD_MELEE, TRUE, src))
 				break
--- a/code/game/objects/items/rogueitems/natural/stones.dm
+++ b/code/game/objects/items/rogueitems/natural/stones.dm
@@ -477,14 +477,14 @@ GLOBAL_LIST_INIT(stone_personality_descs, list(
 	if(istype(W, /obj/item/rogueweapon/pick))
 		if(!isliving(user))
 			return ..()
-		var/mob/living/L = user
-		if(!L.client || !L.client.prefs?.autopicking)
+		var/mob/living/LivingUser = user
+		if(!LivingUser.client || !LivingUser.client.prefs?.autopicking)
 			return ..()
 		user.doing = FALSE
 		while(!QDELETED(src) && user.Adjacent(src))
-			if((L.energy <= 0)
+			if((LivingUser.energy <= 0))
 				break
-			if(!do_after(user, CLICK_CD_MELEE, TRUE, src)
+			if(!do_after(user, CLICK_CD_MELEE, TRUE, src))
 				break
 			..()
 		return
--- a/code/game/objects/items/rogueitems/natural/stones.dm
+++ b/code/game/objects/items/rogueitems/natural/stones.dm
@@ -296,6 +296,33 @@ GLOBAL_LIST_INIT(stone_personality_descs, list(
 			S.set_up(1, 1, front)
 			S.start()
 	if( user.used_intent.type == /datum/intent/chisel )
+		if(istype(W, /obj/item/rogueweapon/chisel/assembly))
+			var/obj/item/rogueweapon/chisel/assembly/chisel_set = W
+			if(chisel_set.wielded)
+				var/turf/work_turf = get_turf(src)
+				var/obj/item/natural/stone/current_stone = src
+				while(current_stone)
+					playsound(current_stone.loc, pick('sound/combat/hits/onrock/onrock (1).ogg', 'sound/combat/hits/onrock/onrock (2).ogg', 'sound/combat/hits/onrock/onrock (3).ogg', 'sound/combat/hits/onrock/onrock (4).ogg'), 100)
+					user.visible_message("<span class='info'>[user] chisels the stone into a block.</span>")
+					if(!do_after(user, work_time, target = current_stone))
+						return
+					new /obj/item/natural/stoneblock(get_turf(current_stone.loc))
+					if(HAS_TRAIT(user, TRAIT_MASTER_MASON)) //double the amount for any in a stone worker role
+						new /obj/item/natural/stoneblock(get_turf(current_stone.loc))
+					new /obj/effect/decal/cleanable/debris/stony(get_turf(current_stone))
+					playsound(current_stone.loc, pick('sound/combat/hits/onrock/onrock (1).ogg', 'sound/combat/hits/onrock/onrock (2).ogg', 'sound/combat/hits/onrock/onrock (3).ogg', 'sound/combat/hits/onrock/onrock (4).ogg'), 100)
+					qdel(current_stone)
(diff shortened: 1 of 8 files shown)


code/game/turfs/closed/minerals.dm
```diff
@@ -93,8 +111,8 @@
 
 /turf/closed/mineral/attack_right(mob/user)
 	var/obj/item = user.get_active_held_item()
-	if(user.used_intent.type == /datum/intent/pick && (user.get_skill_level(/datum/skill/labor/mining) >= SKILL_LEVEL_JOURNEYMAN))
-		if(do_after(user, 4 SECONDS, TRUE, src))
+	if(istype(user.used_intent, /datum/intent/pick) && (user.get_skill_level(/datum/skill/labor/mining) >= SKILL_LEVEL_APPRENTICE))
+		if(do_after(user, 2 SECONDS, TRUE, src))
```
Kroto-Ker, line 119: What's the intention behind halving this do_after time?

ophaq: It was part of the port, if I recall, unless I remember wrongly. Regardless, 4 seconds seems a bit much. 2 seconds is more than enough to stop yourself from walking off the tile and not pick bumping. It is mostly just an arbitrary decision.

Even the auto chop / picking have around a 1.2 seconds do_after time on them. The .8 seconds does not make a huge difference in comparison but you'll feel it in-game, especially with TD.

code/game/objects/items/rogueitems/natural/stones.dm
```diff
@@ -447,6 +474,23 @@ GLOBAL_LIST_INIT(stone_personality_descs, list(
 			S.set_up(1, 1, front)
 			S.start()
 		return
+	if(istype(W, /obj/item/rogueweapon/pick))
+		if(!isliving(user))
+			..()
+			return
```
Xander3359:
```suggestion
			return ..()
```

code/game/objects/items/rogueitems/natural/stones.dm
```diff
@@ -447,6 +474,23 @@ GLOBAL_LIST_INIT(stone_personality_descs, list(
...
+		var/mob/living/L = user
+		if(!L.client || !L.client.prefs?.autopicking)
+			..()
+			return
```
Xander3359:
```suggestion
			return ..()
```

code/game/objects/items/rogueitems/natural/stones.dm
```diff
@@ -447,6 +474,23 @@ GLOBAL_LIST_INIT(stone_personality_descs, list(
...
+		user.doing = FALSE
+		while(!QDELETED(src) && user.Adjacent(src))
+			if((L.energy > 0) && do_after(user, CLICK_CD_MELEE, TRUE, src))
+				..()
+				if(QDELETED(src))
+					break
+			else
+				break
```
Xander3359: Much cleaner to use early returns here than to do if else.


```suggestion
			if((L.energy <= 0)
				break
			if(!do_after(user, CLICK_CD_MELEE, TRUE, src)
				break
			..()
```
And you don't need to check `if QDELETED()` because that's what you are looping over.

Xander3359: You can even put feedback here like a chat message or a popup

code/game/objects/structures/roguetown/newtree.dm
```diff
@@ -108,6 +108,24 @@
...
+	if(user.doing)
+		return
+	user.doing = FALSE
+	while(!QDELETED(src) && user.Adjacent(src))
+		if((L.energy > 0) && do_after(user, 1.5 SECONDS, TRUE, src))
+			if(QDELETED(src))
+				break
+			..()
+		else
+			break
```
Xander3359: Same here, prefer early returns.

code/game/objects/structures/roguetown/rogueflora.dm
```diff
@@ -24,6 +24,24 @@
...
+	if(user.doing)
+		return
+	user.doing = FALSE
+	while(!QDELETED(src) && user.Adjacent(src))
+		if((L.energy > 0) && do_after(user, 1.5 SECONDS, TRUE, src))
+			if(QDELETED(src))
+				break
+			..()
+		else
+			break
```
Xander3359, line 42: early returns

code/game/turfs/closed/minerals.dm
```diff
@@ -62,6 +62,22 @@
...
+		var/mob/living/carbon/human/user = AM
+		var/obj/item = user.get_active_held_item()
+		if(istype(user.used_intent, /datum/intent/pick) && (user.get_skill_level(/datum/skill/labor/mining) >= SKILL_LEVEL_APPRENTICE))
+			if(user.client && !user.client.prefs?.autopicking)
+				return
+			if(do_after(user, 1 SECONDS, TRUE, src, TRUE, null, TRUE))
+				if(!ismineralturf(src))
+					return
+				src.attackby(item, user, multiplier = 2)
+				user.stamina_add(25)
```
Xander3359: Many opportunities for early returns in this proc.

and we don't need `src.attackby()` you can just call `attackby()` directly.

Xander3359 (changes requested):
Also, please don't leave var/L 
Name your variables, avoid single-letter variables whenever possible.

Clocks-scp:
Is this what's causing automatic mining to suddenly not work at all?

ophaq:
> Is this what's causing automatic mining to suddenly not work at all?

Check your toggle settings for enabling automatic mining. When I tested locally it all worked.

Xander3359 (changes requested):
Please use snake_case when making variable names.

We have an established precedent and this is what we abide by in ss13 coding.

ophaq:
> Please use snake_case when making variable names.
> 
> We have an established precedent and this is what we abide by in ss13 coding.

Oops sorry. I saw other variables in the code named like that so I thought it was fine and didn't matter too much. I'll go fix it in a bit.

Clocks-scp:
> > Is this what's causing automatic mining to suddenly not work at all?
> 
> Check your toggle settings for enabling automatic mining. When I tested locally it all worked.

Auto-mining seems to swing extra times, consuming stamina for nothing. You SURE this has parity with before?

PR #1753, opened 2026-05-29
Port dead mob + sleeping NPCs leaving processing

Rayforx:
## About The Pull Request
Ports https://github.com/Azure-Peak/Azure-Peak/pull/4423, which was optimized by WeNeedMorePhoron (ported from scarlet 1210). 
Corpses are processed 3 times less often (every 6s instead of 2s)

NPCs with no players nearby stop the AI from running completely until a player approaches, instead of idling in a processing loop every second.

## Testing Evidence
Tested that the AI still works as normal
Tested that it decreases the time dilation, while not impacting ghosts or the AI starting/stopping.
image

## Why It's Good For The Game
Being able to walk a little faster to your partner seems good to me.

## Changelog

:cl:
add: Dead mobs being processed in a cycle of 6s instead of two seconds, and mob idle processing
/:cl:


jgDarkWolf (approved):

noelle-lavenza:
@Rayforx isn't this a partial revert of https://github.com/Rotwood-Vale/Ratwood-2.0/pull/652 ? how did you ensure that it _actually_ reduced time dilation? time dilation on local servers without player load is very subject to random chance.

Rayforx:
> @Rayforx isn't this a partial revert of #652 ? how did you ensure that it _actually_ reduced time dilation? time dilation on local servers without player load is very subject to random chance.

From what I can see, with the change, it hovers around .327 CPU from the profiler in-game for the dead code. It can technically be removed if it has an impact in-game. I asked the previous maintainer, and nobody knew why it was removed. And with AP still having it, it seemed worth a shot.

The NPC processing does help a lot more from my personal testing, and from looking at what I saw from the profiler, which was also the main point of the PR. It stops the AI when players or ghosts are not nearby.

Rayforx:
> how did you ensure that it actually reduced time dilation? time dilation on local servers without player load is very subject to random chance.

Also, fyi, I can't access the profiler unless I ask a maintainer, and the profiler is horrible to navigate in; I do not have the fancy tool that directly monitors the server, which Xander made. I can only test locally. The straightforward answer to this would be that a maintainer has to figure out, by seeing the stats on the server, if this has a positive impact.

PR #949, opened 2026-03-01
Sir, You Are Being Hunted - makes Hunted much more obvious

kimesik:
## About The Pull Request
image

## Testing Evidence
It's just a string change.

## Why It's Good For The Game
image

PR #1756, opened 2026-05-29
Rule-compliant explanation for HUNTED Vice!

NotGoldenglow:
## About The Pull Request

Makes it clearer that this is not a license to avoid roleplay.

## Testing Evidence

Two lines. 
## Why It's Good For The Game

Makes it clearer what you should and shouldn't expect.

## Changelog

:cl:
code: Hunted vice description cleaned up and elaborated on.
/:cl:

### DIFF
--- a/code/datums/character_flaw/_character_flaw.dm
+++ b/code/datums/character_flaw/_character_flaw.dm
@@ -421,7 +421,7 @@ GLOBAL_LIST_INIT(character_flaws, list(
 	name = "Hunted"
 	desc = "Something in my past has made me a target. I'm always looking over my shoulder. <br>\
 	YOU MAY BE PERMANENTLY REMOVED FROM THE ROUND WITHOUT PRIOR ESCALATION BY YOUR ASSASSIN!<br>\
-	<small>This is not a license to mindlessly frag. No Escalation != No Roleplay. This merely allows you to remove them after confrontation. You should still engage your target as any other person you intend to kill before doing so.</small>"
+	<small>This is not a license for them to mindlessly frag you. No Escalation != No Roleplay. This merely allows them to remove you after confrontation. They should still engage their target as any other person they intend to kill before doing so.</small>"
 	var/logged = FALSE
 
 /datum/charflaw/hunted/on_mob_creation(mob/user)
--- a/code/datums/character_flaw/_character_flaw.dm
+++ b/code/datums/character_flaw/_character_flaw.dm
@@ -419,7 +419,9 @@ GLOBAL_LIST_INIT(character_flaws, list(
 
 /datum/charflaw/hunted
 	name = "Hunted"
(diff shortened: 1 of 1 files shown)


coffeecoffeecoffee:
It is an incredible shame that we just can't escape having Azure policies and trying to be Azure, that this becomes necessary.

NotGoldenglow:
> It is an incredible shame that we just can't escape having Azure policies and trying to be Azure, that this becomes necessary.

It's an incredible shame that people take a lack of exhaustive, clear-cut rules as permission. It's why Azure had them in the first place. 
Someone goes and does something bad and then gets told "This is bad", they'll immediately point to the "Well it wasn't in the rules", so we need to keep adding more and more rules because otherwise we'll be back to this in a week.

coffeecoffeecoffee:
> It's an incredible shame that people take a lack of exhaustive, clear-cut rules as permission. It's why Azure had them in the first place. Someone goes and does something bad and then gets told "This is bad", they'll immediately point to the "Well it wasn't in the rules", so we need to keep adding more and more rules because otherwise we'll be back to this in a week.

I do not think people want these rules, and that we seem to be restricting and administrating our way into trying to be exactly like another server that already exists, is what I'm saying. I'm not saying the rules aren't as they exist - I'm saying Ratwood shouldn't be like this and I don't think I'm alone in saying I wish it weren't moving more in this direction.

NotGoldenglow:
It's honestly quite simple. The server isn't trying to be 'like another server', and saying this severely downplays the effort that goes into maintaining a healthy environment. People don't want others to go around and treat it like CM, left clicking people willy nilly, as is how people are currently acting it.

hugmeorelse:
> It's honestly quite simple. The server isn't trying to be 'like another server', and saying this severely downplays the effort that goes into maintaining a healthy environment. People don't want others to go around and treat it like CM, left clicking people willy nilly, as is how people are currently acting it.

I don't think people are quite understanding that the original PR is talking strictly about the Assassin Job, not Gnolls. The PR changing the description was made on February 28th. Yes, while it was merged recently, it's clear what the intent was with that PR. Honestly, it should have never been merged. Meanwhile, the Gnoll PR was made in March and merged in April. That PR description is aimed SOLELY AT ASSASSIN JOBS as if the language itself is not clear.

I highly recommend changing the description to say "Assassin Job" instead of "an Assassin", as people are extending that to mean "Gnolls" when that was never the design or case.

See #1133 and #949.

IstronglyDislikeHavingToMakeNewAcco:
Why does it still not mention anything about gnolls?
When I first joined the server I built an entire character around this, thinking they would be targeted by humans over a debt, only to have gnolls show up and ask "why are you marked?".

NotGoldenglow:
> Why does it still not mention anything about gnolls? When I first joined the server I built an entire character around this, thinking they would be targeted by humans over a debt, only to have gnolls show up and ask "why are you marked?".

The PR which added this was made before Gnolls were a thing, and Gnolls repurposed the vice. I've put up another PR (https://github.com/Rotwood-Vale/Ratwood-2.0/pull/1763) that seeks to correct this.

PR #1781, opened 2026-05-31
R☮TS every last Cursed Collar lock mechanism there is. | Kills fetish content | Enforces Respect of Consent

JustMeTheIInd:
## About The Pull Request

Makes the Cursed Collar removable freely and entirely, anyone at any time can remove it by hand. 

I considered making it an Eoran Ritual but I figured it's too much work for gooner material.

Chastity content can be removed with the use of settings, so I didn't change anything regarding that.

## Testing Evidence

I removed three lines. It works I swear.

## Why It's Good For The Game

You should be able to withdraw your consent at any time, for any reason. 

The logic here is simple:
-You CC'd someone:

-> That someone (OOCly) WANTS to be kept CC'd -> They will keep the Collar ON.

-> That someone (OOCly)  DOESN'T want to be CC'd anymore. -> They will take the Collar OFF.

"But what if someone ELSE removes the Cursed Collar?"
-> They did it purely for the love of the game, and had otherwise no valid reason to do it.
- That is a rulebreak. Act accordingly.

-> They did it, while having an actual reason.
- That is not a rulebreak. Act accordingly. 

## Changelog

:cl:
balance: Cursed Collar is now removable.
/:cl:

### DIFF
--- a/modular/code/modules/slave_collar/cursed_collar.dm
+++ b/modular/code/modules/slave_collar/cursed_collar.dm
@@ -98,7 +98,6 @@
 			return
 
 		SEND_SIGNAL(C, COMSIG_CARBON_COLLAR_BOUND, collar_master, src)
-		ADD_TRAIT(src, TRAIT_NODROP, CURSED_ITEM_TRAIT)
 		log_combat(user, C, "tried to collar", addition="with [src]")
 	applying = FALSE
 
@@ -169,7 +168,6 @@
 		return
 
 	SEND_SIGNAL(user, COMSIG_CARBON_COLLAR_BOUND, collar_master, src)
-	ADD_TRAIT(src, TRAIT_NODROP, CURSED_ITEM_TRAIT)
 
 	user.visible_message(span_warning("Cursed collar around [user]'s neck clicks shut!"), \
 							span_userdanger("Cursed collar around your neck clicks shut!"))
@@ -192,8 +190,6 @@
 		if(CM && (user in CM.my_pets))
 			CM.remove_pet(user)
 
-	REMOVE_TRAIT(src, TRAIT_NODROP, CURSED_ITEM_TRAIT)
-
 /obj/item/clothing/neck/roguetown/cursed_collar/canStrip(mob/living/carbon/human/stripper, mob/living/carbon/human/owner)
 	// Some strip call sites may not pass owner; infer from loc when possible.
 	if(!owner && ishuman(loc))
@@ -209,7 +205,6 @@
 		owner = loc
 
 	if(stripper?.mind == collar_master)
-		REMOVE_TRAIT(src, TRAIT_NODROP, CURSED_ITEM_TRAIT)
 		if(owner)
 			SEND_SIGNAL(owner, COMSIG_CARBON_LOSE_COLLAR)
 		return owner ? owner.dropItemToGround(src, force = TRUE) : FALSE


The-Sun-In-Splendour:
Oooor you add an OOC remove collar button that alerts admins when you press it. Citadel station had something similar for vore.

IronDragoon:
Personally not a fan of straight up removal. Would be far better to just make this an OOC pref toggle to make collars removable or not.

Randumb-Phantomb:
the cursed collar isn't really gooner material, it's just dominating someone.

Gug1234:
> the cursed collar isn't really gooner material, it's just dominating someone.

Cursed collar is 100% gooner content. It's geared towards ERP. It's able to be as powerful as it is because it's intended for ERP not for mechanical advantage.

90% of the buttons are for sexual slavery stuff, use a castifico or a fake cursed collar if you want non sexual slavery stuff.

Broill:
Like listen, I get it. But this is the only option for prisoners that doesn't make them miserable is the issue. The lesser cursed collar does nothing except being unremovable by the wearer and the castifico collar is one of the lamest situation you can find yourself into if they also send you to jail. So people use the cursed collar for both fetish bullshit but also actual "I'd like to keep this person alive but controlled" type deals. So please make the lesser cursed collar useful for sending shocks or something if that's gonna happen.

Gug1234:
> Like listen, I get it. But this is the only option for prisoners that doesn't make them miserable is the issue. The lesser cursed collar does nothing except being unremovable by the wearer and the castifico collar is one of the lamest situation you can find yourself into if they also send you to jail. So people use the cursed collar for both fetish bullshit but also actual "I'd like to keep this person alive but controlled" type deals. So please make the lesser cursed collar useful for sending shocks or something if that's gonna happen.

Use a whip or RP

As for the electric collar, I believe someone else was working on a dedicated shock collar. That should cover your bases — although I may be mistaken on my assumption.

Broill:
> > Like listen, I get it. But this is the only option for prisoners that doesn't make them miserable is the issue. The lesser cursed collar does nothing except being unremovable by the wearer and the castifico collar is one of the lamest situation you can find yourself into if they also send you to jail. So people use the cursed collar for both fetish bullshit but also actual "I'd like to keep this person alive but controlled" type deals. So please make the lesser cursed collar useful for sending shocks or something if that's gonna happen.
> 
> Use a whip or RP
> 
> As for the electric collar, I believe someone else was working on a dedicated shock collar. That should cover your bases — although I may be mistaken on my assumption.

if the shock thing's the case then sure that would solve it. Otherwise there is really no way to keep someone in check but outside a prison cell than the fetish item which is bit of a problem.

cockatieldevon:
I can support allowing people to refuse to have the collar put on them in the first place, but having just anyone at any time be able to rip the collar off and run off with my CC so I can't use it elsewhere is weird.

Tbh we should just enforce consent on the initial application of the collar if this is a real problem in the game. If they OOCly want it removed later and the owner doesn't arrange to release them, ahelp it.

The collar is an ERP item, but I don't always use it exclusively for ERP. Sometimes (for me) it's a way to keep some attacker in the game without killing them. If they can just walk off screen and remove it or have their buddy remove it and get a free CC out of it, that's kind of boring.

JackSage:
Cursed Collar is gooner content, I am sorry but your 1 slot load out item that costs 0 points that can effectively make someone useless is not meant to be anything but gooner content and should require OOC consent from the affected party.

It also should not be used as an excuse to "convert" someone to wretch/bandit minion. I do not want to fight the KC as a warden because they decided to run into the bog alone sandbag against a bandit then said bandit pull a collar that costs next to nothing that 90% of the server just magically has and now the KC is "mind controlled" and must frag all the garrison.
This is the equivalent to using the free gooner shit in a lustwish vendor on bubber to convert the HOS into being your minion.

Could you imagine if a base SS13 server just allowed anyone and everyone to put a mindslave implant in their loadout for free? That's basically what some people treat the cursed collar like.

But as for the PR- an OOC button or something similar to the chastity system should be added that allows the person to remove the item. I am tired of "coders" making 1-5 line "edits" which is just deleting/commenting out code because they're too lazy or too inexperienced to make an actual fix.

I swear we need to add the "Avoid Quick Fixes" rule from the server to the github. This is like the third or fourth PR this week where someone just straight up removes something or deletes code because it's unbalanced, or being used in an unintended way.

Gamer-Johnson:
you seem to have good intentions but no offense, this is one of the worst ways you could possibly do this. it **will** get abused, and there are more effective ways to enforce consent with cursed collars anyway. a consent dialog box when putting the collar on would be good enough imo, and iirc it already exists in the code but is bugged.

an OOC escape button might get abused but there are ways to mitigate that, such as making the collar drop at the owner's feet rather than the wearer's. otherwise, people will get collared intentionally, only to pop the collar and run off with their new loot.

a cursed collar ERP content toggle would also work, i think. re-word how it refers to collared players (currently they are referred to as "pets" rather than "slaves") and otherwise keep the collar as-is, but certain features (arousal, nudity, forced love, etc.) are either altered or flat out unavailable with the toggle turned off. for instance: forced nudity could disallow the victim from wearing _armor,_ but they can still wear _clothes._

also...
>I considered making it an Eoran Ritual but I figured it's too much work for gooner material.

this is a genuinely good idea and i fully support it. it could also work as a xylixian or matthiosite ritual, since both of them are meant to value freedom and abhor slavery.

also part 2...
>I removed three lines. It works I swear.

are you **positive** this doesnt affect cursed chastity? the cursed chastity belt is effectively a reskin of the collar. they share almost all of their code, so i would not be surprised at all if allowing the collar to be removed freely by hand would also allow the belt to be removed by hand.

coffeecoffeecoffee:
there's a serious point here and a problem that absolutely needs to be solved, but removing the fetish content is not the way. IMHO this should absolutely BE there, but not an actual viable provided method for the duchy to fucking punish people.

remove it from it's spawn points on the map, not from the game

Gug1234:
I don't see a problem with it being removable, the toggle in my piercing, plug, etc workspace does the same thing where it just pops the collar off same as chastity works. I do think the toggle is the cleaner solution rather than just removing the cursed item and no drop traits. When I'm able I'll cherry pick out the relevant snippets from my stuff and send it over.

Jamdrawers:
Closest thing to this I think might be having a hard toggle on like the erotic functions of the collar so that it still might be used for like... forcing compliance from a prisoner or whatever.

Its like... letting anyone have an OOC toggle that lets them break out of grabs or chains or pillories in case someone might use those chains to kidnap and rape them.

Gug1234:
> Closest thing to this I think might be having a hard toggle on like the erotic functions of the collar so that it still might be used for like... forcing compliance from a prisoner or whatever.
> 
> Its like... letting anyone have an OOC toggle that lets them break out of grabs or chains or pillories in case someone might use those chains to kidnap and rape them.

It's not for capturing people mechanically. Its power and ease of access are because it's an ERP item. Chains, ropes, and pillories are all deniable mechanically with enough strength or time — the collar is not. There is no good middle ground here where the collar can be used for mechanical advantage. Full stop, don't use it outside of ERP. If you really need a collar, use the lesser collar craftable cheaply by the smith.

Gug1234:
> > > Closest thing to this I think might be having a hard toggle on like the erotic functions of the collar so that it still might be used for like... forcing compliance from a prisoner or whatever.
> > > Its like... letting anyone have an OOC toggle that lets them break out of grabs or chains or pillories in case someone might use those chains to kidnap and rape them.
> > 
> > 
> > It's not for capturing people mechanically. Its power and ease of access are because it's an ERP item. Chains, ropes, and pillories are all deniable mechanically with enough strength or time — the collar is not. There is no good middle ground here where the collar can be used for mechanical advantage. Full stop, don't use it outside of ERP. If you really need a collar, use the lesser collar craftable cheaply by the smith.
> 
> Simply allow it to be taken off by turning off ERP content, or chastity content or something. You enable the setting and it pops off. Shrimple?

in preferences_toggles.dm
`
/client/proc/toggle_cursed_content()
	set category = "Options"
	set name = "Toggle Cursed Collar Content"
	set hidden = 1
	if(!prefs)
		return
	prefs.cursed_enabled = !prefs.cursed_enabled
	prefs.save_preferences()
	if(prefs.cursed_enabled)
		to_chat(src, "Cursed collar content enabled.")
	else
		if(hascall(src, "modular_handle_cursed_toggle_disable"))
			call(src, "modular_handle_cursed_toggle_disable")()
		to_chat(src, "Cursed collar content disabled.")
`
In chastity_hooks.dm
`
/client/proc/modular_handle_cursed_toggle_disable()
	if(!ishuman(mob))
		return
	var/mob/living/carbon/human/human_mob = mob

	// Strip cursed chastity device
	var/obj/item/chastity/device = human_mob.chastity_device
	if(device?.chastity_cursed)
		device.cleanup_cursed_binding(human_mob)
		device.remove_chastity(human_mob)
		device.forceMove(get_turf(human_mob))
		human_mob.visible_message(span_notice("The divine hand of Eora dissolves the cursed bindings from [human_mob]'s loins!"))

	// Strip cursed collar
	var/obj/item/clothing/neck/roguetown/cursed_collar/collar = human_mob.get_item_by_slot(SLOT_NECK)
	if(istype(collar))
		REMOVE_TRAIT(collar, TRAIT_NODROP, CURSED_ITEM_TRAIT)
		SEND_SIGNAL(human_mob, COMSIG_CARBON_LOSE_COLLAR)
		human_mob.dropItemToGround(collar, force = TRUE)
		human_mob.visible_message(span_notice("The divine hand of Eora shatters the cursed collar from [human_mob]'s neck!"))
`
formatting munted those snippets, but whatever

Elli-Skala:
Okay so in older ratwood 2 this actually how this original way. The no drop was added later because consistently someone would have a collar put on them and they would take it off to look at it and then put it back on but that breaks the link

JustMeTheIInd:
Fine, I shall make it a button.. (and a ritual)

PR #1788, opened 2026-06-01
Revert "Men Can Moan Too"

WolfOrion:
Reverts Rotwood-Vale/Ratwood-2.0#1767

After seeing them in practice beyond just a single test to make sure they work, I do not think they are quality enough to remain as-is. Currently trying to source better moans to use instead.


coffeecoffeecoffee:
Everybody knows male moans are a myth unless it's a femboy.

Jamdrawers:
I enjoy these moans partly because they're just so funny to me. Though we've always been able to use them by just saying *moan, just didn't show up in the noises panel

PR #1735, opened 2026-05-26
Various mapping things, part-merges some church stuff

Jamdrawers:
## About The Pull Request

-Adds a bog witch house in the far southwest in the cursed mire courtesy of Deline (https://github.com/Rotwood-Vale/Ratwood-2.0/pull/1718). The door can be claimed by wretches (and witches) the same way towners claim towner houses! Happy wretching in the bog witch house

-Adds an area south of the church with old buildings you can have quiet one-on-one prayer sessions with. Originally by Deline but trimmed a bit by me. I think it might still be a bit too big and have the churchfolk a bit spread out, but we'll see how this goes

-I WAS gonna add a tampered-with-version of Deline's change to the lowtown church but my SDMM crashed and I don't have the heart to do it again yet. Next time.

-roguegrass-ifies reed bushes, turning them into things that rustle and that can be cut into fibre

-makes use of some old Stairs sprites and integrates them into some existing places

-Nets in the city guard barracks

-Creates 'tough' subtypes of certain doors with increased integrity and lockdifficulty and puts them on various Vaults in a way sorta like https://github.com/Rotwood-Vale/Ratwood-2.0/pull/1716 but not quite as extreme

-Various fixes

## Testing Evidence

image
(Okay I since added more lighting since this screenshot)

## Why It's Good For The Game

We'll see how it goes!

## Changelog

:cl:
map: -Bog Witch house that wretches can claim in the southwest in the cursed mire
new Old Church District south of the church
reed bushes behave like grass
Nets in the city guard barracks
various fixes and tweaks
/:cl:

PR #1794, opened 2026-06-01
[BUGFIX] Restores claiming functionality to the two towner witch houses in the bog

Gamer-Johnson:
## About The Pull Request
The new 'wretched' doors added in PR #1735 were non-functional. Now it works.

The reason _why_ it didn't work is because `resident_role` does not accept lists as inputs, but was given one anyway.

I used `resident_advclass` instead, but to make _that_ work, I had to list off **every single advclass** that was intended to be able to claim the house.

I tried to have it take `/datum/advclass/wretch` as the wretch side of the list, since that would have been _so_ much easier to look at, but that did not work in my testing.

## Testing Evidence
Witch:
door test 1

Wretch:
door test 2

## Why It's Good For The Game
Less bugs is good bugs. Test your shit.

## Changelog

:cl:

fix: fixed the new door from #1735

/:cl:

### DIFF
--- a/code/game/objects/structures/mineral_doors.dm
+++ b/code/game/objects/structures/mineral_doors.dm
@@ -1277,6 +1277,7 @@
 	locked = TRUE
 	keylock = TRUE
 	grant_resident_key = TRUE
-	resident_key_type = /obj/item/roguekey/townie
+	resident_key_type = /obj/item/roguekey/townie// should be every wretch class - ideally we can get resident_role to accept lists but until then this'll do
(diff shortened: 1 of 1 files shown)


cockatieldevon:
This bug inconveniences me. This fix uninconveniences me. This is ideal.

Jamdrawers:
Would be nice to have a slightly more Elegant Solution

Gamer-Johnson:
> Would be nice to have a slightly more Elegant Solution

It would be, but the best I can think of is to either: rewrite how either `resident_role` or `resident_advclass` works, so that it can accept just `list(/datum/job/roguetown/villager, /datum/job/roguetown/wretch)` as a valid input, or just revert to the old `/obj/structure/mineral_door/witch` and drop the whole wretch-claimable house idea entirely. I am not a coder, so the best I could realistically _do_ is that second option.

Either way, a door variant that is mildly annoying to look at but is functional ingame is miles better than one that looks pretty in the code but does not function ingame.

Jamdrawers:
currently fiddling around with seeing how difficult it might be to change how resident_role works - main thing is just that it'd be annoying when a new wretch advclass gets added having to go back (and it'd be useful knowledge to have in general)

Jamdrawers:
Well, haven't gotten it to work quite yet, so until then it's certainly better than not merging it

PR #1820, opened 2026-06-07
Adds 6 Rclick Animations

FreeStylaLT:
## About The Pull Request
This PR adds 6 new animations for the rclick interactions (waving etc), along with a few font / functionality changes:

### Adds a Wave animation
dreamseeker_6mW2sOptya

(In practice it's shorter now, I was too lazy to re-record it)

### Adds a Shoo animation
dreamseeker_wftVsrmRTP

### Adds a Beckon animation
dreamseeker_289l27AFq1

### Adds 3 Taunt animations

The Standard one
dreamseeker_W9VF0cXBbO

The "Enhanced" one
dreamseeker_rZh4WkfPky
This one will trigger if:
- The taunter is stressed (10 stress points or 'stage 2' icon wise, iirc)
- The taunter has the Paranoid vice
- The taunter is a Graggarite

The "Eoran" one
dreamseeker_M16f50mL4Q
This one will trigger if:
- The taunter is an Eoran
- The taunter is a Pacifist

### In addition: 
this PR adds two new spans for beckoning and taunting:
dreamseeker_WnbPytyX1p

All of these emotes (and moodlets) are now visible through the FOV cone. If you ever wondered why someone never waved back, it's likely because they were never transmitted the wave because they required you to be in the field of view.

These little moodlets are ONLY visible to the user and the target. Everyone else will only get the standard chat message.

Using an rclick like this now applies a short (~0.6s or so) click cd as to not spam it.

EDIT;
This PR also adjusts every overhead to only play once (like the fsalutes and stress notifs), this fixes them oddly flickering in and out when used.

## Testing Evidence
Above, though I re-tested everything here, too.

## Why It's Good For The Game
Since these are only visible to the target of the rclick, they should especially help during crowded / chaotic moments without needing to scream at someone to come closer, etc.

All sprites courtesy of Ollanius and with edits / animations done by myself.

## Changelog

:cl:
add: Added 6 new rclick animations.
/:cl:


Elli-Skala:
Could you add deceptive_meekness to the eoran taunt?

PR #1829, opened 2026-06-08
fixes rclick animations not working correctly

FreeStylaLT:
## About The Pull Request
forgot to remove debug functionality oops oops oops, it was only transmitting to the user rather than the appropriate target


FreeStylaLT:
Uh oh, Main's not compiling anymore uh oh. (Nothing to do with this branch as far as I can tell)

PR #1842, opened 2026-06-10
Makes nobility virtue cost triumphs

Xander3359:
## About The Pull Request
Makes nobility virtue cost triumphs

## Why It's Good For The Game
As per the poll in the discord (seen [here](https://discord.com/channels/1240735983276654733/1474558726303387820/1511080926531027105)) . The overall feeling is that nobility is too common and that detracts from the importance of being a noble. 

If you are in the town and you see the local beggar is a noble, that is very immersion breaking.

Setting the triumph cost (open to adjustment) to 7 should make the noble virtue something you have to consider, rather than it just being "rich and shrewd++++"

## Changelog

:cl:
balance: Noble virtue now costs 7 triumphs
/:cl:

### DIFF
--- a/modular_azurepeak/virtues/utility.dm
+++ b/modular_azurepeak/virtues/utility.dm
@@ -1,10 +1,12 @@
 /datum/virtue/utility/noble
 	name = "Nobility"
-	desc = "By birth, blade or brain, I am noble known to the royalty of these lands, and have all the benefits associated with it. I've cleverly stashed away a healthy amount of coinage, alongside a familial heirloom."
+	desc = "By birth, blade or brain, I am noble known to the royalty of these lands, and have all the benefits associated with it. \
+			I've cleverly stashed away a healthy amount of coinage, alongside a familial heirloom."
 	added_traits = list(TRAIT_NOBLE)
 	added_skills = list(list(/datum/skill/misc/reading, 1, 6))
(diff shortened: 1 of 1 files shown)


AllCorruptingWyrm:
You should make it cost more.

LittleEvil9:
> You should make it cost more.

7 is more than enough to discourage people from taking it unless its central to their character. I don't even agree with making it cost triumphs, imo it should be restricted from peasantry roles.

zeroisthebiggay:
> > You should make it cost more.
> 
> 7 is more than enough to discourage people from taking it unless its central to their character. I don't even agree with making it cost triumphs, imo it should be restricted from peasantry roles.

triumphs are literally free and you get five if you join a round at the start, not counting any triumphs you get for actually doing things in round
just play a different character once in a while and you're fine, I do it with my high society character

EmilyFontaine:
Seven seems fine, especially because it'd be competing with High Society (12) at higher costs. It's not like most people will run out of triumphs anyway, it's barely a loss if you play the full round.

annikaRU:
I personally don't think it should cost triumphs, if we're going off the poll the votes for "Don't do it" came in rather close

It's not a virtue problem, it's a player problem which could easy be fixed by restricting roles from the virtue like how the church is restricted from nobility or even restricting species

PR #1852, opened 2026-06-11
The great debris ening

Jamdrawers:
## About The Pull Request

Adds Debris to a great many objects and turfs, meaning that stone walls break into stone, wood walls break into wood, and all sorts of separate things break into wood or sticks or stones when necessary.
Generally speaking objects will break into less material than they cost to build, but its still a nice consolation

## Testing Evidence

Later

## Why It's Good For The Game

Makes broken walls less of a hassle since you get some repair material immediately! Also just a nice bit of interactivity with the world

## Changelog

:cl:
add: a great many objects will break into appropriate materials when broken
/:cl:

### DIFF
--- a/code/game/objects/items/rogueitems/natural/zizobane.dm
+++ b/code/game/objects/items/rogueitems/natural/zizobane.dm
@@ -6,6 +6,7 @@
 	density = FALSE
 	anchored= TRUE
 	var/time_delay = 0
+	debris = list(/obj/item/reagent_containers/food/snacks/zizo_bane = 1)
 
 /obj/structure/zizo_bane/Initialize(mapload)
 	. = ..()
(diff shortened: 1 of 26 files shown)


coffeecoffeecoffee:
Can we do something about doing _away_ with the massive tremendous gargantuan amount of debris that skeleton wars generate, though?

Good change, that said

Jamdrawers:
> Can we do something about doing _away_ with the massive tremendous gargantuan amount of debris that skeleton wars generate, though?

I did recently by making them not drop bows or weapons, which are the parts of the skeleton that don't easily fit inside of a sack and thus aren't easily cleaned.

code/game/objects/items/rogueitems/natural/zizobane.dm
```diff
@@ -6,6 +6,7 @@
 	density = FALSE
 	anchored= TRUE
 	var/time_delay = 0
+	debris = list(/obj/item/reagent_containers/food/snacks/zizo_bane = 1)
```
Arturlang, line 8: new var declarations should be after overrides

Arturlang (changes requested):

PR #1856, opened 2026-06-11
Adds voice commands for yielding.

Xander3359:
## About The Pull Request
Adds voice commands for yielding.

## Testing Evidence
image

## Why It's Good For The Game
Adds a way to yell out for someone to yield. 

This should make it so that yielding is less esoteric knowledge.

## Changelog

:cl:
add: Calling out for yielding will now put up a prompt to nearby people.
/:cl:


NotGoldenglow:
... There's no images of said prompt so I have no idea if this is how it'll work, but.

I don't think auto-opening prompts is a good idea. They take focus away from your window, which makes your stop moving. You yell at someone who's sprinting and they'll stop moving, letting you catch up even if they don't WANT to surrender. Maybe a chatbox prompt!

zBxnny:
If you open a popup on my screen while I'm fighting, I'm round removing you.

Consider making it a loud yell, like the orison one, that floods the chatbox with big text instead of a popup.

Xander3359:
> 
> 
> ... There's no images of said prompt so I have no idea if this is how it'll work, but.
> 
> 
> 
> I don't think auto-opening prompts is a good idea. They take focus away from your window, which makes your stop moving. You yell at someone who's sprinting and they'll stop moving, letting you catch up even if they don't WANT to surrender. Maybe a chatbox prompt!

It's not a prompt, it gives them a little flag over their head, and you get a status effect where your debuffs are (on the top left of your screen). That's what you click on to yield, it pops up the confirmation window when you click on the status.

AisuDracov:
It seems to force compliant people to yield, even if they hear it off screen. Which is.. annoying.

PR #1857, opened 2026-06-11
SSmobs upgrade

PatapoIIIa:
## About The Pull Request

Splits mob processing into alive/dead subsystems and makes the hibernation check itself nearly free.

```mermaid
flowchart TB
    subgraph BEFORE["Before optimizations"]
        direction TB
        A["SSmobs — every 2s<br/>copies the entire mob_living_list"]
        B["Life() — every living mob<br/>even in an empty dungeon"]
        C["DeadLife() — every 2s<br/>every corpse, all round long"]
        D["SShumannpc — every second<br/>process_ai() for all NPCs, incl. sleepers"]
        E["stat = written directly<br/>mob lists maintained by hand"]
        A --> B
        A --> C
    end

    subgraph AFTER["After optimizations"]
        direction TB
        F["set_stat() — single mutation point<br/>moves mobs between lists"]
        G["mob_living_active_list<br/>living only"]
        H["mob_living_dead_list<br/>corpses only"]
        I["SSmobs (Alive) — every 2s<br/>scans living only"]
        M["SSmobs (Dead) — every 6s<br/>low priority"]
        J["Fast path — stays asleep<br/>small fixed fast path, no hibernation predicate calls"]
        K["Life()<br/>awake mobs only"]
        N["Asleep on an empty z<br/>DeadLife() skipped for eligible corpses on inactive z-levels"]
        O["Wake — event-driven<br/>Human NPC wake paths — damage, retaliation, client proximity"]
        L["should_hibernate() — full check chain<br/>runs for awake hibernation candidates + 30s sleeper revalidation"]
        F --> G
        F --> H
        G --> I
        H --> M
        I --> J
        I --> K
        M --> N
        O --> K
        J -.- L
    end

    BEFORE ~~~ AFTER

    classDef hot fill:#F5C4B3,stroke:#993C1D,color:#4A1B0C
    classDef cheap fill:#9FE1CB,stroke:#0F6E56,color:#04342C
    classDef event fill:#FAC775,stroke:#854F0B,color:#412402
    classDef neutral fill:#ECECEC,stroke:#5F5E5A,color:#2C2C2A

    class B,C,D hot
    class J,N cheap
    class O event
    class A,E,F,G,H,I,M,K,L neutral
```

**Alive/dead split:**
- `SSmobs` ("Mobs (Alive)") processes only living mobs from `GLOB.mob_living_active_list` (every 2s)
- New `SSmobs_dead` ("Mobs (Dead)") processes corpses from `GLOB.mob_living_dead_list` every 6s at lower priority
- `set_stat()` is now the single point of `stat` mutation: sends `COMSIG_MOB_STATCHANGE` (new + old), moves mobs between the lists on the DEAD boundary; all direct `stat =` writes converted (death, revive, update_stat, VV, zombie wake, brainmob)

**Hibernation made O(1) per sleeping mob:**
- Cached `hibernating` flag: the full predicate chain (`should_hibernate` → `can_hibernate` → `is_calm` etc., ~7 proc calls) now runs only when a mob *enters* hibernation and on a 30s revalidation pass; every other fire a sleeper costs ~5 field reads and zero proc calls
- Bodypart-sensitive state (wounds/bleeding/embedded) is tracked by cached flags set at the source (wound add, bleed change, embed, dismemberment) — sleepers never iterate bodyparts
- Humans sleep by spatial-grid client proximity (`clients_in_range`), everything else by active z-levels (±1); SSD bodies (`ckey`) and `ignore_hibernation` mobs are exempt
- Human NPCs wake on applied damage, retaliation, and client proximity. Crossing the DEAD boundary clears hibernation state, and sleeping mobs are fully revalidated every 30 seconds.

**AI dedup:**
- `SShumannpc` no longer re-evaluates the hibernation predicate per NPC every second — one flag read; sleeping (`NPC_AI_SLEEP`) NPCs are removed from AI processing entirely and re-enter via wake (`wake_from_hibernation` → `handle_ai`, or client proximity → IDLE → Life)
- `human/Life()` no longer runs the full predicate a third time — restored the cheap `NPC_AI_SLEEP` early return

## Testing Evidence

Tests on a live server are needed.

## Why It's Good For The Game

An attempt to stop the round from degrading due to mobs

## Changelog

:cl:
fix: corpses are processed in a separate slow low-priority pass instead of competing with living mobs
fix: sleeping NPCs no longer consume combat-AI processing every second
code: stat transitions centralized through set_stat() with COMSIG_MOB_STATCHANGE; alive/dead list membership is maintained automatically
refactor: hibernation state is cached and event-invalidated — sleeping mobs cost near-zero per tick, with a 30s self-healing revalidation pass
admin: MC panel shows Mobs (Alive) A:/H: and Mobs (Dead) D:/H: counters; var/ignore_hibernation exempts special NPCs
/:cl:


code/controllers/subsystem/mobs.dm
```diff
@@ -30,28 +33,133 @@ SUBSYSTEM_DEF(mobs)
...
 
+/datum/controller/subsystem/mobs/proc/build_active_z_map()
+	if(!islist(clients_by_zlevel) || !world.maxz)
+		return null
+	var/list/active = new /list(world.maxz)
+	for(var/z in 1 to min(clients_by_zlevel.len, world.maxz))
+		if(!length(clients_by_zlevel[z]))
+			continue
+		active[z] = TRUE
+		if(z > 1)
```
noelle-lavenza, line 47: can you explain why we set the levels directly below and above to active as well? it's very likely this will erase most of the benefits of the activity map at all because then 2 players are enough to cover 6 z-levels if they're positioned properly.

code/controllers/subsystem/mobs.dm
```diff
@@ -30,28 +33,133 @@ SUBSYSTEM_DEF(mobs)
...
 		var/mob/living/L = currentrun[currentrun.len]
 		currentrun.len--
 
 		if(!L || QDELETED(L))
-			GLOB.mob_living_list.Remove(L)
+			GLOB.mob_living_list -= L
+			GLOB.mob_living_active_list -= L
+			GLOB.mob_living_dead_list -= L
+			GLOB.alive_mob_list -= L
+			GLOB.dead_mob_list -= L
```
noelle-lavenza: list removals are not free, it is O(n), so the cost of a list removal increases the larger the list is. this is especially true if the item is not in the list at all. this may seem fine for the sake of thoroughness but will if anything worsen time dilation if something in `currentrun` is harddeleted or GC'd

code/controllers/subsystem/mobs.dm
```diff
@@ -30,28 +33,133 @@ SUBSYSTEM_DEF(mobs)
...
 
 		if(L.stat == DEAD)
-			L.DeadLife()
-		else
-			L.Life(seconds, times_fired)
-			alive_mobs++
+			continue
+
+		if(L.hibernating)
+			if(!full_recheck && !L.client && !L.ckey)
```
noelle-lavenza:
```suggestion
			if(!full_recheck && !L.client)
```
check for client or check for ckey, checking both is redundant. there is not a meaningful scenario where you have a client but no ckey or vice versa, ckeys universally identify clients.

code/modules/mob/living/life.dm
```diff
@@ -111,11 +111,67 @@
 		handle_wounds()
 		handle_embedded_objects()
 		handle_blood()
-	update_sneak_invis()
```
noelle-lavenza, line 114: ~~won't this mean that mobs that die (e.g. of poison or bloodloss) while sneaking will remain invisible?~~ i see, it was moved into `set_stat()`.

code/modules/mob/living/living.dm
```diff
@@ -10,7 +10,13 @@
 		name = "[name] ([rand(1, 1000)])"
 		real_name = name
 	faction += "[REF(src)]"
-	GLOB.mob_living_list += src
+	GLOB.mob_living_list |= src
```
noelle-lavenza: `|=` is more expensive than `+=` because it checks for list membership, it's O(n) rather than (roughly) O(1), which means as these lists get longer it will be more and more expensive.

if you really must prevent duplicates, use a var on the mob that's set to TRUE when the mob is added to the list and FALSE when removed, and don't add it if that var is TRUE.

code/modules/mob/mob.dm
```diff
@@ -89,9 +89,9 @@ GLOBAL_VAR_INIT(mobids, 1)
 	mobid = "mob[GLOB.mobids]"
 	GLOB.mobids++
 	if(stat == DEAD)
-		GLOB.dead_mob_list += src
+		GLOB.dead_mob_list |= src
 	else
-		GLOB.alive_mob_list += src
+		GLOB.alive_mob_list |= src
```
noelle-lavenza: see prior comments about `|=`, it is not free, these changes have non-negligible performance impacts especially with our map size and playercount.

code/modules/mob/living/living.dm
```diff
@@ -1930,13 +1971,6 @@
 		if ("maxHealth")
 			if (!isnum(var_value) || var_value <= 0)
 				return FALSE
-		if("stat")
-			if((stat == DEAD) && (var_value < DEAD))//Bringing the dead back to life
-				GLOB.dead_mob_list -= src
-				GLOB.alive_mob_list += src
-			if((stat < DEAD) && (var_value == DEAD))//Kill he
-				GLOB.alive_mob_list -= src
-				GLOB.dead_mob_list += src
```
noelle-lavenza, line 1939: ~~i figure this should just call `set_stat(var_value)`.~~ huh, it was moved up to `/mob`-level, okay.

noelle-lavenza (changes requested):

code/controllers/subsystem/server_maint.dm
```diff
@@ -74,6 +80,193 @@ SUBSYSTEM_DEF(server_maint)
...
+	var/duplicate_count = 0
+	var/invalid_entry_count = 0
+	var/qdeleted_count = 0
+	var/membership_issue_count = 0
+
+	for(var/mob/candidate as anything in GLOB.mob_list)
+		if(isliving(candidate))
+			living_candidates[candidate] = TRUE
+
+	for(var/list_name in lists_to_check)
```
noelle-lavenza: this is a horrendous amount of iteration to be doing every 60 seconds, these lists can get rather large on a live server, and this bookkeeping is not actually that relevant to optimization anyway.

code/controllers/subsystem/server_maint.dm
```diff
@@ -74,6 +80,193 @@ SUBSYSTEM_DEF(server_maint)
...
 
+/datum/controller/subsystem/server_maint/proc/record_mob_processing_issue(list/issues_by_ref, datum/entry, issue)
+	if(!entry)
+		return
+
+	var/ref_id = REF(entry)
+	var/list/issue_data = issues_by_ref[ref_id]
+	if(!issue_data)
+		issue_data = list(
+			"entry" = entry,
```
noelle-lavenza: what's the point of binning them by `REF(entry)` if you're going to use a direct reference anyway? that means it can still cause harddels, *and* it will have the high string operation costs of `REF()` due to triggering string tree rebalancing.

noelle-lavenza (changes requested):

code/controllers/subsystem/mobs.dm
PatapoIIIa: If you remove the !ckey, hibernation will freeze the bodies.

code/controllers/subsystem/mobs.dm
PatapoIIIa: I agree that adjacent-level activation isn't an ideal final solution.

It is currently a conservative workaround to prevent mobs involved in cross-z interactions from freezing, since reliable event-driven wake-up is not implemented for every mob type yet.

A proper proximity- or event-driven hibernation system is a larger piece of work and should be addressed separately. For now, this is effectively a placeholder. Your concern is valid, and it does need to be addressed in a follow-up.

code/controllers/subsystem/mobs.dm
noelle-lavenza: if you want to not hibernate SSD/ghosted players, you could check `!L.client && !L.mind`. you could also just check `!L.ckey` and remove the `!L.client` one in that case.

PatapoIIIa: Really. Thank you.

noelle-lavenza:
maintainers, please note that (in my opinion, at least) this PR is both not necessary and not up to our quality standards and should not be merged

PR #1858, opened 2026-06-11
Who gave a swift weapon role a str setup? Fixes eastern warrior

AllCorruptingWyrm:
## About The Pull Request

When I did my previous pr I noticed it has the wrong stat spread so consider this atomization.

## Testing Evidence

 yeah
 
## Why It's Good For The Game

The stats you get dont work with that weapon.

## Changelog

:cl:
fix: 2spd swap for the sword class for easterner
/:cl:

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/combat/foreigner/eastern.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/combat/foreigner/eastern.dm
@@ -60,6 +60,8 @@
 				armor = /obj/item/clothing/suit/roguetown/armor/basiceast
 				H.adjust_skillrank_up_to(/datum/skill/combat/shields, 3, TRUE)
 				H.adjust_skillrank_up_to(/datum/skill/combat/swords, 4, TRUE)
+				H.change_stat(STATKEY_STR, -2)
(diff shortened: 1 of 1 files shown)


anconfuzedrock:
"This weapon has a stat that is better for speed users, so all classes that use this weapon need to be really fast" is a protozoan take. Give everyone the exact same statline while you're at it.

AllCorruptingWyrm:
the

The stat block doesnt even work.
What kind of take is this.

"The weapon that you are given doesnt work because you have no str even though its a heavy weapon so you must suffer."

AllCorruptingWyrm:
its not even that SWIFT intent literally relies off of speed.

Xaytan:
Not really? Weapons just don't get the bonus parry/dodge bypass from swift balance if you aren't faster than the person you're attacking.

Swift INTENT doesn't interact with the speed stat in any way.

Venithefox:
okay so there is one major glaring issue here...

the hwando is not this classes only weapon...it also has the choice of a naginata and one other weapon I forgot, I think tanto? unless the PR is specifically for the hwando but even THEN. 20 percent more damage matters on a class that has literally nothing else to its name except for a cool weapon and a straw hat

Kroto-Ker:
I don't care about every role being 'optimized'.

PR #1865, opened 2026-06-11
makes fluted half plate use a fluted cuirass

Gug1234:
## About The Pull Request
Changes the fluted half-plate recipe to use the actual fluted cuirass rather than the the normal plate cuirass.

## Testing Evidence
2 word change.

## Why It's Good For The Game
The recipe is dumb and makes no sense, makes it make more sense.

## Changelog

:cl:
fix: Changes the fluted half-plate recipe to actually use a fluted cuirass.
/:cl:

### DIFF
--- a/code/modules/roguetown/roguejobs/blacksmith/anvil_recipes/armor.dm
+++ b/code/modules/roguetown/roguejobs/blacksmith/anvil_recipes/armor.dm
@@ -428,8 +428,8 @@
 	created_item = /obj/item/clothing/suit/roguetown/armor/plate
 
 /datum/anvil_recipe/armor/steel/halfplate/fluted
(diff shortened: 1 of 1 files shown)


Jamdrawers:
I thiiiink it might accept the fluted cuirass anyway without this PR as the fluterd cuirass is still a subtype of cuirass. Could be mistaken though

Gug1234:
> I thiiiink it might accept the fluted cuirass anyway without this PR, since the fluted cuirass is still a subtype of cuirass. Could be mistaken though

It doesn't. They are a subtype but the recipe calls for specifically the normal cuirass in its additional steps. Doesn't permit the fluted variant. Case sensitive.

I wouldn't have made this PR if it worked as expected.

Flyguy715:
Hello, player who plays too much blacksmith here: yes, it is impossible using a fluted cuirass to make fluted halfplate.

PR #1868, opened 2026-06-12
Make send_speech much less expensive

noelle-lavenza:
`send_speech` was written in a horribly inefficient way, this attempts to make it somewhat faster by reducing unnecessary `range()` checks, limiting our player list iteration, etc. We also use `get_hearers_in_range` instead of `get_hearers_in_view` if we aren't LOS-only so that hearing people through walls doesn't require iterating over the player list at all. For all-Z yells, we still have to iterate the player list.

Also makes it clearer that GHOSTWHISPER and GHOSTEARS are an admin-only thing. Needs retesting after this change, opened for CI because I have to go for a bit.

### DIFF
--- a/code/modules/client/preferences.dm
+++ b/code/modules/client/preferences.dm
@@ -2890,7 +2890,7 @@ Slots: [job.spawn_positions] [job.round_contrib_points ? "RCP: +[job.round_contr
 					else
 						user.stop_sound_channel(CHANNEL_LOBBYMUSIC)
 
-				if("ghost_ears")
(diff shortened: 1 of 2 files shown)


noelle-lavenza:
do not TM until it's off draft please

noelle-lavenza:
image

> testing evidence, works cross-z and as a ghost.
> sounds as well, i proccalled `emote("laugh")` on the goblin and heard it from above

noelle-lavenza:
playsound changes have been removed due to bugs, will be separated into another PR

PR #1876, opened 2026-06-13
Optimize mouseover hints by using regex instead of splittext

noelle-lavenza:
Instead of a bunch of separate string and list operations (notoriously slow) to turn the screen_loc into numbers, we use regex to decompose it in one step. This was much, MUCH faster in local testing but needs a testmerge to see if it holds up on live.


noelle-lavenza:
Slower.

PR #1877, opened 2026-06-13
Fix CBT not recompiling on modular code changes

noelle-lavenza:
New top-level code folders (which really shouldn't exist by the way!!) need to be added to this list or CBT won't know the game needs to be recompiled when files in it are touched.


PatapoIIIa:
Nice) Can I have a personal folder? :с

PR #1040, opened 2026-03-09
Vis overlays

PatapoIIIa:
## About The Pull Request

Migrates highly active HUD overlay updates from runtime overlay rebuilding to preallocated vis_contents layers.
Adds reusable HUD component/container support for screen objects, including safer /datum/hud ownership binding and cached human owner access.
Converts zone_select to a reactive vis_contents model with prebuilt slots for limb/base/wound/bleed/selection visuals and reused highlight flash layers.
Converts the blood HUD to persistent child layers for tox/oxy/pain indicators instead of rebuilding overlays during update_health_hud().
Converts hand HUD status overlays to persistent child layers instead of dynamic overlay assembly.
Moves zone_select refreshes off the generic update_health_hud() path onto event-driven updates tied to limb damage, wound changes, bandaging, embedding, dismemberment/reattachment, and manual zone selection.
Removes dead legacy zone_select hover-preview code left over from the old overlay approach.
Adds a side-effect-free bleed helper for HUD rendering so zone HUD updates do not mutate bandage state.
## Testing Evidence

At first glance everything worked, I'm tired, boss

## Why It's Good For The Game

Reduces HUD update overhead for very active UI elements, especially combat-heavy limb targeting displays.
Cuts unnecessary UI churn and avoids repeated runtime overlay construction in hot paths.
Lowers the chance of interface lag during combat and other rapid body-state updates.
Makes the HUD code more maintainable by replacing broad rebuilds with targeted reactive updates.

PR #1880, opened 2026-06-13
Optimize zone_sel overlay updates

noelle-lavenza:
Rewrote the zone_sel hud object's dynamic health icon generation to actually be laid out in a way that 1) makes sense and 2) is good for performance. Worked fine in local testing but may need another local test + a maintainer sniff test, I'm not familiar enough with the edge cases (taurs, etc) to test them properly.

### DIFF
--- a/code/_onclick/hud/screen_objects.dm
+++ b/code/_onclick/hud/screen_objects.dm
@@ -1339,59 +1339,55 @@
 	if(!hud?.mymob)
 		return
 
-	icon_state = "[hud.mymob.gender == "male" ? "m" : "f"]-zone_sel"
+	var/gender_tag = hud.mymob.gender == "male" ? "m" : "f"
+	icon_state = "[gender_tag]-zone_sel"
 
 	if(hud.mymob.stat != DEAD && ishuman(hud.mymob))
 		var/mob/living/carbon/human/H = hud.mymob
 		var/list/missing_bodyparts_zones = H.get_missing_limbs()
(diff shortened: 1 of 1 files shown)


PatapoIIIa:
https://github.com/Rotwood-Vale/Ratwood-2.0/pull/1040

I think that atomization of body parts is a better option.

I think the main problem is this:

All zone_sel elements remain a single overlays list.

For example, changing the selected zone calls update_icon(), which then rebuilds the following:
- body and limbs;
- damage and bleeding;
- missing limbs;
- the selected zone's overlay.

Why did you come to this particular optimization?

noelle-lavenza:
This PR is intended to make the zone_sel icon generation less overly-complicated and therefore both faster and easier to read. The taur check needs to be pulled out of the loop because it's wasteful when we know there will only be one matching organ at most, and usually zero. It also unrolls some unnecessary list iteration.

These changes are good to do regardless of how else `zone_sel` is changed and optimized, and unlike #1040 this PR doesn't make it extremely hard to follow what the code is actually doing. That PR can still be considered, but this one is necessary either way.

PatapoIIIa:
> These changes are good to do regardless of how else zone_sel

#1880 optimizes the generation of the overlay appearances list.
#1040 removes the "generate list of overlay appearances" model itself and replaces it with persistent vis_contents layers, which are created once and then updated via update_zone_layers().

noelle-lavenza:
makes it worse LOL

PR #1881, opened 2026-06-13
Save one cast var assignment in is_species

noelle-lavenza:
This is the microopt to end all microopts. is_species gets called SO OFTEN due to dullahan code snowflakery in speech/hearing that it may actually be worth doing things like this. Needs a TM.

image
This just happened on the live server. It's being called excessively but we can at least try to microoptimize it until we figure out what the culprit is.

### DIFF
--- a/code/__HELPERS/mobs.dm
+++ b/code/__HELPERS/mobs.dm
@@ -454,12 +454,8 @@ GLOBAL_LIST_EMPTY(species_list)
 	if(progbar)
 		qdel(progbar)
 
-/proc/is_species(A, species_datum)
-	. = FALSE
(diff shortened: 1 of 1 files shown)


Xaytan:
code/modules/mob/living/carbon/human/species_types/roguetown/dullahan/dullahan_head.dm

If you look at lines 563 through 570, it uuuh...

Iterates through the ENTIRE global player list (Which includes lobby players and ghosts!) and checks if they are a revenant, then checks if they are THE revenant the head belongs to.

Why the head doesn't just use a reference to its owner for that bit I could not possibly tell you?

PR #1883, opened 2026-06-13
Optimize playsound

noelle-lavenza:
Separated out of #1868 due to issues.

Playsound no longer loops over all clients and dead players in the Z level. Instead it gets client-mobs in range/view. This should be much much faster.

Testing evidence:

image

I heard combat sounds from those two NPCs fighting with an SCOM nearby (that's what broke it in the last PR) and while not on a turf directly (which caused other issues).

### DIFF
--- a/code/game/sound.dm
+++ b/code/game/sound.dm
@@ -35,8 +35,7 @@
 	if(!extrarange)
 		extrarange = 1
 	var/maxdistance = (world.view + extrarange)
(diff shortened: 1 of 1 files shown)


PatapoIIIa:
https://github.com/Rotwood-Vale/Ratwood-2.0/pull/1885 

You might want to review the requests together.
I would be more comfortable if my PR was considered AFTER this.

PR #1885, opened 2026-06-13
Optimize sound loop handling

PatapoIIIa:
## About The Pull Request

Issue: The soundlooper subsystem calls `C.update_sounds()` for every client inside its persistent-loop iteration. Since `update_sounds()` also iterates over every active loop, the total cost grows quadratically: active loops² ×  (clients) every half-second.

- Reduces unnecessary sound update calls.
- Prevents looping sounds from overwriting each other on the same channel.

## Before:
image
image

## After:

image
image

## Why It's Good For The Game

Improves performance and prevents broken volume or mute behavior.

## Changelog

:cl:
fix: Fixed looping sounds conflicting with each other
/:cl:


PatapoIIIa:
There is no point in this pull request existing, as it covers fewer cases and is less efficient than 1910

PR #1886, opened 2026-06-13
Small optimizations to sleep healing and heartbeat sound code

Xaytan:
## About The Pull Request

Updates to heart organ to avoid creating new sound objects every life tick for every player unnecessarily.

Added early exit to on sleep stamina damage healing. Previously was being called and unnecessarily, and was also doing a forced health UI update when it wasn't needed.

DISCLOSURE: Disgusting SLOP was involved in the heart code refactor, I tried to use ChatGPT because surely it can handle a straightforward refactor right? Ahahaha?

But it didn't work in testing and I had to rewrite it myself.

## Testing Evidence

Set Jitteriness, confirmed heartbeat. Confirmed it loops.

Use VV panel to set toxin and oxyloss to critical thresholds, confirmed heartbeat and loop.

Went to sleep, applied various damage types. Confirmed healing works as expected.

image

## Why It's Good For The Game

Is faster.

## Changelog

:cl:
refactor: refactored heart on_life
code: added early exit to sleep stamina healing.
/:cl:


code/modules/surgery/organs/heart.dm
```diff
@@ -26,8 +26,23 @@
...
+	var/sound/slowbeat
+	var/sound/fastbeat
+
+/datum/heart_sound_pool/New()
+	..()
+	slowbeat = sound('sound/health/slowbeat.ogg', repeat = TRUE)
+	fastbeat = sound('sound/health/fastbeat.ogg', repeat = TRUE)
+
+/proc/get_heart_sounds()
+	return GLOB.heart_sound_pool
```
noelle-lavenza: i believe we could just make these static vars on the heart organ
```dm
	var/static/sound/slowbeat = sound('sound/health/slowbeat.ogg', repeat = TRUE)
	var/static/sound/fastbeat = sound('sound/health/fastbeat.ogg', repeat = TRUE)
```
then you reference them as if they're normal instance variables, but they're shared for all hearts

code/modules/surgery/organs/heart.dm
```diff
@@ -107,33 +122,39 @@
...
-			if(H.health > HEALTH_THRESHOLD_FULLCRIT && (!beat || beat == BEAT_SLOW))
-				H.playsound_local(get_turf(H),fastbeat,40,0, channel = CHANNEL_HEARTBEAT)
-				beat = BEAT_FAST
-		else if(beat == BEAT_FAST)
-			H.stop_sound_channel(CHANNEL_HEARTBEAT)
-			beat = BEAT_NONE
+			beat = new_beat
+			last_beat_sound = 0
+		//Our oggs are not the same length.
+		var/interval = ((beat == BEAT_SLOW) ? 1200 : 285)
```
noelle-lavenza: not a huge fan of these hardcoded lengths, can we avoid them somehow? we know this will be called every 200 seconds, which means the slow beat will happen every sixth run and the fast one roughly every other one.

code/modules/surgery/organs/heart.dm
```diff
@@ -107,33 +122,39 @@
...
+		if(beat != BEAT_NONE && (last_beat_sound == 0 || world.time > last_beat_sound + interval))
+			last_beat_sound = world.time
+			var/turf/T = get_turf(H)
+			var/datum/heart_sound_pool/pool = get_heart_sounds()
+			switch(beat)
+				if(BEAT_SLOW)
+					H.playsound_local(T, pool.slowbeat, 40, FALSE, channel = CHANNEL_HEARTBEAT)
+				if(BEAT_FAST)
+					H.playsound_local(T, pool.fastbeat, 40, FALSE, channel = CHANNEL_HEARTBEAT)
```
noelle-lavenza: in theory this could be written with fewer checks. you can also remove the `last_beat_sound == 0` check by having last_beat_sound default to `-INFINITY`. and i think we need `>=` or else it would take 7 cycles for the slow heartbeat to run again rather than 6. this assumes you made the static var change i mentioned elsewhere as well
i'm not 100% sure on replacing `T` with `null`, but in theory since the sound is coming from inside the mob it shouldn't be affected by any of the turf effects, and the way to do that is just not passing a turf source.
```suggestion
		var/interval //Our oggs are not the same length.
		var/sound/heartbeat_sound
		switch(beat)
				if(BEAT_SLOW)
						interval = 12 SECONDS
						heartbeat_sound = slowbeat
				if(BEAT_FAST)
						interval = 2.85 SECONDS
						heartbeat_sound = fastbeat
		if(heartbeat_sound && (world.time >= last_beat_sound + interval))
			last_beat_sound = world.time
			H.playsound_local(null, heartbeat_sound, 40, FALSE, channel = CHANNEL_HEARTBEAT)
```

code/modules/surgery/organs/heart.dm
```diff
@@ -26,8 +26,23 @@
 	/// Associated maniac key
 	var/inscryption_key
 
+	var/last_beat_sound = 0
```
noelle-lavenza:
```suggestion
	var/last_beat_sound = -INFINITY
```
this ensures the first check will always succeed without you having to special-case a `last_beat_sound == 0` check

code/datums/status_effects/debuffs.dm
```diff
@@ -122,7 +122,8 @@
 			break //Only count the first bedsheet
 		if(health_ratio > 0.8)
 			owner.adjustToxLoss(healing * 0.5, FALSE, TRUE)
-		owner.adjustStaminaLoss(healing)
+		if(owner.getStaminaLoss()) //How often do we even get staminaloss in this codebase?
+			owner.adjustStaminaLoss(healing, FALSE, TRUE)
```
noelle-lavenza, line 126: fun fact, staminaloss is entirely unused and should really REALLY be just yanked out at this point

noelle-lavenza (changes requested):
great PR, i was actually about to optimize this myself before i remembered you already did it. just a few changes needed to fit our style guidelines and then you're good to go

Xaytan:
> great PR, i was actually about to optimize this myself before i remembered you already did it. just a few changes needed to fit our style guidelines and then you're good to go

Whoa, style guidelines?

Since when has Slopwood had coding standards?! Really cleaning the place up.

code/datums/status_effects/debuffs.dm
Xaytan: Figured, maybe that'll be my next next PR (My next PR is going to involve the character pref mannequins, those things completely rebuild themselves every time you change something in the UI.)

I figured better to leave it in for now, just in case somebody somehow manages to get themselves stamina damaged.

code/modules/surgery/organs/heart.dm
Xaytan: Me neither, I'm testing a couple of alternatives.

Xaytan: Easy solution: Just don't. I don't think it actually looped originally, and honestly it's kindof annoying if it goes on any longer than a single time.

noelle-lavenza (approved):

PR #1906, opened 2026-06-16
Remove the sensory effect system

noelle-lavenza:
It doesn't work well, doesn't work at all for blind characters unless their eyes are closed, and causes a bunch of lag. It's one of the top 5 items by self-cpu on the profiler, both the BYOND profiler and the Tracy one.


noelle-lavenza:
nuclear option may not be necessary, i could switch it to using hearers()

PR #1911, opened 2026-06-16
Prevent unnecessary processing on lights that don't use fuel

noelle-lavenza:
Lights that don't use fuel will stop processing.


Xander3359:
I would fall in love with you if you made it early return instead :)

noelle-lavenza:
> I would fall in love with you if you made it early return instead :)

done

PR #1922, opened 2026-06-17
Woe: Ducal colors upon ye

Randumb-Phantomb:
## About The Pull Request

Makes vanguards, city guards, watch captain, wardens, and the master warden start with tabards of the ducal colors

## Testing Evidence

image

## Why It's Good For The Game

In lore it makes sense for the duke to give their colors to the soldiers of their domain, even the lower ones. It makes it easier in combat to identify who is pro or anti-duchy.

## Changelog

:cl:
balance: all soldiers of the duchy now start with ducal colored tabards.
/:cl:

### DIFF
--- a/code/__HELPERS/cloaksetup.dm
+++ b/code/__HELPERS/cloaksetup.dm
@@ -26,16 +26,49 @@
 			allowed_cloaks = list(
 			"Jupon" = 			/obj/item/clothing/cloak/stabard/surcoat/guard,
 			"Tabard" = 			/obj/item/clothing/cloak/stabard/guard,
-			"Guard hood" = 		/obj/item/clothing/cloak/stabard/guardhood
-			)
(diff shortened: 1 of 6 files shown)


deadmomo69:
I think the Wardens should keep their cloaks. The vanguard and city guard though -- it makes absolute sense.

UntoldTactics:
I like city guards keeping their specific color and it's rather easy to identify them. Their armor doesn't mesh well with most tabards or such ducal color covers.

Jamdrawers:
You can identify vanguards, wardens, city guards and the master warden by their unique cloaks, which is better because not only can you tell they're a guard from a first glance, but you can tell what *kind* of guard they are, rather than having all the different types of the retinue blend into the same soup

Jamdrawers:
(and also because they look cool and characterful and match the colours much better)

PR #1928, opened 2026-06-17
[TM ONLY] Remove update_body_parts_head_only

noelle-lavenza:
Replaces `update_body_parts_head_only()` with `update_body()` temporarily. UBPHO doesn't use the limb cache, it doesn't get queued, it doesn't properly use apply_overlay or remove_overlay, it's just bad overall, boo, we don't like you, go home.

Also, removes those calls entirely from reagents, these don't modify your head icon so they shouldn't be called at all. It's probably just a holdover from a time where they did, and people copied the space drugs code not knowing why those lines existed.

Based on https://github.com/tgstation/tgstation/pull/89015, except our issue isn't from blinking, it's from space drugs. It's probably even worse for us given the limb cache and queueing.


noelle-lavenza:
fuck it yolo

PR #1929, opened 2026-06-17
Daemonslayer str check.dm

lalaria1:
## About The Pull Request
I think someone forgot to add the hard requirement for STR on this sword. It's supposed to be immense and very heavy, so like the maul, people who do not reach the 15 STR requirment won't be able to wield it at all.

Not sure if this is balance or adding more, since I think the original coder missed adding this hard requirement that's present on other "huge" weapons.

## Changelog

:cl:
balance: rebalanced something
/:cl:


Ollanius:
Oopsie-doodles.

PR #1935, opened 2026-06-18
[TM ONLY] Make human limb icon cacheing actually work

noelle-lavenza:
EXTREMELY EXPERIMENTAL, MAY BREAK THINGS

Previously the icon cache was entirely redundant, for reasons you can see here. Trying to just naively re-enable it meant a bunch of bodypart features were broken, and that took a while to fix.

This will make icon cache keys for our sort of very-complex human mobs very long. That might not be worth it at highpop, it's really, REALLY hard to tell. I'd like to maybe try testing this on the secondary server to ensure nothing breaks, but literally no one plays it...

### DIFF
--- a/code/modules/mob/living/carbon/human/update_icons.dm
+++ b/code/modules/mob/living/carbon/human/update_icons.dm
@@ -57,12 +57,6 @@ There are several things that need to be remembered:
 			jazz += 2
 	return jazz
 
-//HAIR OVERLAY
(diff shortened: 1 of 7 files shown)


noelle-lavenza:
there's some issues with equipment and stuff, have to investigate

noelle-lavenza:
this has been testmerged for over a month

PR #1942, opened 2026-06-18
Woundhealjak for Astratans

ophaq:
## About The Pull Request
Gives ASTRATANS (acolytes + missionaries) the wound heal miracle. Just a copy/pasted line change in the patron file. Not a big PR, so hopefully it qualifies under the feature freeze but lemme know if it doesn't.

## Testing Evidence
woundhealastrata

## Why It's Good For The Game
Astratan miracle kit is pretty lacking at the moment. As a class that is second in best to healing compared to Pestra, they could really use it.

Wound heal clears the limb of wounds but it doesn't fix the brute or burn. It pretty much converts wounds into brute, I believe. It is a good miracle that basically just saves a TON of time stitching people up with 10+ slash wounds on one specific part instead of sitting minutes at a time spamming miracle on them. Not to mention it is just one limb. It is especially useful for transplants, arterial bleeds and nearly chopped off heads and skullcracks.

Is it too strong? Maybe. Now compared to before not -all- the acolytes get it. It is -just- the Astratan patron in this PR.

## Changelog
:cl:
add: Astratans (acolytes + missionaries) now have the wound heal miracle.
/:cl:

### DIFF
--- a/code/datums/gods/patrons/divine/abyssor.dm
+++ b/code/datums/gods/patrons/divine/abyssor.dm
@@ -18,6 +18,7 @@
 					/obj/effect/proc_holder/spell/invoked/call_dreamfiend		= CLERIC_T3,
 					/obj/effect/proc_holder/spell/invoked/abyssal_infusion		= CLERIC_T4,
 					/obj/effect/proc_holder/spell/invoked/resurrect/abyssor		= CLERIC_T4,
+					/obj/effect/proc_holder/spell/invoked/wound_heal			= CLERIC_T4,
 	)
 	confess_lines = list(
 		"ABYSSOR COMMANDS THE WAVES!",
--- a/code/datums/gods/patrons/divine/abyssor.dm
+++ b/code/datums/gods/patrons/divine/abyssor.dm
@@ -18,7 +18,6 @@
 					/obj/effect/proc_holder/spell/invoked/call_dreamfiend		= CLERIC_T3,
 					/obj/effect/proc_holder/spell/invoked/abyssal_infusion		= CLERIC_T4,
 					/obj/effect/proc_holder/spell/invoked/resurrect/abyssor		= CLERIC_T4,
-					/obj/effect/proc_holder/spell/invoked/wound_heal			= CLERIC_T4,
 	)
 	confess_lines = list(
 		"ABYSSOR COMMANDS THE WAVES!",
--- a/code/datums/gods/patrons/divine/astrata.dm
+++ b/code/datums/gods/patrons/divine/astrata.dm
@@ -15,7 +15,6 @@
 					/obj/effect/proc_holder/spell/invoked/heal					= CLERIC_T2,
 					/obj/effect/proc_holder/spell/invoked/revive				= CLERIC_T3,
 					/obj/effect/proc_holder/spell/invoked/immolation			= CLERIC_T4,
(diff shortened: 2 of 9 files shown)


The-Sun-In-Splendour:
Isn't wound heal literally an aheal on a limb so long as the limb has some sort of wound?
I think healing is already way too easy as is. We don't need this.

ophaq:
> Isn't wound heal literally an aheal on a limb so long as the limb has some sort of wound? I think healing is already way too easy as is. We don't need this.

It clears the limb of wounds but it doesn't fix the brute or burn. It pretty much converts wounds into brute, I believe. It is a good miracle that basically just saves a TON of time stitching people up with 10+ slash wounds on one specific part instead of sitting minutes at a time spamming miracle on them. Not to mention it is just one limb. It is especially useful for transplants, arterial bleeds and nearly chopped off heads and skullcracks. Also, the patient still experiences pain after a while, despite the wounds being healed. I'm not sure if that was intended or a bug but they will pain crit unless they rest a bit even after getting healed from their major wound.

I'm willing to reduce the patrons that have access to it. This is a starting point. I think Astratans, Eorans, and Pestrans should have it at the very least.

HalfAwake691:
Did this not used to be a thing? It was removed because having everyone running around with wound healing was really strong. Like in a hyper war you'll have to start decapping people instead of skull cracking because you don't know if a healer will run up and insta heal their head putting them back into the fight. I could just be remembering azure but giving every healer an aheal is weird.

It does aheal the wound but converts the damage into PAIN.

zeroisthebiggay:
yeah not for this

Sgt-Solid01:
This miraxle HEALS a limb that has wounds and also fixes wounds instantly, including skullcracks

Do not merge this, very very bad idea

ophaq:
The miracle isn't THAT strong (you still have to heal up the brute damage caused by all the wound closing) but I understand it would be too strong during hyperwars if -every- acolyte had it.

So I am just going to give it to only -Astratans- instead since their kit is pretty lacking at the moment compared to other acolytes... and they're the dedicated revival and healer other than Pestrans.

TamedAchilles:
I guess my question is,

Why do Astratan's need this aside from just beefing their miracle portfolio up. Pestran's have and will always be to the go to for wound healing. Giving another Goddess' miracle to them seems off, as if we're bridging a gap that shouldn't be bridged since Astrata isn't the Goddess of Medicine lol

ophaq:
> I guess my question is,
> 
> Why do Astratan's need this aside from just beefing their miracle portfolio up. Pestran's have and will always be to the go to for wound healing. Giving another Goddess' miracle to them seems off, as if we're bridging a gap that shouldn't be bridged since Astrata isn't the Goddess of Medicine lol

It's not another goddess's of medicine though. I don't think Pestrans get this either. It's an old miracle that was removed and forgotten.

It is very handy for healing. Astrata may not be for healing but they're one of the only ones that can give fortitude for increased healing ability. I don't really get the hate behind the spell anyway. It lets you clear -one- single limb of all wounds and replaces it with brute damage which you still have to heal back. And it has a -long- cooldown. You can just spam 5 or 6 miracles that recovers the whole body in the time it takes to go back off cooldown.

ophaq:
Im gonna close this for since more people would prefer the status quo for now! Maybe if enough people request such a thing I'll reopen but I don't see it happening.

PR #1946, opened 2026-06-18
Make Zizite Knight Migrant able to properly train their squire. 

Gug1234:
## About The Pull Request
Ziziod Knights get an objective to train their squire, the small issue being that they only have expert swords and thus cant actually train their squire to anything above journeyman... which is what they start with. This changes that and also swaps their scabbard from a longsword one to a greatweapon strap so they can actually store their flamberg. 

## Testing Evidence
2 line change

## Why It's Good For The Game
Makes the ziziod knight able to do their objective properly, brings them up to speed with the keep and other wretches skill-wise.  Also unfucks their loadout. 

## Changelog

:cl: Yuckuza
balance: increased Zizite Knight Migrant sword skill to 5, so they can train their squire. 
fix: changed the zizite knight's scabbard to a great weapon strap so they can store their sword properly.
/:cl:

### DIFF
--- a/code/modules/antagonists/roguetown/villain/dark_itinerant.dm
+++ b/code/modules/antagonists/roguetown/villain/dark_itinerant.dm
@@ -110,7 +110,7 @@
 	shoes = /obj/item/clothing/shoes/roguetown/boots/blacksteel/modern/plateboots
 	beltl = /obj/item/flashlight/flare/torch/lantern
 	belt = /obj/item/storage/belt/rogue/leather/steel/tasset
-	beltr = /obj/item/rogueweapon/scabbard/sword
+	backl = /obj/item/rogueweapon/scabbard/gwstrap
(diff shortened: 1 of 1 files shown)


zeroisthebiggay:
might be wiser to give them `TRAIT_GOODTRAINER` rather than creating another antagonist role that has master skills

Gug1234:
> might be wiser to give them `TRAIT_GOODTRAINER` rather than creating another antagonist role that has master skills

Maybe. It's a rare enough role, I think it warrants the buff, so it's on par with every other knight antagonist: hedgeknight, and deserter knight, being the standouts.  
If we ever do a broad pass over all our classes to bring their skills and stats in line, expert would make more sense, but as it stands currently, the blacksteel ziziod knight has the same weapon skill as wardens and garrison when it is by all accounts a duo antag role between them and their squire.

PR #1954, opened 2026-06-19
Stun/Immobilize Var to end stuns early from dropped pins

SneakySlinky:
## About The Pull Request

Adds a `var` to store any existing stun/immobilize durations (likely zero), so that when a pin is ended early by letting it go in some way, the duration is put back to what it was prior to the pin (still likely zero.)

## Testing Evidence

dreamseeker_Zv8LyR4Dd7

Hard to show this directly, but the gist of testing was:

Made two mobs, first tested the pin working period by grabbing the other, pinning it, and direct controlling while the pin was held. Worked.

Then, tested the pin stun ending early if let go by doing the same thing, but letting go before direct controlling. Also worked. _Mostly._ The visual indicator still shows it being there, but the stun effect itself stops.

## Why It's Good For The Game

With a combination of high STR and Wrestling skill, it became possible to effectively do upwards of an *eight* second "ghost" pin where by starting a pin and immediately dropping it, the person would be stuck to the ground while their 'pinner' was free to do anything else. Just a weird interaction overall.

## Changelog

:cl:
fix: Pins no longer magically 'held' long after the grab is let go
/:cl:

### DIFF
--- a/code/modules/mob/living/grabbing.dm
+++ b/code/modules/mob/living/grabbing.dm
@@ -343,6 +343,10 @@
 				var/stun_dur = max(((65 + (skill_diff * 10) + (user.STASTR * 5) - (M.STASTR * 5)) * combat_modifier), 20)
 				var/pincount = 0
 				user.stamina_add(rand(1,3))
+				var/pin_stun_baseline = M.AmountStun()
+				var/pin_stun_applied = FALSE
+				var/pin_immob_baseline = M.AmountImmobilized()
+				var/pin_immob_applied = FALSE
 				while(M == grabbed && !(M.mobility_flags & MOBILITY_STAND) && (src in M.grabbedby))
 					if(M.IsStun())
 						if(!do_after(user, stun_dur + 1, needhand = 0, target = M))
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/grabbing.dm
```diff
@@ -354,17 +358,26 @@
...
+						pin_immob_applied = TRUE
 						user.stamina_add(rand(1,3) + abs(skill_diff) + stun_dur / 1.5)
 						pincount += 2
 						M.visible_message(span_danger("[user] pins [M] to the ground!"), \
 							span_userdanger("[user] pins me to the ground!"), span_hear("I hear a sickening sound of pugilism!"), COMBAT_MESSAGE_RANGE)
+				if(!(src in M.grabbedby))
+					if(pin_stun_applied)
+						M.SetStun(pin_stun_baseline)
+					if(pin_immob_applied)
+						M.SetImmobilized(pin_immob_baseline)
```
Arturlang, line 380: are you sure this won't just let you clear all stuns if you pin someone and let them go? Better to make a subtype of datum/status_effect/incapacitating or stun, that can be removed once the grabhand is dropped or grab is canceled somehow else, just make sure to make it stun resist flags since annoyingly those are handled in the setters

Arturlang (changes requested):

code/modules/mob/living/grabbing.dm
Arturlang: alternative is checking the amount of stun you actually apply and using adjust procs instead, tricky since you need to kinda account for time

SneakySlinky: This will clear the stun/immobilization placed BY pin itself, and recall the stored durations, as per the variable.
In most situations, this is going to be stored as '0' because pinning someone who is simply on the ground will just store their duration as 0, to be called back when the grab is dropped/ended.

Edit: Wait.

SneakySlinky:
The more I stare into grappling code and thought myself clever to fix it, I realize it only makes more and more issues due to how many awkward situations you can wind up in WHILE attempting to grapple and/or pin someone... 

I'll try this again later, maybe.

PR #1956, opened 2026-06-19
Optimize animal food checks

noelle-lavenza:
It's time for some *fast food!* Ha, get it? Get it?

Replaces these uses of `is_type_in_list` with a typecache lookup. I know it works in testing because the FUCKING CHICKEN ate the oats I spawned before I could even pick it up, and feeding more oats to it manually also worked. Woo.


Trexdude:
Thank you for optimizing this! It feels good to see.

PR #1971, opened 2026-06-21
Loose Straps Fix

Lokaka21:
## About The Pull Request
Currently, Loose Straps seemingly has the unintended side effect of allowing your armor to be destroyed if hit sufficiently hard whilst being flung at the same time. This should make it nigh-impossible to occur, by virtue of making clothing itself much more difficult to destroy save via fire.

Partial port of https://github.com/Azure-Peak/Azure-Peak/pull/6911, primarily just the aspect that clamps damage done to broken clothing.
## Testing Evidence
Attempted to replicate the issue described to me with regards to losing armor pieces after implementing changes, have not been successful. Will be happy to test further if deemed necessary.

## Why It's Good For The Game
In essence what it actually does is make it much harder to willfully destroy **broken** armor in general (again, save by fire.) But it also makes Loose Straps function as intended!


Randumb-Phantomb:
This is true!

PR #1973, opened 2026-06-22
Fix/Bal: HF Knight Weapon Choice reflection

Teragenic:
## About The Pull Request
Heartfelt Knight:
Currently its a Claymore. Fixes the "Zweihander" option to be actually be a Zweihander

Adds a normal Greatsword option as well.

Weapon option actually reflects choice.
## Why It's Good For The Game

Makes weaponchoice reflect choice.

## Changelog

:cl:
add: Adds Greatsword option to HF Knight
fix: Makes the Zweihander option for HF Knight a Zweihander.
/:cl:


petwir1:
Would be great to have the regular Duchy heavy knights have the same options / switching out their claymore for a real steel greatsword.

PR #1976, opened 2026-06-22
inq weapons balancejak

The-Sun-In-Splendour:
## About The Pull Request
daybreak and silver whips now require 10str instead of 12
eucharist force 20 -> 25
the creed is no longer a subtype of the forgotten blade, but now a psydonian kriegmesser

## Testing Evidence
works on my machine.

## Why It's Good For The Game
these 3 weapons are complete dogass and never used because the other options are just better in everyway
encourages inquisitors to choose different weapons than just halberd(inspector) and or flail(ordinator)
creed was just a really really mid gsword and made you drop your shield or your satchel to carry (since it only fit in a greatsweapon strap)
daybreak needing 12 str made it basically anathema to inspectors because all of the +2str statpacks kinda fucks with the stats the class wants
eucharist is because rapiers are all criminally underused and also because halberd is just better (dodge experts without shields really would rather have a weapon they can WEILD)

## Changelog

:cl:
bal: daybreak and silver whips now only require 10str to use rather than 12str
bal: eucharist's force increased from 20 to 25
bal: creed is now a psydonian kriegmesser
/:cl:


petwir1:
It always struck me as odd that the classes that have access to these whips are really disadvantaged trying to use them. Good change.

Gug1234:
You and I were chatting about this on Discord, but it could be fun to have it be a subtype of executioner sword with its one-handed damage kept at 20. Lets it fit in a scabbard, but still fit the ordinators' vibe + give them a choice between the two predominant heavy sword options we have.

The-Sun-In-Splendour:
closing this until my psydonic refactor gets in (i found out there is an actual silver kriegmesser in the game, and psydonic one could be a subtype of it)

PR #1986, opened 2026-06-23
Fix duplicate lighting updates on move

noelle-lavenza:
We called update_light twice, once via signal and once via this. Waaaagh

### DIFF
--- a/code/modules/lighting/lighting_atom.dm
+++ b/code/modules/lighting/lighting_atom.dm
@@ -133,13 +133,7 @@
 		T.opaque_atom_count--
 	T.reconsider_lights()
 
-/atom/movable/Moved(atom/OldLoc, Dir)
-	. = ..()
-	var/datum/light_source/L
-	var/thing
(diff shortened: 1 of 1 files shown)


code/modules/lighting/lighting_atom.dm
```diff
@@ -133,13 +133,7 @@
...
 	T.reconsider_lights()
 
-/atom/movable/Moved(atom/OldLoc, Dir)
-	. = ..()
-	var/datum/light_source/L
-	var/thing
-	for (thing in light_sources) // Cycle through the light sources on this atom and tell them to update.
-		L = thing
-		L.source_atom.update_light()
+// No /atom/movable/Moved override here, that's handled by a signal now.
```
Xander3359, line 136: comment referencing dead code is not needed.

Xander3359 (changes requested):

code/modules/lighting/lighting_atom.dm
noelle-lavenza: it is needed so that no one tries to readd it later thinking we overlooked it

PR #1987, opened 2026-06-23
Optimize reg_in_areas_in_z

noelle-lavenza:
Makes this proc faster by eliminating the area contents loop that was entirely pointless


noelle-lavenza:
may be better to port the area contents system and do it that way, since the current way areas_in_z works is wrong anyway and only tracks the first z-level

PR #1990, opened 2026-06-23
Make sunlight and weather effects use underlays instead of objects

noelle-lavenza:
EXPERIMENTAL CHANGE.

The rationale here is that we can reduce maptick this way, by cutting down from nearly every tile having an object dedicated to just the weather effects, to, well... none of them having it. We use underlays.

I am almost certain this is far more effective than making lighting use underlays, since sunlight updates way less.

We also share one global weather mutable appearance for all turfs, because they were literally all identical. Probably doesn't do much because appearances are deduplicated and stored in a tree internally, but like, it can't hurt, can it? It's also necessary in order to make the underlay updates work right, I think...

Also, makes the second Z of roguetest use the outdoor area so that it actually has weather and sunlight, for the purpose of testing those. Yippee.


Arturlang:
have a solution that uses render-targets and overlays on sunlit turfs

noelle-lavenza:
> have a solution that uses render-targets and overlays on sunlit turfs

that's what this already does, just with underlays instead of overlays because underlays get mutated less often.

PR #1993, opened 2026-06-23
ERP Hearts (and plap text)

WolfOrion:
## About The Pull Request
Ports ERP hearts and 'Plap!' text from Azure Peak. There is also anger emoticons if you have combat mode on. It respects subtle actions as well and will not display them during such.

## Testing Evidence
The hearts and plap text
image

Anger emoticons with combat mode enabled by one person but not the other.
image

## Why It's Good For The Game
It makes mechanical ERP more visual beyond just one sprite moving onto another until a white decal appears underneath one of them.

## Changelog

:cl:
add: Added heart and anger emoticons during non-subtle ERP actions
add: Added 'Plap!' balloon notifications to many ERP actions
/:cl:


UntoldTactics:
Foundational load-bearing PR required for all intents and purposes.

ophaq:
We don't need content on this codebase, just ERP improvements and code optimizations.

All jokes aside, cool PR. Is there an option to toggle it off if we get tired of seeing hearts?

Skyglazer:
A toggle would be nice.

Xander3359:
Fully agree, add a toggle to this.

zeroisthebiggay:
good pr, github user @WolfOrion !

WolfOrion:
Added a preference check, and it works sort of, but it probably needs more work and I'm not sure how to do that

ophaq:
I think that the to chat message where it says 'creates a heart' isn't really necessary in my opinion. It just adds more spam to the chat log as it's constantly created every few seconds as you're verb spamming.

WolfOrion:
> I think that the to chat message where it says 'creates a heart' isn't really necessary in my opinion. It just adds more spam to the chat log as it's constantly created every few seconds as you're verb spamming.

That was indeed a dev test message that I accidentally left in because I have been extremely tired as of late.

WolfOrion:
After sleeping on it, I still cannot think of any way to make the preference check work both ways, and I don't want to think about this anymore, so I'm closing it.

Someone else can pick up the mantle later.

PR #1999, opened 2026-06-24
Fix ghost protection blocking scrying

noelle-lavenza:
untested. also everything using ghost protection needs to be rewritten to not use get_hidden_ghosts_for_target because that is a lag monster


petwir1:
thank you, this is something I had been noticing being an active CM player.

Sgt-Solid01:
I've made a report on this, i love you for fixing it

PR #2005, opened 2026-06-24
Put catchpole maul option on equal grounds to cityguard maul. 

Gug1234:
## About The Pull Request
City guard maul option allowed most players to take the maul option and not be immediately shafted because they didn't meet the 14 strength minimum by granting +1 strength to their 2 strength stat, which any player with a strength statpack could utilize to play with a maul or grand maul if taking muscular, thuggish, or struggler. 

Catchpole, on the contrary, has no such buff, meaning players had to choose the above three statpacks and couldn't use a grand maul unless playing as Drakian or Lamia. This PR tweaks the catchpole maul option to work much the same as cityguard's option does, giving +1 strength at the cost of -1 int and speed. 

TLDR:
Adjusted maul option for catchpole

- +1 strength
- -1 intelligence
- -1 speed
- can use grandmaul with +2 statpack

## Testing Evidence
easy peasy 3 line change taken directly from townguard.dm

## Why It's Good For The Game
Less statpack restrictions on catchpole, more variety for catchpole maul users.

## Changelog

:cl: Yuckuza
balance: Catchpole maul option gets +1 strength at the cost of -1 int and speed in line with cityguard's maul option.
/:cl:

### DIFF
--- a/code/modules/jobs/job_types/roguetown/garrison/manorguard/gormless.dm
+++ b/code/modules/jobs/job_types/roguetown/garrison/manorguard/gormless.dm
@@ -50,6 +50,9 @@
 			if("Maul - 14STR Minimum")
 				r_hand = /obj/item/rogueweapon/mace/maul
 				backl = /obj/item/rogueweapon/scabbard/gwstrap
+				H.change_stat(STATKEY_STR, 1) /// so even without a strength-specific statpack, you can at least use the weapon. With one, it allows great maul just like on city guard.
+				H.change_stat(STATKEY_SPD, -1)
+				H.change_stat(STATKEY_INT, -1)
 
 	backpack_contents = list(//Iron dagger and ale instead of red.
 		/obj/item/rogueweapon/huntingknife/idagger = 1,


Jamdrawers:
I was surprised it didn't already work like that by default

Originally I was gonna just remove catchpole from MAA because it'd be in city guard instead but I knew some people would be attached to it.

Gug1234:
> I was surprised it didn't already work like that by default
> 
> Originally I was gonna just remove catchpole from MAA because it'd be in city guard instead but I knew some people would be attached to it.

I was so disappointed that I couldn't use the grand maul as an MAA while looking over the stats. The people crave their bigger, funnier hammer.

UntoldTactics:
This is a neat PR but honestly city guard and MAA need a total 'reorientation' - funnily enough in many ways the city guard are better skilled and equipped than MAAs. 

Notably, MAAs overall get journeyman shields at best, whilst city guard by default all get expert shields etcera. The general 'creep' of ratwood balance and the fact that city guard are more 'latest' compared to MAAs has led to them being more well-rounded in the current balance. It's a bit sad.

PR #2013, opened 2026-06-25
fixes lacrima not checking if you're actually dead

The-Sun-In-Splendour:
## About The Pull Request
what it says on the tin

## Testing Evidence
tested on my machine and works on my machine

## Why It's Good For The Game
necromancers were being rerally strong lately not because they were insanely strong but because there is a bug that lets them get 50 times more lux than they normally should be able to get

## Changelog

:cl:
fix: lacrima's "is dead" check
/:cl:


Xander3359:
You still need the old check.

The-Sun-In-Splendour:
> You still need the old check.

done

PR #2017, opened 2026-06-26
Fix flavortext color parsing

noelle-lavenza:
Makes flavortext color parsing less laggy. Confirmed elaborate FTs that formerly caused lag still work and don't spike tidi on local:
image


noelle-lavenza:
image

multiline colors in flavortext fixed with latest commit

PR #2018, opened 2026-06-26
[TM ONLY] Disable parallax TEMPORARILY

noelle-lavenza:
Disables parallax to see if that helps at all with lag.


noelle-lavenza:
death to all perma-testmerges

PR #2023, opened 2026-06-26
Make get_bodypart use a zone lookup list

noelle-lavenza:
Instead of using a for loop, we do a single lookup in an alist. Easy peasy. Did a bunch of combat to test, everything seemed fine. TM is still nice to have though.


noelle-lavenza:
For some bizarre reason only the lamia tail I tested works, and all other taur limbs don't. Fuck.

noelle-lavenza:
this should work now

PR #2037, opened 2026-06-29
adds fat-related content to the game (lie) (it's actually tallow)

JustMeTheIInd:
## About The Pull Request

This PR adds tallow, and tallow pots to every role that starts with a custom seal (and the court chaplain because it makes sense for him to have one)

It also removes the calendula bouquet from the Merchant's house on dunworld which I predict will lead, in the future, to the total and entire disruption of balance, finally culminating in the long-awaited 'Total Antag Death' Event and the final AP-ification of RW. I think.

## Testing Evidence

It's a mapping change, but if the map fucking explodes because I accidentally put an out of quote 's' somewhere I will fucking kill myself irl.

## Why It's Good For The Game

I think we should be able to make use of in-game features. (controversial opinion)

## Changelog

:cl:
map: Adds tallow and tallowpots to all the roles that get their own special little seals. And the Court chaplain.
/:cl:

### DIFF
--- a/_maps/map_files/dun_world/dun_world.dmm
+++ b/_maps/map_files/dun_world/dun_world.dmm
@@ -3846,6 +3846,8 @@
 "aNd" = (
 /obj/structure/table/wood/long_table/right,
 /obj/machinery/light/rogue/candle/floorcandle,
+/obj/item/inqarticles/tallowpot,
+/obj/item/reagent_containers/food/snacks/tallow/soft,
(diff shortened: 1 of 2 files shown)


Dumplinz:
Just wanted to note, if this could also have a tallow pot and candle added to the diplomacy area in kingsfield

JustMeTheIInd:
@Jamdrawers can you take a look at this? it's been open for 3+ weeks now but all it does is add tallow...

PR #2040, opened 2026-06-30
"Why didn't you EXPLODE?" - Makes bombs actually useful

annikaRU:
makes bombs less of a joke

## About The Pull Request

Be me

Make four Satchel bombs

Put them in a ring to explode someone

By some miracle they survive with zero damage done to them with zero armor.

## Testing Evidence

2 Line change, compiles on local host

## Why It's Good For The Game

Currently bombs kinda suck, they don't even delimb you while impact grenades will, this makes it so they WILL delimb you no matter what, if you get hit by a satchel bomb...I don't think you should shrug that off.

## Changelog

:cl:
fix: The supplier of gunpowder to the duchy has been executed for high treason following faulty gunpowder mix, satchel bombs and bomb sticks will now properly explode people
change: the supply of longer fuses has ran out resulting in shorter times to explode for blastsand sticks, a more volatile gunpowder mix has been included in the sticks, much lighter and able to be thrown farther
/:cl:

### DIFF
--- a/code/game/objects/items/rogueitems/bombs.dm
+++ b/code/game/objects/items/rogueitems/bombs.dm
@@ -300,7 +300,7 @@
 	slot_flags = ITEM_SLOT_HIP
 	throw_speed = 0.5
 	throw_range = 3
-	var/fuze = 7.5 SECONDS
+	var/fuze = 5.5 SECONDS
 	var/lit = FALSE
 	var/prob2fail = 1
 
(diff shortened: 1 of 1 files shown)


UntoldTactics:
The humble slight bit of spitfire that turns any bomber into a puddle:

annikaRU:
mistakes have occured...

AranyaPTH:
I like the changelog comment, it reads like Ea-nāṣir was our gunpowder supplier.

zoktiik:
Reverse Gunpowder Plot.

People should probably know the fuse length decrease and power increase in changelog, that is the difference in whether or not you blow yourself up

annikaRU:
> Reverse Gunpowder Plot.
> 
> People should probably know the fuse length decrease and power increase in changelog, that is the difference in whether or not you blow yourself up

Good catch, I will put that there

PR #2043, opened 2026-07-01
Fixes "failure to fire" caused by beginning charge on Mob, moving cursor off Mob, then returning and upclicking on Mob

PPooch:
## About The Pull Request

I ran into this bug on my bow hunter. When I started "charging" on a mob, and they or I moved and my cursor departed off of the sprite of the target mob, then I returned the cursor to the mob to "unclick" and release the arrow - nothing happened.

Further testing could reproduce this with any atom, turf etc. This also occurred with buff spells and invisibility. 

This PR puts a short circuit in MouseUp that should reliably bypass the mob-drag smear when using charged casts.

It also bypasses it if the mob is using a ranged attack. Seems to work ok in testing.
## Testing Evidence
https://cdn.discordapp.com/attachments/1338300314683441324/1521700483205435503/arrowtest2.mp4?ex=6a45c983&is=6a447803&hm=b6941e6e142f72e7a71dab7cda5e19e7995a57a1573a5e1ea7f9ccd21c19f70d&

## Why It's Good For The Game
disclosure, AI generated code (Opus 4.8). This can be done better by those more familiar

## Changelog

:cl:
fix: Fixed a bug where starting a spell on a mob, then the cursor moving off the mob, would cause that spell to no longer work when released on the mob.
fix: Same as above but also for arrow targets when firing a bow.
/:cl:

### DIFF
--- a/code/_onclick/drag_drop.dm
+++ b/code/_onclick/drag_drop.dm
@@ -240,7 +240,7 @@
 
 	if(tcompare)
 		var/atom/target_atom = object
-		if(istype(target_atom) && tcompare != mob && (mob.atkswinging == "middle" || (mob.atkswinging && object != tcompare)))
+		if(istype(target_atom) && tcompare != mob && (mob.atkswinging == "middle" || mob.used_intent?.tranged || (mob.atkswinging && object != tcompare)))
 			target_atom.Click(location, control, params)
 		tcompare = null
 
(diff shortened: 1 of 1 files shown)


Xander3359:
Gonna nuke this simply for AI usage. But I think we've identified the problem and it should be fixed by https://github.com/Rotwood-Vale/Ratwood-2.0/pull/2044

PR #2062, opened 2026-07-04
Increase the recharge time of raise lesser skeleton so one necromancer can't spawn 492 skeleton simple mobs in the span of maybe 4 skirmishes.

Gug1234:
## About The Pull Request

- Upped recharge time for raise lesser skeleton formation from 20 seconds to 60

## Testing Evidence
number change

## Why It's Good For The Game
image_2026-07-03_200523220

The above figure is from a recent round, maybe 3 big fights where the spell was used. Each summon is 3 simplemobs, so 492 skeletons were spawned over those three scuffles. 

Zizo's greatest bodyblockers can be spawned far faster and far more cheaply than the actual forcewall spells can, not to mention this wall fights back with mobs immune to sunder wounds. 

Also I died to it once. 

## Changelog

:cl: Yuckuza
balance: increased raise lesser undead spell recharge from 20 seconds to 60
/:cl:

### DIFF
--- a/code/modules/spells/roguetown/necromancer.dm
+++ b/code/modules/spells/roguetown/necromancer.dm
@@ -84,7 +84,7 @@
 	chargedloop = /datum/looping_sound/invokegen
 	gesture_required = TRUE // Summon spell
 	associated_skill = /datum/skill/magic/arcane
-	recharge_time = 20 SECONDS
+	recharge_time = 60 SECONDS
 	var/cabal_affine = FALSE
 	var/is_summoned = FALSE
 	var/to_spawn = 4
@@ -136,7 +136,7 @@
 /obj/effect/proc_holder/spell/invoked/raise_undead_formation/necromancer
 	cabal_affine = TRUE
 	is_summoned = TRUE
-	recharge_time = 35 SECONDS
+	recharge_time = 60 SECONDS
 	to_spawn = 3
 
 


wikimody:
Maybe 40s instead? Its to fast currently but i think 2 times slower is more fair change then 3 times slower

HalfAwake691:
I love over correcting balancing pr's i love the fact we cant do slight changes to anything.

Gug1234:
> Maybe 40s instead? Its to fast currently but i think 2 times slower is more fair change then 3 times slower

force wall is 100, 60 is more than reasonable

HalfAwake691:
> > Maybe 40s instead? Its to fast currently but i think 2 times slower is more fair change then 3 times slower
> 
> force wall is 100, 60 is more than reasonable

image

No its not

The-Sun-In-Splendour:
god forbid necromancers actually necromance ppl eith human wave tactics. nooo we have to nerf it so the garrison can be better human wave tactics than the fucking necromancer.

zeroisthebiggay:
they're a little annoying but it really just takes one guy on cleanup duty with a greatsword to gigakill them and render them worthless
a little bit of extra cooldown at most but sixty seconds is way too long for how little value they usually generate during fights

Dagonlives:
Gotta love these overreactive P.Rs that render Zizo's only useful cleric spell aside Rituos useless.

Garrison will be better at human wave tactics then the necromancers.

UntoldTactics:
I think a bit of extra cooldown is warranted but generally speaking the only time the simplewave swarms become relevant is on lowpop. (And we all know lowpop necromancers are usually scum of the earth anyway), otherwise they're a joke.

PR #2071, opened 2026-07-06
Removes the Forced Outlander Trait on Migrants

Teragenic:
Some Migrants are part of Ferencia and should probably not have it.

This is literally just for I believe Heartfelt and Fablefield.

## About The Pull Request

Some Migrants are part of Ferencia and should probably not have it.

This is literally just for I believe Heartfelt and Fablefield.

## Testing Evidence

3 Line Change

## Why It's Good For The Game

Expectations

## Changelog

:cl:
fix: Migrants being forced to be Outlanders
/:cl:

### DIFF
--- a/code/controllers/subsystem/migrants.dm
+++ b/code/controllers/subsystem/migrants.dm
@@ -285,7 +285,6 @@ SUBSYSTEM_DEF(migrants)
 	to_chat(character, span_notice(wave.greet_text))
 	to_chat(character, span_notice(role.greet_text))
 
-	ADD_TRAIT(character, TRAIT_OUTLANDER, TRAIT_GENERIC)
 
 	if(role.outfit)
 		var/datum/outfit/outfit = new role.outfit()
--- a/code/modules/antagonists/roguetown/villain/dark_itinerant.dm
+++ b/code/modules/antagonists/roguetown/villain/dark_itinerant.dm
@@ -97,6 +97,7 @@
 	H.cmode_music = 'sound/music/combat_heretic.ogg'
 	ADD_TRAIT(H, TRAIT_SQUIRE_REPAIR, TRAIT_GENERIC)
 	ADD_TRAIT(H, TRAIT_STEELHEARTED, TRAIT_GENERIC)
+	ADD_TRAIT(H, TRAIT_OUTLANDER, TRAIT_GENERIC)
 	wretch_select_bounty(H)
 
 /datum/outfit/job/dark_itinerant_knight/pre_equip(mob/living/carbon/human/H)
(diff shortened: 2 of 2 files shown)


gr4ts:
I am reasonably certain if someone is moving from a different city to another but it's within the same country, that does not make them migrants.

The difference between a Towner and a Migrant is that one is from Ferentia and the other is not.

RadiantFlash7:
Migrants aren’t meant to be locals. Towners are locals. If you want to be a local migrant, use the resident Virtue!

gr4ts:
> Migrants aren’t meant to be locals. Towners are locals. If you want to be a local migrant, use the resident Virtue!

Wouldn't this just make them bona fied Towners? Why not just pick a towner role and save the virtue slot for something else then?

Teragenic:
> Migrants aren’t meant to be locals. Towners are locals. If you want to be a local migrant, use the resident Virtue!

So, the PR only does it stop the Subsystem from making all Migrants no matter like the role they are Outlanders. This only affects Fablefield and Heartfelt.

gr4ts:
> > Migrants aren’t meant to be locals. Towners are locals. If you want to be a local migrant, use the resident Virtue!
> 
> So, the PR only does it stop the Subsystem from making all Migrants no matter like the role they are Outlanders. This only affects Fablefield and Heartfelt.

So this is ONLY going to affect the characters that spawn through the Migration system then?

Teragenic:
> > > Migrants aren’t meant to be locals. Towners are locals. If you want to be a local migrant, use the resident Virtue!
> > 
> > 
> > So, the PR only does it stop the Subsystem from making all Migrants no matter like the role they are Outlanders. This only affects Fablefield and Heartfelt.
> 
> So this is ONLY going to affect the characters that spawn through the Migration system then?

Yes. Why wouldn't it? I mentioned Migrants and touched only the Migrant Subsystem.

All other migrant waves are still ALSO TRAIT_OUTLANDER on their actual subclass due to being adventurers/actual foreigners.

emsuhr:
Fablefield and heartfelt are also foreigners, even if they live in Ferentia.

The concept of a unified nation isn’t really a thing yet, the people who live in Rockhill/the surrounding countryside are locals. People from other duchies or vassals to the duchy are not locals, they’re foreigners from their respective fief.

RasbeRissa:
As someone who plays a character FROM Heartfelt, I'm actually a tad opposed to this. The way it is now, if the Heartfelt Migration pops while I'm playing said character, we can have a shared history from the word go. This change would actually force a minor discrepancy, and would either need to be overlooked or explained away in some hand-wavey fashion. Foreigner should be applied to anyone who is not native to the duchy specifically.

Otherwise, if "They're from Ferentia!" means they're not a foreigner anymore, EVERYONE who has the Ferentia origin should lose the Foreigner tag.

Side note unrelated to the PR, someone should really port the sub-area origins. I would love to have Heartfelt be my known homeland again, rather than have it JUST be a note in my notes that I hope people read.

Xander3359:
Migrants is in the name lol

PR #2076, opened 2026-07-07
Blood miracle hotfix

zoktiik:
## About The Pull Request
Prevents blood miracle from triggering with existing do_after. Currently if you're in a do_after and you try to use Blood Transfer Miracle, it takes 50 devotion from you and immediately stops.
I don't like that; it's very annoying on Mystagogue!

## Testing Evidence
dreamseeker_fbt8LjHkm2

## Why It's Good For The Game
Fix fix I like fix

## Changelog

:cl:
fix: Blood Transfer Miracle will no longer cast and then immediately cancel if the caster is in a do_after. The spell will not be put on cooldown, and you will not be charged 50 devotion.
/:cl:

### DIFF
--- a/code/modules/spells/roguetown/acolyte/general.dm
+++ b/code/modules/spells/roguetown/acolyte/general.dm
@@ -397,7 +397,7 @@
 	return FALSE
 
 /obj/effect/proc_holder/spell/invoked/blood_heal
-	name = "Blood transfer Miracle"
+	name = "Blood Transfer Miracle"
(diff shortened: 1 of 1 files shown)


zoktiik:
What would be more ideal is if we added a "channeling" variable to spells that flagged if they require a do_after while you're casting them, and then cast() would just automatically block casting a spell if it's going to require concentration

Xander3359:
> What would be more ideal is if we added a "channeling" variable to spells that flagged if they require a do_after while you're casting them, and then cast() would just automatically block casting a spell if it's going to require concentration

feel free to make that PR

PR #2080, opened 2026-07-08
Revert "Adds Lua scripting"

Xander3359:
Reverts Rotwood-Vale/Ratwood-2.0#2045


noelle-lavenza:
will be readded at a later date when we can coordinate updating the tgs scripts to support it properly

PR #2085, opened 2026-07-08
Have old age retain the +1 to willpower from middle-aged

Gug1234:
## About The Pull Request

- old now gives +1 to willpower, similar to middle aged.

## Testing Evidence
Very very simple one line addition. Compiles.

## Why It's Good For The Game
Struck me as odd old age is kinda its own thing rather than a tier above middle aged with heftier trade offs. This makes the age traits build on top of one another better. 

Adults get nothing, but lose nothing.

The middle aged are slower, but have willpower and fortune.

The old retain their fortune and willpower alongside the intelligence from a long life, but have hefty physical stat nerfs. 

## Changelog

:cl: Yuckuza
balance: Old age now gives +1 willpower, making the total statline: -1 str, -1 per, -2 con, -2 spd, +1 will, +2 int, +1 fortune
/:cl:

### DIFF
--- a/code/modules/client/preferences.dm
+++ b/code/modules/client/preferences.dm
@@ -1825,7 +1825,7 @@ Slots: [job.spawn_positions] [job.round_contrib_points ? "RCP: +[job.round_contr
 							if (AGE_MIDDLEAGED)
 								to_chat(user, "Muscles ache and joints begin to slow as Aeon's grasp begins to settle upon your shoulders. (-1 SPD, +1 WIL)")
 							if (AGE_OLD)
-								to_chat(user, "In a place as lethal as PSYDONIA, the elderly are all but marvels... or beneficiaries of the habitually privileged. (-1 STR, -2 SPE, -1 PER, -2 CON, +2 INT, +1 FOR)")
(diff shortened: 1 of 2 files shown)


zoktiik:
This is making it harder for me to beat grandpa in a marathon pls fix

legalillegaldoc:
wouldn't it make sense that old people are less able to endure longer physical activity?

Gug1234:
> wouldn't it make sense that old people are less able to endure longer physical activity?

They ENDURE long enough to see their hair grey or something. I want to see more old people, and the extra willpower helps offset that horrible con loss.

zeroisthebiggay:
doesn't really make sense from a logic perspective but from a balance perspective it's probably fine

Skyglazer:
This is a buff to Veteran and any mage that wants to choose Old for the bonus spellpoints-- which are already very strong trade-offs for the character being old, as well as the extra intelligence and fortune. This buff is wholly unnecessary. It also doesn't really make much sense for an elderly character to somehow have more stamina and physical endurance than someone in their prime.

Gug1234:
> This is a buff to Veteran and any mage that wants to choose Old for the bonus spellpoints-- which are already very strong trade-offs for the character being old, as well as the extra intelligence and fortune. This buff is wholly unnecessary. It also doesn't really make much sense for an elderly character to somehow have more stamina and physical endurance than someone in their prime.

Medicare often fully insures Viagra, old people have stamina in abundance.

Kroto-Ker:
No, old is the way it is for a reason. It's not a higher tier version of middle-aged. Some character choices should just be negatives, it's okay!

PR #2088, opened 2026-07-09
Absolver revive change

wikimody:
## About The Pull Request
Absolver reviving someone now gives him devitalised debuff (the same as lux extraction, lasts 15 minutes)
Reviving someone when debuffed that way will kill absolver

## Testing Evidence
Tested on a own server, very very anoying to test do to needing 2 creatures with mind (old dreamaker -> guest join trick)

Zrzut ekranu 2026-07-09 014901

Zrzut ekranu 2026-07-09 014914

## Why It's Good For The Game
Absolver being able to revive someone is something that existed as a option for a long time, but was extremely rare as there is next to no situation where its worth trading in the inquisitions only healer for anyone on frontline. From the conversations with other absolver players they don't use it for the same reason.

Now absolver gets to once per fairly long time (15 minutes) bring someone back to life without dying, and still has a option of sacrificing himself to perform a second resurrection.

This also means that finally absolver and stigmata distinction is meaningful, as before the only meaningful thing they lacked compared to absolver is his slightly better self healing. Now his ability to resurrect with they lack will be used.

Apparently this is idea already implemented on AP, but since this is second hand information i cant guarantee that's true.

## Changelog

:cl:
add: Absolver now can absolve death at a cost of getting the lux drain debuff instead of dying, doing so while drained will kill him like before.
/:cl:

### DIFF
--- a/code/modules/spells/roguetown/oldgod.dm
+++ b/code/modules/spells/roguetown/oldgod.dm
@@ -352,9 +352,17 @@
 			return FALSE
 		// Create visual effects
 		H.apply_status_effect(/datum/status_effect/buff/psyvived)
-		// Kill the caster
-		user.say("MY LYFE FOR YOURS! LYVE, AS DOES HE!", forced = TRUE)
-		user.death()
+		if(user.has_status_effect(/datum/status_effect/debuff/devitalised))
+			// Kill the caster
+			user.say("MY LYFE FOR YOURS! LYVE, AS DOES HE!", forced = TRUE)
+			user.death()
+		else
+			// Once per 15 minutes you get to not die for it
+			user.say("MY LUX FOR YOURS! LYVE, AS DOES HE!", forced = TRUE)
+			user.apply_status_effect(/datum/status_effect/debuff/devitalised)
+			to_chat(user, span_warning("You reach out and restore [H]'s Lux, at the cost of your own! If you dont let it recover, you will not live though next attempt."))
+			user.Knockdown(5 SECONDS)
+			user.Stun(5 SECONDS)
 		// Revive the target
 		H.revive(full_heal = TRUE, admin_revive = FALSE)
 		H.adjustOxyLoss(-H.getOxyLoss())
--- a/code/modules/spells/roguetown/oldgod.dm
+++ b/code/modules/spells/roguetown/oldgod.dm
@@ -361,6 +361,7 @@
 			user.say("MY LUX FOR YOURS! LYVE, AS DOES HE!", forced = TRUE)
 			user.apply_status_effect(/datum/status_effect/debuff/devitalised)
 			H.apply_status_effect(/datum/status_effect/debuff/revived)
+			H.apply_status_effect(/datum/status_effect/debuff/devitalised)
 			to_chat(user, span_warning("You reach out and restore [H]'s Lux at the cost of your own! If you don't first let it recover, you will not lyve though another."))
 			user.Knockdown(5 SECONDS)
 			user.Stun(5 SECONDS)
--- a/code/modules/spells/roguetown/oldgod.dm
+++ b/code/modules/spells/roguetown/oldgod.dm
@@ -360,6 +360,7 @@
 			// Once per 15 minutes you get to not die for it
 			user.say("MY LUX FOR YOURS! LYVE, AS DOES HE!", forced = TRUE)
 			user.apply_status_effect(/datum/status_effect/debuff/devitalised)
+			H.apply_status_effect(/datum/status_effect/debuff/revived)
 			to_chat(user, span_warning("You reach out and restore [H]'s Lux at the cost of your own! If you don't first let it recover, you will not lyve though another."))
 			user.Knockdown(5 SECONDS)
 			user.Stun(5 SECONDS)
--- a/code/modules/spells/roguetown/oldgod.dm
+++ b/code/modules/spells/roguetown/oldgod.dm
@@ -360,7 +360,7 @@
 			// Once per 15 minutes you get to not die for it
 			user.say("MY LUX FOR YOURS! LYVE, AS DOES HE!", forced = TRUE)
 			user.apply_status_effect(/datum/status_effect/debuff/devitalised)
-			to_chat(user, span_warning("You reach out and restore [H]'s Lux, at the cost of your own! If you dont let it recover, you will not live though next attempt."))
+			to_chat(user, span_warning("You reach out and restore [H]'s Lux at the cost of your own! If you don't first let it recover, you will not lyve though another."))
 			user.Knockdown(5 SECONDS)
 			user.Stun(5 SECONDS)
 		// Revive the target
--- a/code/modules/spells/roguetown/oldgod.dm
+++ b/code/modules/spells/roguetown/oldgod.dm
@@ -352,16 +352,15 @@
 			return FALSE
 		// Create visual effects
 		H.apply_status_effect(/datum/status_effect/buff/psyvived)
-		if(user.has_status_effect(/datum/status_effect/debuff/devitalised))
+		if(user.has_status_effect(/datum/status_effect/debuff/absolver))
 			// Kill the caster
 			user.say("MY LYFE FOR YOURS! LYVE, AS DOES HE!", forced = TRUE)
 			user.death()
 		else
 			// Once per 15 minutes you get to not die for it
 			user.say("MY LUX FOR YOURS! LYVE, AS DOES HE!", forced = TRUE)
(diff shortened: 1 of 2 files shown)


code/modules/spells/roguetown/oldgod.dm
```diff
@@ -352,9 +352,17 @@
...
-		user.death()
+		if(user.has_status_effect(/datum/status_effect/debuff/devitalised))
+			// Kill the caster
+			user.say("MY LYFE FOR YOURS! LYVE, AS DOES HE!", forced = TRUE)
+			user.death()
+		else
+			// Once per 15 minutes you get to not die for it
+			user.say("MY LUX FOR YOURS! LYVE, AS DOES HE!", forced = TRUE)
+			user.apply_status_effect(/datum/status_effect/debuff/devitalised)
+			to_chat(user, span_warning("You reach out and restore [H]'s Lux, at the cost of your own! If you dont let it recover, you will not live though next attempt."))
```
zeroisthebiggay:
```suggestion
			to_chat(user, span_warning("You reach out and restore [H]'s Lux at the cost of your own! If you don't first let it recover, you will not lyve though another."))
```
`Lux` isn't really supposed to be capitalized but it's capitalized elsewhere in the file so I guess

zeroisthebiggay:
since it doesn't already you should probably update this to give devitalized to the person reviving
originally it was a tradeoff for the fact that you'd kill the person who tried reviving you but now it's free just on a cooldown
I don't really want to see absolvers loitering around the church to use their better magical aheal fix everything revival and deprive the churchies of their surgery time

it might even warrant an even stronger debuff for the person revived since you are quite literally reviving them from whatever might've afflicted them instantly and with no downsides as of now

Ephemeralis:
I recommend looking at (and possibly porting) https://github.com/Scarlet-Reach/Scarlet-Reach/pull/1327 instead, since it does functionally the same thing but with much better user-facing explanation.

wikimody:
I added both normal resurection penalty and devitalised penalty to the "free" resurrection, i left the self sacrificial one free of those. I will now look over the scarlets version of this

wikimody:
> I recommend looking at (and possibly porting) [Scarlet-Reach/Scarlet-Reach#1327](https://github.com/Scarlet-Reach/Scarlet-Reach/pull/1327) instead, since it does functionally the same thing but with much better user-facing explanation.

I wouldn't say the user facing explanation is much better, the different debuff is interesting but i think thematically connecting it to already existing lux lack debuff works better. 
This connection also lets things like lux extraction surgery or the necromancers lux ripping meaningfully weaken absolver by denying him his revive

The-Sun-In-Splendour:
Having the resurrection be very very rare to actually be worth it is a good thing imo. Absolver is already the strongest healer in the entire game, we don't need to make them better.

Skyglazer:
Absolver does not need this whatsoever, they are already an insanely strong healer. The devitalized debuff will effectively mean almost nothing to them, so it's just a free revive every 15 minutes.

wikimody:
> Absolver does not need this whatsoever, they are already an insanely strong healer. The devitalized debuff will effectively mean almost nothing to them, so it's just a free revive every 15 minutes.

The debuff for person revived is worse then the others now, staking normal revival debuff and the lux surgery one

zeroisthebiggay:
maybe make it give the absolver a custom debuff that giganukes their con and even strength (with maybe an aura so you can tell when an absolver is super grabbable and weak)
this makes them grabbable and killable in fights and makes absolving bleeds a much riskier ask

wikimody:
Removed the lux surgery weakness from revive (both on absolver and on target), added new debuff with removes 5 con, 3 will and 3 strength from absolver, with the same 15 minute duration as lux debuff. 
I lament the lost conection to lux surgery, but it turns out that the function checking for status effects goes by ID not subtype, and i dont want to make a maintanance nightmare if statment, so i will leave it as separate thing

UntoldTactics:
The church can use fucking _bread_ to revive people.

I think this is fine, imo. The absolver is a ridiculously strong healer, yes, but they're also capped by the fact that the majority of its players are kind of silly and weep someone who's unconscious on the floor and basically kill themselves. They're rather slow and most people aren't cognizant enough to stay around the guy that can tank 3x their damage and heal them of said damage. Inevitably, Inquisition overextends without them and you can get kills that way, or you can intentionally lure people away so you can then target the absolver, etc.

The Inquisition in many ways are balanced around their absurd healer, as individual roles all but the Inquisitor fall behind compared to even the average MAA or city guard with their town buffs. Underwhelming roles that in a team of five manage to hold their own.

zeroisthebiggay:
> The church can use fucking _bread_ to revive people.

the difference between that and this is that churchies have to actually fix the body and be near a cross, making it much more convenient to take corpses aaaall the way back to the church medbay
the absolver's revive is a full on aheal
it just fixes you and all your problems

so it should have at least a form of non neglgible downside to compensate for its power over other revives

the con nuke is a good step forwards but it might need to be increased further since it's not uncommon for absolvers to straight up have twenty con

UntoldTactics:
> > The church can use fucking _bread_ to revive people.
> 
> the difference between that and this is that churchies have to actually fix the body and be near a cross, making it much more convenient to take corpses aaaall the way back to the church medbay the absolver's revive is a full on aheal it just fixes you and all your problems
> 
> so it should have at least a form of non neglgible downside to compensate for its power over other revives
> 
> the con nuke is a good step forwards but it might need to be increased further since it's not uncommon for absolvers to straight up have twenty con

Could do a large chunk of damage/force the absolver to absolve all the non-lethal damage they can from them during the revival? Might be annoying to code however and simply modifying the stat debuff may end up the better choice.

wikimody:
> the con nuke is a good step forwards but it might need to be increased further since it's not uncommon for absolvers to straight up have twenty con

First, the cross limitation can be straight up ignored if you got 2 stone with you. They are very cheap to make

Second, yes its not uncomon to see absolver with 20 con, but i dont want to excesivly punish anyone who decided to not con max as absolver. I think the con being nuked to 15, with is less then most actual combat clases get with the same enduring the 20 con absolver propably has is enough of a weekness, especialy since for any absolver without con statpack they will be set to 10-12, low enough that their method of healing nessesiting taking on multiple victim worth of damage becomes fairly dangerus to them

The-Sun-In-Splendour:
alternatively we can just close this pr because i still believe that the absolver is already the strongest healer in the game and he does not need to be made better.
try doing literally anything as an antagonist with an absolver on the screen.

Quink-Fix:
based. give us our delicious ABSOLVE revive again...

AllCorruptingWyrm:
Absolver is the strongest healer in the game so i'll be blocking this.

DogExtrordinaire:
Scarlet reach had this and it was pretty balanced and fair. THOUGH they had the extra condition that the absolver COULD revive again but would just die and stay dead(I think). This is probably better.

zeroisthebiggay:
> Second, yes its not uncomon to see absolver with 20 con, but i dont want to excesivly punish anyone who decided to not con max as absolver. I think the con being nuked to 15, with is less then most actual combat clases get

con softcaps at fifteen

> enduring the 20 con absolver propably has is enough of a weekness, especialy since for any absolver without con statpack they will be set to 10-12, low enough that their method of healing nessesiting taking on multiple victim worth of damage becomes fairly dangerus to them

`STATKEY_CON = 7,`
+1 CON from good meal
+1 CON from blessing yourself

you now have nineteen CON without even taking a CON-affecting statpack
if you're a wildkin or any race with plus CON, you now have twenty CON

enduring is larp

wikimody:
-I will be away from home for a week, so there will be no changes made in that time-

I came back

Ephemeralis:
Scarlet Reach absolvers also didn't perform full aheals via ABSOLVE and required you to WEEP the corpse first which usually made you immediately and terribly vulnerable, so people tended to kill bad absolvers who did this at inopportune times.

For some reason, you've (whoever is handling Ratwood, royally) ported only *half* of the very deliberate set of changes I made on Scarlet Reach to make the changes listed in the current PR make sense. ABSOLVE is still a full aheal here which is pants-on-head stupid. I recommend addressing that.

legalillegaldoc:
soul removal

Quink-Fix:
> I recommend looking at (and possibly porting) [Scarlet-Reach/Scarlet-Reach#1327](https://github.com/Scarlet-Reach/Scarlet-Reach/pull/1327) instead, since it does functionally the same thing but with much better user-facing explanation.

Can we genuinely get this version ported? I'd do unspeakable things for this.

PR #1742, opened 2026-05-27
Toggleable Genitals + New Genitals + Fixes Mirror Magic Swapping

ophaq:
## About The Pull Request
- Fixes mirror magic not updating sexcon when swapping to a new pintle. This includes the knotting option, especially with characters who did not have pintles before joining the round.
- Adds in two new pintles. They’re sheathless equine ones, with both the knotted and non-knotted variants.
- Adds a toggle to the sexcon menu to hide genitals. This means that your vagina, balls, and pintle will be like they aren’t even there once you toggle it. Primarily, this is used for taur characters or for people who for some reason don’t want their genitals shown when naked.

## Testing Evidence
sheathlesshonsetest
sxconmenutest

## Why It's Good For The Game
First, some of this was part of a bounty. One that someone had picked up months ago but there was no progress on it, as far as I'm aware. But since I was already doing something in the sexcon menu, I figured I'd fix a few other things that bothered me and add in a feature another player has been wanting for a while. After all, not everyone wants a sheath with their equine pintle I guess.

Anyway, this PR is really good for centaurs since it gives them an option to hide the goods without having to just toggle 'unaroused' in the menu. Also good for players that just wanna hide the bits down there for whatever reason.

Mirror magic pintle swapping has been kinda buggy for a while and it bothered me. Basically if a player didn't already start with a pintle in-game when they spawned in, the game would never update the sexcon menu to give them the verbs and knotting options once they used their pintle. It also made pintles occasionally look like it was vertically flat on their sprite instead of properly updating to be horizontal.

## Changelog
:cl:
add: Centaurs and people who are interested rejoice! Sexcon menu now lets you hide your genitals.
add: Two new pintles. Sheathless equine ones of both variants.
fix: Fixes mirror magic genital swapping. It now updates the sexcon menu so you actually use the verbs of the pintle / etc you switched too, plus you can knot. It should also no longer make your pintle look flat and update properly.
/:cl:

PR #2093, opened 2026-07-09
Freeuse Sexcon Toggle

EvenInDeathIStillServe:
## About The Pull Request

adds a "freeuse" sexcon toggle that makes it so that sex actions targeting you will bypass any positioning or clothing checks.

## Testing Evidence

image

image

image

## Why It's Good For The Game

sometimes u wanna stuff your homie's face into your armpit without taking everything off yk
it's purely consensual, so it's really just for QoL so you don't have to fuck around with grabs in a scene

## Changelog

:cl:
add: Added a "freeuse" sexcon toggle that makes it so that sex actions targeting you will bypass any positioning or clothing checks.
/:cl:

### DIFF
--- a/code/datums/sexcon/sexcon.dm
+++ b/code/datums/sexcon/sexcon.dm
@@ -77,6 +77,8 @@
 	var/mob/living/carbon/knotted_recipient = null // whom took the knot
 	/// Allow crotch to be exposed and bypass clothes check
 	var/bottom_exposed = FALSE
+	/// Bypasses positioning and exposure checks entirely
+	var/freeuse = FALSE
 	// Moved here from proc/get_generic_force_adjective to reduce list initialization/destruction
 	var/static/list/stealth_force_adjectives 	= list("subtly", "sneakily", "covertly", "stealthily", "quietly")
 	var/static/list/low_force_adjectives 		= list("gently", "carefully", "tenderly", "gingerly", "delicately", "lazily")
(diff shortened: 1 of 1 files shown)


RasbeRissa:
I think the only way this could be better is if it saved it's status between rounds. Maybe putting it into the toggles menu we have now? That way we don't need to open the ERP menu to fix it. I know I'll end up forgetting and by the time I fix it someone will have already just grabbed instead lol

ophaq:
I like this idea but it might also conflict with my old PR https://github.com/Rotwood-Vale/Ratwood-2.0/pull/1742/ that still isn't TM'd or merged yet to fix other issues. An option in toggles would be better... maybe.

Aracino:
Also would be nice if this let you treat yourself naked when interacting with others. Clothed zuri and grapple beast upscale

EvenInDeathIStillServe:
> Also would be nice if this let you treat yourself naked when interacting with others. Clothed zuri and grapple beast upscale

It does this.

Xander3359:
Why not just make this a pref?

EvenInDeathIStillServe:
> Why not just make this a pref?

I will make it a pref along with a toggle on the menu. The reason I made it part of the menu is because it's an "in the moment" thing and I would rather not people have to navigate the toggles menu in order to get to it, but I do see the benefit of having it there, too.

PR #2094, opened 2026-07-10
fix revenants being able to ignore head wounds if their head is removed

Arturlang:
## About The Pull Request
adds a system to make wounds persist on removed wounds because revenants, as a trait that is applied on a limb, now applied to revenant heads only.
also fixes several organs healing all wounds when inserted because wtf?

## Testing Evidence
gave the revenant skullcrack, tried to move around, decapped it, tried to move around, couldn't in both cases, re-attached the head, still paralyzed, ahealed, worked properly.

## Why It's Good For The Game
people asked me to do this, doh

## Changelog

:cl:
fix: inserting organs into the head no longer heals all wounds
fix: revenants can no longer run around without downsides from their head wounds if they remove their head
/:cl:


RasbeRissa:
If my suspicions are correct, this is, in fact, bugged badly. Revenants are NOT able to be fixed from a skull crack through in-game measures. Surgery, Pestra Ritual, nothing works to unparalyze. Even ahealing them doesn't fix their name. Should be unmerged until it can be debugged.

Yeetus223:
feels unnecessary, they already get a ton of downsides from just having their head off and have a ton of bugs with having their head off anyways.

Arturlang:
> If my suspicions are correct, this is, in fact, bugged badly. Revenants are NOT able to be fixed from a skull crack through in-game measures. Surgery, Pestra Ritual, nothing works to unparalyze. Even ahealing them doesn't fix their name. Should be unmerged until it can be debugged.

should be fixed

PR #2099, opened 2026-07-10
Stone 2 clay

DorfGuy1:
Inverse of another recipe

You can transform clay to stone. Why not the inverse. I think this will be useful.
:cl:
Alchemy can make clay from stone
/:cl:


stealthkibbler2-byte:
lol2clay

Xander3359:
Allows you to grind alchemy xp quicker.
But ngl you can just grind it super easily anyways so it's fine.

DorfGuy1:
Strangely enough I think that I screwed up and now it gives only one clay?

PR #2112, opened 2026-07-13
Resurrect oxyloss removal before the revive check

PPooch:
## About The Pull Request

#497 moved the "	target.adjustOxyLoss(-target.getOxyLoss()) //Ye Olde CPR" line after the "if(!target.revive(full_heal = FALSE))" check for both lux surgery and leechtick surgery: which leads to an edge case where a corpse may too suffocated to actually be resurrected by these means. This PR puts the Oxyloss heal before the revive check, which aligns with the way the other forms of resurrect handle it: namely the revive chair and the Anastasis resurrects.

Here's a picture of me running into the issue in a round.
image

## Testing Evidence
Before, recreated the bug.
[old_oxyloss720.webm](https://github.com/user-attachments/assets/bb3419bf-5f7d-40ee-b7d7-d1fcd2094ea3)

After.
[oxyloss.webm](https://github.com/user-attachments/assets/3f0639c0-98a0-4e25-a8a6-75322afe7dd1)

## Why It's Good For The Game
Barber surgeons 

## Changelog

:cl:
fix: fixed edge case of a corpse being too suffocated to be resurrected by lux surgery
/:cl:

### DIFF
--- a/code/modules/surgery/surgeries_hearth/revival.dm
+++ b/code/modules/surgery/surgeries_hearth/revival.dm
@@ -65,6 +65,7 @@
 				"[user] tries to infuse [target] with lux, but it refuses to take.")
 			target.visible_message(span_danger("[target]'s body convulses violently, rejecting the light!"), span_warning("Something is terribly wrong..."))
 			return FALSE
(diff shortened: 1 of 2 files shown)


zoktiik:
Very reasonable fix, thank you!

PR #2113, opened 2026-07-13
Caps bow and crossbow scaling up to 15 perception

Xander3359:
## About The Pull Request
Caps bow and crossbow scaling up to 15 perception

## Why It's Good For The Game
Balance jakking to make bows not do double damage for minmaxxers

## Changelog

:cl:
balance: bow and crossbow damage now stop scaling above 15 perception
/:cl:

### DIFF
--- a/code/game/objects/items/rogueweapons/ranged/bows.dm
+++ b/code/game/objects/items/rogueweapons/ranged/bows.dm
@@ -233,9 +233,12 @@
 			BB.damage -= (BB.damage * (user.client.chargedprog / 100))
 			BB.embedchance /= 2
 			BB.accuracy -= 15
-		else
-			BB.damage = BB.damage
-		BB.damage *= damfactor * (user.STAPER > 10 ? user.STAPER / 10 : 1)
+		var/perception_modifier = user.STAPER
+		if(perception_modifier > 15) // Bow damage hardcapped at 15 PER
+			perception_modifier = 15
+		if(perception_modifier < 10)
+			perception_modifier = 10
+		BB.damage *= damfactor * (perception_modifier / 10)
 	return ..()
 
 /obj/item/gun/ballistic/revolver/grenadelauncher/bow/update_icon()
--- a/code/game/objects/items/rogueweapons/ranged/crossbows.dm
+++ b/code/game/objects/items/rogueweapons/ranged/crossbows.dm
@@ -164,8 +164,12 @@
 		BB.damage *= damfactor
 
 		if(damage_from_perception)
-			if(user.STAPER > 10)
-				BB.damage *= (user.STAPER / 10)
+			var/perception_modifier = user.STAPER
+			if(perception_modifier > 15) // crossbow damage hardcapped at 15 PER
+				perception_modifier = 15
+			else if(perception_modifier < 10)
+				perception_modifier = 10
+			BB.damage *= (perception_modifier / 10)
 
 	cocked = FALSE
 	..()


kimesik:
Normal crossbows don't scale from PER either way, so this affects slurbows rather than normal crossbows

Yeetus223:
this seems like it's just nerfing people that make ranged builds more than people that make str builds, even str above 15 at least gives diminishing returns

CloverIsLucky:
Imo crossbows should be the only PER based weapon and bows should require STR damage scaling because you tense a bow string to shoot an arrow.

NotGoldenglow:
I've a few suggestions. Mainly concerning force scaling.

Mostly, bring it in line to STR scaling. 10% per point up to 14, 3% onwards. 
Uncap slurbows. They're still dogshit at 20 PER anyway.

Longbows have no cap, but the scaling should be brought down to 5% per point. Instead, add STR scaling for another 5%. To achieve the same level as before, you'll need 15 PER and STR (to equate 20 PER), while giving it the niche it deserves as a heavier draw bow. 

This means longbows will not be the de-facto go-to for everyone, and will encourage people to know their builds and pick between Recurve vs. Longbow properly.

Plum-but-gayer:
THANK YOU HOLY CRAP.
PER is balanced at 1/2 value of STR despite being better at increasing damage for it's respective weapon - a weapon that can do a type of damage to the head that procs skullcracks/removes eyes and that is extremely difficult to negate with armor from range and across Z levels. Regardless of what the solution is, something needs to be done about it.

Kneesox44:
I think this is a good change, but at the very least should also include a "Archery Unbound" which allows the cap to be broken in the same vein as "Strength Unbound" for a few specific characters, like the Master Warden.

Ephemeralis:
Longbow broadheads are still 72 force at full charge with 15 PER. Most people were rolling 18 PER probably (assuming +3 PER statpack or warden/woods bonuses), and were doing 86~ force at full charge previously.

All you've done with this nerf is drop 14 force off a shot that was already going to probably hit crit thresholds for the major puncture wound types anyway.

This functionally does *nothing*. Even a cursory understanding of RT damage math would've illustrated this.

I'm not advocating for you to increase it or decrease it but maintainer adjustments need to be more than just blind edits that hope for the best.

To avoid being annoying and unproductive:
- Consider dropping the damfactor increase off longbows or lowering it from 1.2x to 1.15 or 1.10.
- Drop broadhead damage down from from 40 to 35, or alternatively, stop making every single archery class in the entire game spawn with a quiver full of the highest force arrows available.
- Lower iron broadhead AP down to 10 from 20 to further make armor more effective against them.
- Leave slurbows alone entirely. They're already a 40% damage reduction over an ordinary crossbow and 50% of the AP - at 20 PER they just *barely* get back to standard bolt damage and still aren't anywhere near as threatening because their AP is literally halved.

kimesik:
> I think this is a good change, but at the very least should also include a "Archery Unbound" which allows the cap to be broken in the same vein as "Strength Unbound" for a few specific characters, like the Master Warden.

I don't think there's any non-major-antag that inherently gets Strength Unbound, so I don't think any non-major-antag should get inherent Archery Unbound either.

n3tis:
> > I think this is a good change, but at the very least should also include a "Archery Unbound" which allows the cap to be broken in the same vein as "Strength Unbound" for a few specific characters, like the Master Warden.
> 
> I don't think there's any non-major-antag that inherently gets Strength Unbound, so I don't think any non-major-antag should get inherent Archery Unbound either.

Archery and ranged builds are fucked anyways if you're not running Toxophilite to gain higher skill ceilings in them, cause otherwise you're limited to just expert even if you're playing a retinue/wretch role that SHOULD have some advantages (they don't), lol
Wardens, who are supposedly vanguard "veterans" only get ranged skills up to Expert, compared to the MAA that gets master in all three different skills. And not to mention, the slings having a separate skill level ceiling that allows you to gain legendary level.
This is just more CBT PRs to against the `[insert]` players that as stated by @Ephemeralis hardly does anything. Maybe people should propose a new change to farming instead of "**sex, combat rebalance, more sex and maybe we'll add another CBT feature that will have to process 150+ people's character states worldwide**"

wikimody:
> I don't think there's any non-major-antag that inherently gets Strength Unbound, so I don't think any non-major-antag should get inherent Archery Unbound either.

All garagorites miraclists? The base T0 miracle gives that

wikimody:
Also i agree that per should have the strength lower scaling, with 2% instead of 10% per per above 15. Why shouldn't that mechanic be consistent between stats if you make caps consistent

PR #2132, opened 2026-07-15
Make strong intent actually work with Unarmed damage

ariaworld:
## About The Pull Request
Strong intent is described as if attacking with "+1 STR". This gives a bonus to damage to items (weapons), but looking at the code I noticed this never actually applied to bare fists, this fixes that.
Code for weapons:
a

## Testing Evidence
It's a simple change copy-pasted from weapons' strength intent damage buff

## Why It's Good For The Game
More consistency, thing does what thing says it do, also I'm pretty sure azure peak fixed this as well

## Changelog

:cl:
fix: Strong intent now correctly gives a bonus to unarmed fists as well.
/:cl:


gr4ts:
You mean I have been using strong intent up till now and it's done _absolutely nothing to help?_

PR #2135, opened 2026-07-16
fix some infinite loop errors when launching the game

Arturlang:
## About The Pull Request
fix some infinite loop errors when launching the game

## Testing Evidence
builds, doesnt seem to break stuff

## Changelog

:cl:
fix: some unlimited loop runtimes
/:cl:

### DIFF
--- a/code/game/area/areas.dm
+++ b/code/game/area/areas.dm
@@ -503,6 +503,7 @@ GLOBAL_LIST_EMPTY(teleportlocs)
  *
  */
 /area/proc/update_areasize()
+	set background = 1
 	if(outdoors)
 		return FALSE
 	areasize = 0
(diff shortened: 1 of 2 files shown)


code/game/area/areas.dm
```diff
@@ -503,6 +503,7 @@ GLOBAL_LIST_EMPTY(teleportlocs)
  *
  */
 /area/proc/update_areasize()
+	set background = 1
```
noelle-lavenza: don't think this is actually a proper solution, we should likely port the area contents subsystem to fix this correctly. but that's an ordeal and a half (i've been doing it on cm for months)

noelle-lavenza (changes requested):

PR #2140, opened 2026-07-16
Added a pair of brackets to damage calculation, or, how I learned why Gnolls and Werewolves die horribly

SneakySlinky:
## About The Pull Request

So, presently, damage is calculated by `damage = max(damage - blocked + armor, 0` ; The `armor` here is the interesting thing, cause in the case of gnolls and werewolves, this value is `30`. Because of how the math crunches out, this means WWs and Gnolls are, in every single instance of damage taken, having an additional 30 added on to hits they take. This explains a few things, but most notably why people feel like Gnolls and WWs are made of glass. This adds a second set of brackets, making it `damage = max(damage - (blocked + armor), 0)`, properly accounting for the species armor present on things such as gnolls and werewolves in particular, who were feeling particularly frail under fire.

## Testing Evidence

dreamseeker_DnqNZm4bJO
dreamseeker_ZopD3MTSUS
dreamseeker_0IGeAJBn2i

First two pics: Swung at self with claw on CUT, 13 str. 49 brute each.
Last pic, after adding brackets: Swung at self with same setup, didn't break armor or do damage to self at all, had to break armor entirely to meaningfully hurt self. The gnoll no longer explodes.

## Why It's Good For The Game

This oddity of the armor computation most notably arose when hitting gnolls with a lot of small attacks that you think would not be nearly as effective, but for, whatever reason, were BLOWING GNOLLS THE FUCK UP, and werewolves by extension.

Fire in particular was quite good at this, as it would hit multiple zones at once with damage that was then being vastly increased by this order of operations.

## Changelog

:cl:
balance: buff to ww and gnoll armor (it was broken)
code: added additional set of brackets to line of code in species.dm, Line 2036
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species.dm
+++ b/code/modules/mob/living/carbon/human/species.dm
@@ -2033,7 +2033,7 @@ GLOBAL_LIST_EMPTY(roundstart_races)
 /datum/species/proc/apply_damage(damage, damagetype = BRUTE, def_zone = null, blocked, mob/living/carbon/human/H, forced = FALSE, spread_damage = FALSE)
 	SEND_SIGNAL(H, COMSIG_MOB_APPLY_DAMGE, damage, damagetype, def_zone)
 	var/hit_percent = 1
-	damage = max(damage-blocked+armor,0)
+	damage = max(damage-(blocked+armor),0)
 //	var/hit_percent =  (100-(blocked+armor))/100
 	hit_percent = (hit_percent * (100-H.physiology.damage_resistance))/100
 	var/atom/movable/screen/zone_sel/zone_sel


RasbeRissa:
Seems like this might be doing the opposite now. Gnolls seem to be taking 30% less damage from all sources instead of the proper 0'd amount. Three retinue wailing on a single gnoll and the gnoll ended the fight "lightly bruised" after killing all three of them.

Edited to Add: I have no idea why this wasn't test merged before full merge to ensure it worked as intended. Why is this a full merge?

AlucardHellsingChapel:
the pr needs to be unmerged instantly the gnolls are taking even less damage than they should making them almost unkillable

DorfGuy1:
is this -30 damage taken unreducable in anyway? because 30 is how hard a good weapon hits in the hands of a 10 str char, so with mult and str+ you would get like 10 damage off.

AllCorruptingWyrm:
Test merge your fucking balance PRs.

(Niko is not at fault).

SneakySlinky:
You can just adjust the species armor define to be WHATEVER NUMBER YOU WANT to balance it now, because now it properly works.

Kneesox44:
I wonder if this is why Gnolls would take 300+ damage from fireballs.

PR #2151, opened 2026-07-17
I changed a pair of vars to 0, or, how I learned sometimes you just have to do the balancing yourself.

SneakySlinky:
## About The Pull Request

~~Instead of doing the brackets change as per #2140, given that nothing else besides WW and Gnoll have a set value that isn't 0...  We'll just... Reduce gnoll and WW's armor var to 0, so they DON'T take a bonus 30 damage from NOTHING in particular.~~

I lied, we're just going to change the computation to properly treat species.dms armor var as a % reduction, since it seems to be the intended use of it (brutemod exists to be flat reductions, instead).

## Testing Evidence

dreamseeker_9q1MUF8SRa
dreamseeker_bBmtr4S37T

It works, sire. Same setup as previous PR, now the damage is reduced by 30%.

## Why It's Good For The Game

~~Regardless of whether or not the math sucks, gnolls and WWs should not be taking a flat 30 bonus damage to every singular instance of damage against them. The computation is broken again (due to being reverted), but a 0 is a 0.~~

Whether or not 30% winds up being too much, the fact is the computation was *broken*, so changing it to this at least means if for whatever reason, you want to set a species to have a % reduction to damage (or not at all), you can do so.
Adjust as necessary.

## Changelog

:cl:
balance: Gnolls and WWs now take 30% reduced damage from hits, rather than +30 damage from all hits.
/:cl:

### DIFF
--- a/code/game/objects/items/rogueweapons/mmb/steal.dm
+++ b/code/game/objects/items/rogueweapons/mmb/steal.dm
@@ -15,7 +15,7 @@
 	if(!isnum(range_add))
 		range_add = 0
 	var/steal_radius = 1 + range_add
-	var/list/stealablezones = list("chest", "neck", "groin", "r_hand", "l_hand")
+	var/list/stealablezones = list("chest", "neck", "groin", "r_hand", "l_hand", "r_leg", "l_leg")
 	// Pickpocketting checks
 	if(get_dist(thief, victim) > steal_radius)
 		to_chat(thief, span_warning("[victim] is too far away."))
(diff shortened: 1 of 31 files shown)


SneakySlinky:
I will reiterate this in comment, if it's too strong just reduce the numbers! The previous formula was literally just making the value INCREASE damage taken!!

AllCorruptingWyrm:
yeah this works, thank you niko.

kimesik:
Why is this still unmerged?

PR #2161, opened 2026-07-17
Make neighboverlay_list a lazylist

noelle-lavenza:
Untested, will test soon. This will save a maximum of 4MB per Z-level, likely less in practice, but it's a good start.


noelle-lavenza:
death to all perma-TMs

PR #2163, opened 2026-07-18
removes smash from whetstones: on the right branch

zeroisthebiggay:
## About The Pull Request

PR title

shouldn't affect intended functionality of whetstones

## Testing Evidence

image

## Why It's Good For The Game

I heard that people were using the smash on whetstones to break grabs; I think using smash to break grabs is already kind of an unintended mechanic, but it's counterable by grabbing arms and whatnot

... but being able to do it with a fucking whetstone is a little obscene, at least go and find a rock with smash intent

## Changelog

:cl:
balance: removes smash from whetstones
/:cl:

### DIFF
--- a/code/game/objects/items/rogueitems/natural/stones.dm
+++ b/code/game/objects/items/rogueitems/natural/stones.dm
@@ -175,7 +175,7 @@ GLOBAL_LIST_INIT(stone_personality_descs, list(
 	w_class = WEIGHT_CLASS_TINY
 	experimental_inhand = FALSE
 	mill_result = /obj/item/reagent_containers/powder/mineral
-	possible_item_intents = list(/datum/intent/hit, /datum/intent/mace/smash/wood, /datum/intent/dagger/cut)
+	possible_item_intents = list(/datum/intent/hit, /datum/intent/dagger/cut)
 	sharpening_factor = 21
 	spark_chance = 80
 


ariaworld:
Wow, didn't expect this information to spread this fast
Yes, as someone who fights using grapples, people are starting to cheese out of them by utilizing smash, it's quite annoying, it doesn't feel right
Maybe it would be better to make it impossible to smash knockback someone grabbing a person, and letting whetstones keep smash

UntoldTactics:
You can technically just grab someones arms to prevent them actually using any smash tool. Ofc they can just then bait you, but thems the tradeoffs.

zeroisthebiggay:
> Maybe it would be better to make it impossible to smash knockback someone grabbing a person, and letting whetstones keep smash

Whetstones don't lose anything by losing this intent other than the bizarre ability to inflict knockback.

In its current state, I think letting smash break grabs is fine considering the balance of grabs basically just being a win button if you pass the statrolls. Maybe if grabbing was a little bit less obnoxious I'd patch out smash intent breaking them, but until then a pocket cudgel is basically the only way to quickly break a grab if you aren't building ridiculous amounts of constitution.

HalfAwake691:
> You can technically just grab someones arms to prevent them actually using any smash tool. Ofc they can just then bait you, but thems the tradeoffs.

You can grab their arm then target chest, Because- bait only works on the limb you're TARGETING not grabbing

DogExtrordinaire:
It is an intended mechanic. Shields and blunt weapons are intended to be used for this.

RasbeRissa:
The current state of grabs means that everyone and their mother has high wrestling skill. Larger scale interactions often feature some random towner, acolyte, or other non-traditional 'front liner' racing in to get a passive grab on you so that you get utterly slaughtered.

With wrestling being the single most powercrept thing on the entire server at this time, I am strongly opposed to anything that nerfs or removes a counter to it. Kicking doesn't do much because Swift Intent Grapple. Headbutt doesn't do much because if they're already grappling you, chances are pretty high they've got a better chance to grapple than you do - so they'll auto resist out of your grab. Baiting only works on limbs they're targeting, so you better hope they're going for something that isn't chest AND keep the target on that limb.

If you wanna remove the smash intent emergency escape tool, remove wrestling from every single non-combat class, and drop it to novice or apprentice for 'back line' 'combat' classes like Acolyte.

> at least go and find a rock with smash intent

How is this any different from a whetstone?

zBxnny:
I keep a stone in my belt for this reason, DONT REMOVE IT FROM THE ROCK OR ILL DIE

zeroisthebiggay:
> If you wanna remove the smash intent emergency escape tool, remove wrestling from every single non-combat class, and drop it to novice or apprentice for 'back line' 'combat' classes like Acolyte.

see my other PR, and this PR does not affect using smash to kick back grabbers, it just removes it from whetstones in particular because they're a very common item that people carry and fit in a singular slot in your inventory

> > at least go and find a rock with smash intent
> 
> How is this any different from a whetstone?

a whestone is a useful item that basically anyone with a sword should be carrying
a rock with smash:
A: is something you have to actively find rather than any rock plus a stick
B: annihilates your sharpness if you use it to sharpen your weapon
C: an actual viable weapon if you find one with enough force

I want carrying anti-grab implements like a pocket cudgel to be something you have to think about, rather than just stuffing a whetstone in your belt; ideally I'd make it so you can't use any old rock to completely neuter any and all grabbers, but I don't want to take smash intent from rocks for soul reasons - maybe making it so you need to have a certain amount of force for smash intent to knock people back

PR #2166, opened 2026-07-19
Fixes Hookahs' targeting and adds the ability to make you able to force others to puff the Hookah.

awwliedacoder:
## About The Pull Request

So. I felt this in game. I clicked someone, and thought: "hey, I want to share this smoke without dropping the Hookah: so I clicked X' and then noticed, I got the smoke effect. I crawled through the code noticed that target is not referenced at all, it is only the user. So this pull-requests fixes that and makes it so that you can infact make others smoke the hookah with you.

## Testing Evidence

image
image
image

Works on self others.

## Why It's Good For The Game

Why is this great? well-- now it is a bit more realistic, you can smoke with others: sharing the hookah from your hand, without having to always **drop it** down on the ground. Now you can just-- target someone with the smoke intent and click them so that they get the drug-effects. 

## Changelog

:cl:
add: Added the ability to make others smoke the Hookah's (Portable and Stationary) 
Fix: Fixed targeting of Hookahs. (you could click someone else and it would make **you** smoke the Hookah)
/:cl:


Xander3359 (changes requested):
Not sure what's going on with your PR but it looks like 2 unrelated changes managed to slip on by. Maybe pull upstream and see if that fixes it, or else you'll have to manually revert the changes to the unrelated files.

awwliedacoder:
I will make a new pr about this from another branch to save my soul.

PR #2182, opened 2026-07-21
Old age stat/job restoration and parry buff

RadiantFlash7:
## About The Pull Request
This PR re-enables old age for a few guardsmen roles that had old age available back in RW1.
It also ports the 'veteran' men at arms class rebranded as Retainer. The Retainer has a broader stat spread, though it still adds up to a weighted 7 stat points, like the other classes. 
We have also restored the Int bonus from being old from 2, back to three. AP originally reduced the int bonus to 2, back when feints were king and left people able to be hit for an extended period of time, but never adjusted it back after the parry changes we presently have.
Finally, For the past few days I have been grappling with the fact that if you play any martial role that wasn't a mage, being old was shooting yourself in the foot. Non martial roles got benefits by virtue of having their skill level increased in their specific jobs- but doing this naturally to martial roles presented a significant balance challenge, as well as making it difficult to handle roles that already have high martial skill that can't be increased without causing problems.
Thusly, in an effort to make Old martial characters a bit less suffering, Non-mage old characters get a bonus when it comes to parrying, simply for being old, something I think thematically makes sense. 

## Testing Evidence
image

## Why It's Good For The Game
Previously, Playing old martial characters who aren't living legends like the Veteran was functionally kneecapping yourself. You lost 1 strength, 2 speed, 2 con, 1 perception for two intelligence and one fortune. When applying the standard stat weight, that Speed and strength at worth '2', that ends up being a net six stat loss- something that is exceptionally painful. In the past, that was fairly mitigateable due to how busted feints were, leaving people open for repeated strikes for an extended period. However, with the changes to how feints work, int based characters, which old characters predominantly exclusively are, suffered.

I looked back to RW1 to see how best to improve the situation, and noticed the discrepancy between stat blocks. That alone is an easy fix to restore old folks getting 3 int again, as well as restoring the option for being old in some guard roles as rather soulful, But it still leaves us with the problem that with feinting no where near as good as it used to be (And it should probably never return to how oppressive it was), old martial roles are fairly shafted.
The introduction of the parry buff to old characters is a fairly reasoned change that I think helps reflect an old character having lived as long as they have, while preventing a balance nightmare of actually giving them an extra skill level- something that affects training other roles and affects bypassing the opponents parry.

## Changelog

:cl:
add: Old age buff to parrying
add: Retainer Subclass to men at arms for old age MAA.
add: MAA & Town Guards are once more able to be old
balance: old age once more grants 3 int instead of 2.
/:cl:

### DIFF
--- a/code/modules/jobs/job_types/roguetown/garrison/manorguard/cavalry.dm
+++ b/code/modules/jobs/job_types/roguetown/garrison/manorguard/cavalry.dm
@@ -42,7 +42,7 @@
 	..()
 	neck = /obj/item/clothing/neck/roguetown/gorget
 	gloves = /obj/item/clothing/gloves/roguetown/chain/iron
-
+	beltl = /obj/item/rogueweapon/mace/cudgel
 	H.adjust_blindness(-3)
 	if(H.mind)
 		var/weapons = list("Bardiche","Sword & Shield")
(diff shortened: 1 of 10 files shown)


Goriusrok:
More hags 👍

PR #2185, opened 2026-07-21
Forced pop-up of the Voting-UI & a null-refrence hang on update_temperature_hud

awwliedacoder:
## About The Pull Request

What is this? Well- it is quite simple. This new voting system for whatever reason only had a message and forced the users to be proactive. Surprise, surprise. People are busy once the round ends or is about to start leading to only about maybe, at tops 30% of the people voting. 

So. 
This forces the voting-menu to open as a vote is called, and it also fixes a silly null-ref hangup that can happen from forcing a start of the game.

## Testing Evidence

image

## Why It's Good For The Game

People lack the concentration (me included), to spot out the 'vote now' button as it gets buried underneath 200 ooc-messages leading to VERY, VERY low voting %. It should be without a doubt something **really** simple to fix with this one feature of adding a simple FORCED interaction. 

## Changelog

:cl:
fix: null-ref hangup on update_temperature_hud.
add: added a forced pop-up of the voting ui whenever a vote is called.

/:cl:

### DIFF
--- a/code/controllers/subsystem/vote.dm
+++ b/code/controllers/subsystem/vote.dm
@@ -234,7 +234,7 @@ SUBSYSTEM_DEF(vote)
 	to_chat(world, span_infoplain(vote_font("\n[span_bold(to_display)]\n\
 		Type <b>vote</b> or click <a href='byond://winset?command=vote'>here</a> to place your votes.\n\
 		You have [DisplayTimeText(duration)] to vote.")))
-
+	
 	// And now that it's going, give everyone a voter action
 	for(var/client/new_voter as anything in GLOB.clients)
 		var/datum/action/vote/voting_action = new()
@@ -243,10 +243,13 @@ SUBSYSTEM_DEF(vote)
 
 		new_voter.player_details.player_actions += voting_action
 		generated_actions += voting_action
-
+		
 		if(current_vote.vote_sound)
 			SEND_SOUND(new_voter, sound(current_vote.vote_sound))
 
+		if(SSvote.initialized)
+			SSvote.ui_interact(new_voter.mob)
+			
 	return TRUE
 
 /**
--- a/code/controllers/subsystem/vote.dm
+++ b/code/controllers/subsystem/vote.dm
@@ -247,9 +247,8 @@ SUBSYSTEM_DEF(vote)
 		if(current_vote.vote_sound)
 			SEND_SOUND(new_voter, sound(current_vote.vote_sound))
 
-		if(SSvote.initialized)
-			if(new_voter.prefs.voting_popup)
-				SSvote.ui_interact(new_voter.mob)
+		if(SSvote.initialized && new_voter.prefs.voting_popup)
+			SSvote.ui_interact(new_voter.mob)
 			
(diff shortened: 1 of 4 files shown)


ariaworld:
By the way I noticed the new voting UI doesn't take into account voting power depending on role anymore, it just counts a single vote?

annikaRU:
Could a forced popup be made into a on by default toggle, not all of us want it blatantly in our face and having it as a toggle so those who do not want it can disable it

awwliedacoder:
> Could a forced popup be made into a on by default toggle, not all of us want it blatantly in our face and having it as a toggle so those who do not want it can disable it

I personally think the only way to encourage voting is to toss it in the face of people. Especially when during **PEAK** hours, from 110 connections, only 32 people vote.  That is **LESS** than 30% of the active connections voting at all.



> By the way I noticed the new voting UI doesn't take into account voting power depending on role anymore, it just counts a single vote?

This new voting-system only takes 1 singular vote **from** anyone connected, and everyone has equal voting power.

awwliedacoder:
alright, I lied. I added the toggle. 
it works and compiles.

annikaRU:
> alright, I lied. I added the toggle. it works and compiles.

Thanks much

code/modules/client/preferences.dm
```diff
@@ -119,6 +119,7 @@ GLOBAL_LIST_EMPTY(chosen_names)
 	var/extreme_erp = FALSE
 	var/edging = FALSE
 	/// If a cursed collar can be equipped to them at all
+	var/voting_popup = TRUE
```
Xander3359: This breaks the autodoc for the variable below. The autodoc has to always be the line before the variable.

code/modules/client/preferences_toggles.dm
```diff
@@ -454,6 +458,23 @@
...
+		return
+
+	prefs.voting_popup = !prefs.voting_popup
+	prefs.save_preferences()
+	if(prefs.voting_popup)
+		to_chat(src, "You will see a popup of the voting ui as a vote is called.")
+		return
+	else
+		to_chat(src, "You will no longer see a popup of the voting ui as a vote is called.")
+		return
```
Xander3359: You don't need to call `return` in an `if/else` statement

code/controllers/subsystem/vote.dm
```diff
@@ -243,10 +243,14 @@ SUBSYSTEM_DEF(vote)
...
 		new_voter.player_details.player_actions += voting_action
 		generated_actions += voting_action
-
+		
 		if(current_vote.vote_sound)
 			SEND_SOUND(new_voter, sound(current_vote.vote_sound))
 
+		if(SSvote.initialized)
+			if(new_voter.prefs.voting_popup)
+				SSvote.ui_interact(new_voter.mob)
```
Xander3359:
```suggestion
		if(SSvote.initialized && new_voter.prefs.voting_popup)
			SSvote.ui_interact(new_voter.mob)
```

Xander3359 (changes requested):

PR #2152, opened 2026-07-17
Revert "Makes Pickpocketing Not Utter Bullshit"

CameronWoof:
Reverts Rotwood-Vale/Ratwood-2.0#1924

PR is actually a massive BUFF to pickpocketing, allowing mute knaves to steal from inside containers (including belts, where people keep their keys) without any indication is actually worse than satchel snatching. At least you have a chance in hell of noticing your satchel being stolen, noticing keys being taken from inside a closed container is absolutely impossible.

PR #2191, opened 2026-07-23
Pickpocketing rework

Arturlang:
## About The Pull Request
reworks pickpocketing to let you see what items you pickpocket, % chances on a per-item basis if you're opening inventories & make belts and keys harder to pickpocket out of
also makes pickpocketing now scale with pickpocketer's luck stat to compensate, as you need higher ceiling to avoid being caught whatsoever
also also, the contest calculation factors in victim's stat speed and not just perception
the new pickpocketing UI does require both to stay still, but should show up fast enough to not be too big a issue
see https://arturlang.github.io/pages/pickpocket-rng.html for a RNG chart

closes https://github.com/Rotwood-Vale/Ratwood-2.0/pull/2152

## Testing Evidence

https://github.com/user-attachments/assets/ff6321f7-9a48-4583-bef3-b23d7b626e7e

https://github.com/user-attachments/assets/37311a5e-be64-4ff7-a99f-38dbc5009dc0

## Why It's Good For The Game
more interaction both ways, probably.

## Changelog

:cl:
balance: stealing from belts and stealing keys themselves both separately make it 25% harder to be stolen, which can stack
balance: a lot more semi-failure states for pickpocketing that show the victim that something was stolen
balance: pickpocketing scales now with the speed stat as well for it's success chances, but the victim now also contests their speed in addition to their perception
add: new UI for stealing from containers that shows your chances
/:cl:

### DIFF
--- a/code/game/objects/items/rogueweapons/mmb/steal.dm
+++ b/code/game/objects/items/rogueweapons/mmb/steal.dm
@@ -53,18 +53,19 @@
 		sate_addiction(/datum/charflaw/addiction/kleptomaniac)
 	grant_pickpocket_xp(victim, exp_to_gain)
 
-/datum/component/storage/proc/pickpocket_show(mob/living/carbon/human/thief, is_belt = FALSE)
+/// A steal-odds label floated over one grid item. Client-local, so only the thief sees it, never the mark.
+/atom/movable/screen/pickpocket_odds
+	plane = ABOVE_HUD_PLANE
+	layer = ABOVE_HUD_LAYER + 1
+	mouse_opacity = MOUSE_OPACITY_TRANSPARENT // clicks fall through to the item beneath
+	maptext_x = 0
+	maptext_y = 20
+	maptext_width = 32
+	maptext_height = 12
+
+/datum/component/storage/proc/pickpocket_show(mob/living/carbon/human/thief)
 	if(isnull(thief.client) || !show_to(thief))
 		return FALSE
-	var/atom/real_location = real_location()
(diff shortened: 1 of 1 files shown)


Ephemeralis:
Don't scale pickpocketing with SPD. We tried this on Scarlet Reach briefly and it was heinously bad, since the pairing of high mobility + probable DE + cranked theft chances produces a class that is super good at theft, and on the off chance it does get caught, can immediately disengage with typically very little counterplay.

Contesting user FOR vs target FOR was much more interesting and also linked into stress events pretty nicely too.

Arturlang:
will make it use FOR

wikimody:
I realised, with this rework you can reverse pickpocket people. Since the pickpocket opens the targets equipment, you can add item from your other hand into said equipment

wikimody:
It very very cool thing to be able to do, for example planting evidence on somebody, but i wish it was a intended mechanic instead of this weird half bug (for example you cant reverse pickpocket a big item onto somones back for example, and cant open the equipment of a belt or sathel that are empty)

wikimody:
If i coudl also ask for a way to chose with of the items im robing in the inicial 50-50. If somone has a shield on their bag for example i tend to pick up the shield instead of getting to rob their bag with is both more obvius and less ufull to me as a pickpocketer

Arturlang:
> I realised, with this rework you can reverse pickpocket people. Since the pickpocket opens the targets equipment, you can add item from your other hand into said equipment

i should make that a official feature lmao

PR #2198, opened 2026-07-23
Minor vampire refactor and clan coven changes

Arturlang:
## About The Pull Request
fixes a lot of small issues with vampires not reverting their changes when the antag datum is removed, also ENSURES that victims that are being sired are fully revived and can't be zombie and a vampire at the same time.

also fixes obfsucate not working by making sneaking not just always set your alpha to 255 if it's not at 255, this is a potentially a fucky change, but we'll see, it's worth it in the long run for other features

also also this fixes custom vampire lords not getting the traits that *some* static ones get, instead it's a thing that all clan lords now get, unlimited stam, heavy armor, extra bloodpool, etc, so you're not fucking yourself over by playing a custom clan
also fixes some of the custom clans getting a duplicate blood heal, replacing that with something else, nosferatu got quietus, eora got siren,
there's a wretch clan leader but it wasn't used earlier and i didn't make it be used now, was trying to minimize balance changes

fixes https://github.com/Rotwood-Vale/Ratwood-2.0/issues/1124

## Testing Evidence
lots of joining the game, adding antag datum to a mob, removing it, seeing if all the vampire properties are gone

## Why It's Good For The Game
vampires working gooder is a benefit to all vampire mains

## Changelog

:cl:
fix: being de-vamped should correctly remove all vamp traits and vamp eyes and vampiric visuals
fix: some clans not getting the same essential traits and buffs for clan leaders, including custom clans
fix: obfsucate not working due to sneak making you always invisible
fix: vampire siring not reviving you properly and removing zombie-ism
fix: nosferatu being scared if you self-examine
balance: remove duplicate blood healing from nosferatu and eora, giving them quietus and siren covens instead
balance: all clan leaders who wouldn't usually get a lord-inherited spell get the bat shapeshift
/:cl:

### DIFF
--- a/code/__DEFINES/traits.dm
+++ b/code/__DEFINES/traits.dm
@@ -263,6 +263,8 @@
 #define TRAIT_GOODWRITER "Great Writer"
 #define TRAIT_COMBAT_AWARE	"Combat Aware"
 #define TRAIT_SILVER_WEAK "Silver Weakness"
+/// Held while wearing a silver psycross in a valid slot. Wards off a vampire trying to drink from you.
+#define TRAIT_WORN_SILVER_PSICROSS "worn_silver_psicross"
 #define TRAIT_ASSASSIN	"Assassin"
 #define TRAIT_EQUESTRIAN "Equestrian"
 #define TRAIT_REGROW_LIMBS "Regrow Limbs" //OV ADD, regrows lost limbs at cost of nutrition
--- a/code/datums/components/hideous_face.dm
+++ b/code/datums/components/hideous_face.dm
@@ -20,7 +20,7 @@
 
 /datum/component/hideous_face/proc/on_examine(mob/source, mob/user, list/examine_list)
 	var/mob/living/carbon/human/H = parent
-	if ((H.wear_mask?.flags_inv & HIDEFACE) && (H.head?.flags_inv & HIDEFACE))
+	if ((H.wear_mask?.flags_inv & HIDEFACE) && (H.head?.flags_inv & HIDEFACE) && user == parent)
 		return
 	examine_list += span_warning("[H]'s face is horrifying!")
 	if(!iscarbon(user))
--- a/code/datums/components/vampire_disguise.dm
+++ b/code/datums/components/vampire_disguise.dm
@@ -69,7 +69,13 @@
 	disguised = TRUE
 	COOLDOWN_START(src, transform_cooldown, 5 SECONDS)
 
-	// Restore human appearance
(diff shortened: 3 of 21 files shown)


Arturlang:
let me know about any further issues i can and need to fix, prefferrably via github issues

UntoldTactics:
> let me know about any further issues i can and need to fix, prefferrably via github issues

https://github.com/Rotwood-Vale/Ratwood-2.0/issues/1124

Though since then we've learned that it's really just that the silver psycross ONLY (and may not even then) functions when worn on neck. Ideally it should work no matter where its worn. Maybe it should work against all sort of mental-control vampire abilities as well.

Arturlang:
> > let me know about any further issues i can and need to fix, prefferrably via github issues
> 
> #1124
> 
> Though since then we've learned that it's really just that the silver psycross ONLY (and may not even then) functions when worn on neck. Ideally it should work no matter where its worn. Maybe it should work against all sort of mental-control vampire abilities as well.

fixed

Venithefox:
the punish minion verb for all thralls underneath you as the clan leader in the vampire tab just, does not work. clicking on it does absolutely nothing. it is supposed to let you stun, burn or even ash your own minions on command. but the verb is completely bugged and pressing it will not bring up the menu for you to punish. I was told that the menu CAN come up if you stand RIGHT NEXT to a minion but clicking the punishment still does nothing, haven't verified that though.

Dagonlives:
The vampire hypnotize is kind of weak. I once as an antag vampire failed to snare someone for 5 casts in a row.

This could use a look at perhaps without making it gobsmackingly overpowered, but currently it's quite weak!

UntoldTactics:
> The vampire hypnotize is kind of weak. I once as an antag vampire failed to snare someone for 5 casts in a row.
> 
> This could use a look at perhaps without making it gobsmackingly overpowered, but currently it's quite weak!

Honestly it's stupid strong for what it is. The code for calculating it goes something like d2 x every few points of int or something vs the vampires d2 x skill level of blood magic

However there's a +10 bonus to each roll if the victim has combat mode enabled (Making it that much harder to use against a prepared enemy)
I've only ever seen vampires using it to try to cheese people, honestly. 
The main 'issue' if you wanna say that is actually putting someone to sleep takes two successes or three.

Arturlang:
> The vampire hypnotize is kind of weak. I once as an antag vampire failed to snare someone for 5 casts in a row.
> 
> This could use a look at perhaps without making it gobsmackingly overpowered, but currently it's quite weak!

separate PR

PR #2199, opened 2026-07-23
Heat Dispersal System

Kroto-Ker:
## About The Pull Request

Adds a new heat dispersal system that allows for passive heating while next to fire sources, such as braziers and campfires. 

This also removes the get_hearers check on process that campfires rely on for their AOE buff. In exchange, being in a heated tile now regenerates your blue bar and heals you.

## Testing Evidence

It functions perfectly during a previous TM.

## Why It's Good For The Game

Fire now passively heats you simply by being near it. 
The campfire buff is replaced by a more efficient system.

## Changelog

:cl:
add: Heat dispersal system. Standing next to fires warms and heals you.
/:cl:


RadiantFlash7:
Kroto is correct in truth. While regenerating bluebar is nice, the point of it was to encourage players to hang around a heat source, creating dynamic gathering points. I even noticed recently that a lot of players were using the campfires to gamify the bluebar regen for doing tasks- which just wasn't what the goal was.

Adding a requirement to be sitting at a chair, or laying on a bed is a much better way to handle this.

ariaworld:
A third option should be wall leaning imo, so that standing still works but only if you stay still (I know wall leaning already does this, but being near to a fire should make you more relaxed)

zoktiik:
Heat dispersal: good
Trying to push fatigue regen nerf a second time, after the first had a joke title ("Temperature QoL") and got downvoted hard prior to Xander calling you out on it: bad

wikimody:
I dont see how making energy regen even less plesant afair then it alredy is helps anything

Unrelated, but every single heat source giving the warning heat mesage about sweeting isnt particularly good. Maybe remove it from the minor heating the firesources provide so its easier to see people who actualy need cooling down?

Shadow-Quill:
a potential middle-ground I can see is removing the stam regen buff from campfires, leaving it solely on hearths/fireplaces, as those are less of a "build and leave wherever" item, and generally are already in places people will gather.

Kroto-Ker:
> Heat dispersal: good Trying to push fatigue regen nerf a second time, after the first had a joke title ("Temperature QoL") and got downvoted hard prior to Xander calling you out on it: bad

It wasn't even intended to be a fucking nerf, it just got caught in the switch to the new system. I didn't realize they made campfires so OP.

Kroto-Ker:
I've changed it so being in a heated tile now provides the previous campfire buff's healing / blue regen effects.

PR #2204, opened 2026-07-24
Removes (most) polearm sweetspots

Pidgey-chan:
## About The Pull Request

Removes the sweetspot requirement from the spear thrust. As compensation, lowers the penetration of the 
intent to 45.

## Testing Evidence

Worked for years without these lines in the file.

## Why It's Good For The Game

De-Azureification.

Spears and thrusting polearms have been miserable and less than viable with the sweetspot requirement. Maintaining a consistent 2-tile range to use a slow intent with modest AP is not something workable. There are far superior options that do not occupy two hands like warhammers and knuckles - let's bring greatweapons back up to snuff.

## Changelog

:cl:
balance: Spear thrusts no longer require a sweetspot to penetrate, and penetrate for less.
/:cl:

### DIFF
--- a/code/game/objects/items/rogueweapons/melee/polearms.dm
+++ b/code/game/objects/items/rogueweapons/melee/polearms.dm
@@ -10,10 +10,8 @@
 	clickcd = CLICK_CD_CHARGED
 	warnie = "mobwarning"
 	hitsound = list('sound/combat/hits/bladed/genstab (1).ogg', 'sound/combat/hits/bladed/genstab (2).ogg', 'sound/combat/hits/bladed/genstab (3).ogg')
-	penfactor = 50
+	penfactor = 45
 	item_d_type = "stab"
-	effective_range = 2
-	effective_range_type = EFF_RANGE_EXACT
 
 /datum/intent/spear/thrust/oneh
 	name = "one-handed thrust"


Gug1234:
If you aren't able to keep out of someone's range and use the sweet spot, it's not an issue with the weapon; it's an issue with the player's stat line and skill.

This PR isn't necessary.

AzuriteEchelons:
I use polearms religiously, this PR is a massive cope.

zeroisthebiggay:
i do agree that polearms are kinda bad but they definitely hurt when someone is actually good with spacing them
they could probably be buffed in other ways that are not this

Pidgey-chan:
> oh so you just want to go back to the... polearms penetrating all armor from 1 tile range great... -1 from me

With 45 AP and the 30 damage of a wielded spear, 12 STR is required to do one (1) point of penetrating damage through plate or maille.

Yeetus223:
awful and unneeded

animalsandsuch:
This PR is great and beautiful! These guys have never played a round of Ratwood in their lives! I love you!

AllCorruptingWyrm:
this is absurd having just witnessed some stuff in game.

@Kroto-Ker please untm this for the time being.

AlucardHellsingChapel:
I think i have proved with last round how broken this pr is by being a spear skeleton stabbing 7 people to death

LittleEvil9:
Spears are weak trust, surely not a skill issue that they're unable to backpedal and hit from a 2 tile range on a weapon that pierces even plate armor.

This was changed ages ago because spears were absolutely dominant when used, we don't need to go back to that.

Pidgey-chan:
> Spears are weak trust, surely not a skill issue that they're unable to backpedal and hit from a 2 tile range on a weapon that pierces even plate armor.
> 
> This was changed ages ago because spears were absolutely dominant when used, we don't need to go back to that.

That's correct that skill doesn't play into it because backpedaling is inherently slower, yes. Relying on "proper spacing" is not something that's possible in this jittery combat system, a facehugger with a warhammer will always succeed at staying within 1 tile because they can't walk through their opponent even holding the movement key.

Carl made this change to great controversy and then fucked off without implementing any of the boons that accompanied it on Azure like 1-tile alternative intents. I daresay it is an experiment that you have simply gotten used to.

The spear is perhaps weaker than it's ever been in terms of damage potential. It existed in a stronger state than this for a long time alongside warhammer picks, smash bleedthrough, peeling, and grapplemaster iconoclasts. And I still don't think thrusting polearms are going to outclass those.

AzuriteEchelons:
Massive cope on your part.. greatswords got 1 tile bash and cut to cut away through armor, maces and warhammers do blunt damage and destroy any non padded armor. 

Spears got RAW ap as a trade off, you bleed people to death and chip their armor from a distance, once that is gone watch them suffer as you artery them from 2 tiles away. Massive cope, please never make PRs again brodacho

AlucardHellsingChapel:
@Pidgey-chan 80% of current garrison members in present round of desert town was using glaives/halberds/eagle's beak etc as it goes through all armor even, lux armor which is the strongest armor in the game stacking infinite bleeds on everyone this pr is massive cope on your part

Pidgey-chan:
> Massive cope on your part.. greatswords got 1 tile bash and cut to cut away through armor, maces and warhammers do blunt damage and destroy any non padded armor.
> 
> Spears got RAW ap as a trade off, you bleed people to death and chip their armor from a distance, once that is gone watch them suffer as you artery them from 2 tiles away. Massive cope, please never make PRs again brodacho

What? Is English your third language?



> @Pidgey-chan 80% of current garrison members in present round of desert town was using glaives/halberds/eagle's beak etc as it goes through all armor even, lux armor which is the strongest armor in the game stacking infinite bleeds on everyone this pr is massive cope on your part

Soldiers use polearms? Seems historically accurate to me, haha. But anyway, these will not pen blacksteel without significant strength (more than 14) or one of the steel ones. Polearms existed without sweetspots for a long time alongside warhammer picks, grapple-and-smash knuckles, and peeling, and never truly dominated even with 50 AP. Anecdotes about the garrison carrying spears can probably be attributed to people like (you) saying that they penetrate all armor. They don't.

Pidgey-chan:
Xander self-merges his templar buffs but I make a whole weapon category not useless and it's a wordless close OK. LOL.

UntoldTactics:
This shit is closed sadly but I honestly think this was a good move, better than current. Sweetspots ARE rewarding, but they genuinely do not work in the current balance dichotomy where the wide variety of blunt polearms are blatantly superior and do obscene damage and have good defense. Greatswords and specific swift balance swords etc are busted and delete people super fast, etcera. In comparison, polearms are super bad because of the need to maintain distance in a game where distance is harder to maintain when you're the person moving backwards (or even forwards sometimes), and the tools to keep distance don't really exist besides an occasional use special ability.

I simply had no reason to use a polearm that had sweetspots when grand maces, eagles beak and other polearms or simply put superior weapons exist that don't have such finnicky mechanics to work around and just let me wail on people.

halbdog:
Having played before the sweetspot system was added, I would consider it a good addition, as at that time polearms were very, very strong compared to other options. People would just click on your chest, stab you right through steel armour while being out of your reach, and even when you closed that gap, they would continue to do it while enjoying solid defence values as you bled out faster and faster. Sweetspots gave a viable counter to that, while keeping the weapon strong when used well.

However, the combat balance and general playstyle was very different. Most blunt weapons were extremely weak, barely able to break armour or hurt people, leaving peel as the primary strategy. But combat was also less common, was much less likely to be done in BIS gear, had fewer people involved, and just overall had less min-maxxing. It was slower, so the slow-but-steady bleed and pain buildup that polearms specialised in was much more effective. 

But with the massive blunt buffs, other balance changes, increase in players, and dramatic increase in minmaxxing, there are just so many better options, and the positioning that polearms need is much harder to achieve. They are bad because they are balanced, and so many other options aren't. 
So, I really don't think this PR should have just been shut down like this, removing sweetspots is a reasonable change with the current state of the game. (Though I think that keeping them and bringing other things in line would be better, but thats a much bigger change)

PR #2209, opened 2026-07-25
Lizard Accent back to Brazillian Portuguese

jgDarkWolf:
## About The Pull Request

Brings lizard accent back to portuguese

## Testing Evidence

image

image

## Why It's Good For The Game

Having more unique accents are better than having two of the same accent slightly different
Theres a hissy accent to add more ssss' to when people sssssssay a word with an sssss
As it previously was lizard accent was the same has hissssssssy accent but with like 1 less s.

## Changelog

:cl:
qol: Reverts lizard accent back to RW 1.0 lizard accent
/:cl:

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/combat/cleric.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/combat/cleric.dm
@@ -14,8 +14,8 @@
 		STATKEY_CON = 2,
 	)
 	subclass_skills = list(
-		/datum/skill/combat/wrestling = SKILL_LEVEL_JOURNEYMAN,
-		/datum/skill/combat/unarmed = SKILL_LEVEL_JOURNEYMAN,
+		/datum/skill/combat/wrestling = SKILL_LEVEL_EXPERT,
+		/datum/skill/combat/unarmed = SKILL_LEVEL_EXPERT,
 		/datum/skill/combat/polearms = SKILL_LEVEL_APPRENTICE,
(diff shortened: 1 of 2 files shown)


LittleEvil9:
If you want it to be fully accurate its easier to delete these:

"what's": "qual é",
"who's": "quem é",

Because it will lead to strange translations of: who's talking? > "Quem é talking?"
or
what's wrong? > "Qual é wrong?"
what's going on? > "Qual é going on?"

which is weird ASF.

You could change "who's" to > "who": "quem"
but it might still lead to weird speech.

Imo just add one which is "who are you" > "quem é você"

jgDarkWolf:
> If you want it to be fully accurate its easier to delete these:
> 
> "what's": "qual é", "who's": "quem é",
> 
> Because it will lead to strange translations of: who's talking? > "Quem é talking?" or what's wrong? > "Qual é wrong?" what's going on? > "Qual é going on?"
> 
> which is weird ASF.
> 
> You could change "who's" to > "who": "quem" but it might still lead to weird speech.
> 
> Imo just add one which is "who are you" > "quem é você"

Added the one line at the bottom, this json already existed and was pulled from RW 1.0

jgDarkWolf:
Its working on my local end on dream daemon, but on live test merge theres an issue with letters with dashes
image

image

jgDarkWolf:
I believe it was the multi word phrase added last second by your recommendation, that is now fixed

zoktiik:
Hood classic

Genevesh:
I missed the 'retardado', it better be back.

PR #2210, opened 2026-07-25
Give Froggemund neck peel protection

Gug1234:
## About The Pull Request

- add var bodyinherent = HEAD|NECK to froggemund helm

## Testing Evidence
Boilerplate var edit.

## Why It's Good For The Game
A minor tweak to make the helm a bit better. For all the vision you give up, having it work like a helm/gorget combo shouldn't shake up balance too much. 

## Changelog

:cl: Yuckuza
balance: added neck peel protection to the froggemund helm
/:cl:

### DIFF
--- a/code/modules/clothing/rogueclothes/headwear/helmet/heavy_helmet.dm
+++ b/code/modules/clothing/rogueclothes/headwear/helmet/heavy_helmet.dm
@@ -695,6 +695,7 @@
 	icon_state = "frogmouth"
 	item_state = "frogmouth"
 	emote_environment = 3
+	body_parts_inherent = HEAD|NECK // can't peel off the solid plate bottom portion of the helm
 	body_parts_covered = FULL_HEAD|NECK
 	flags_inv = HIDEEARS|HIDEFACE|HIDEHAIR
 	block2add = FOV_RIGHT|FOV_LEFT


jgDarkWolf:
Remove the 45 degree angle vision 🙏🏻

Gug1234:
> Remove the 45 degree angle vision 🙏🏻

No, it's soul. Without the limited vision cone, the helmet becomes the best in slot for anyone with maille training or higher before Blacksteel.

coffeecoffeecoffee:
I really, really wish it were... something like a 60 degree or 75 degree vision cone to make it not _quite_ so totally unbearable, but having played around with the vision vignette files before, I can imagine how hard that could be.

code/modules/clothing/rogueclothes/headwear/helmet/heavy_helmet.dm
```diff
@@ -695,6 +695,7 @@
 	icon_state = "frogmouth"
 	item_state = "frogmouth"
 	emote_environment = 3
+	body_parts_inherent = HEAD|NECK // can't peel off the solid plate bottom portion of the helm
```
Xander3359, line 698: Shouldn't this also be FULL_HEAD

Xander3359 (changes requested):

code/modules/clothing/rogueclothes/headwear/helmet/heavy_helmet.dm
Gug1234: don't believe so; FULL_HEAD would make ALL head target areas unpeelable. We only want head and neck specifically to be peel-proof, not eyes, skull, mouth, ears, etc.

PR #2218, opened 2026-07-27
Chestjak

ophaq:
## About The Pull Request
- Does the most incredibly OP chest buff in history and increases its mob container amount to 2.

## Testing Evidence
Literally a single number change.
chest

## Why It's Good For The Game
Closets could already do this. I feel like this just fixes an oversight. Please tell me how this could possibly be a negative to gameplay.

Anyway, I wanted to ERP in a chest the other day and couldn't even close the lid. This bumps it to 2 so someone can actually do such a thing if one wanted to. Chests already inherited rough sexcon rocking mechanics like beds and closets do too so it's just sovlfvl to allow them to make use of it with a higher body count.

## Changelog
:cl:
qol: Allows sex in chests by bumping their mob storage to 2 instead of 1.
/:cl:

### DIFF
--- a/code/game/objects/structures/crates_lockers/roguetown.dm
+++ b/code/game/objects/structures/crates_lockers/roguetown.dm
@@ -12,7 +12,7 @@
 	sellprice = 1
 	max_integrity = 200
 	blade_dulling = DULLING_BASHCHOP
-	mob_storage_capacity = 1
+	mob_storage_capacity = 2
 	allow_dense = FALSE
 
 /obj/structure/closet/crate/chest/gold


ophaq:
> Just admit that you want to larp as a mimic LE and i'll bribe xander to approve this.

It is true. I do want to larp as a mimic. I was also thinking of making non-subtle emotes show up instead of getting eaten when inside closed containers but I need to figure out the code for that.

Xander3359:
Massive chest buff of the year. Chest enjoyers rejoice.

PR #2222, opened 2026-07-27
Mordhau Grip adjustments

CloverIsLucky:
## About The Pull Request

Mordhau grip gets its bash intent nerfed from 1.2 damage multiplier to 1.1, gets 15 ClickCD and some swingdelay added back.
Sword Pomel Strike in compensation gains minuscule blunt chipping.

## Testing Evidence

Trust me, I believe it'll work.

## Why It's Good For The Game

I think I treated mordhau grip too well while Carl used to maintain here.

## Changelog

:cl:
balance: Mordhau Grip: BASH Damfactor and clickcd nerfed from 1.2 and 13 to 1.1 and 15 respectively; added some swing delay back.
code: Pomel Strike gets minuscule blunt chipping.
/:cl:

### DIFF
--- a/code/game/objects/items/rogueweapons/melee/polearms.dm
+++ b/code/game/objects/items/rogueweapons/melee/polearms.dm
@@ -135,7 +135,8 @@
 	attack_verb = list("bashes", "strikes")
 	penfactor = BLUNT_DEFAULT_PENFACTOR
 	damfactor = 1.1
-	clickcd = 13
+	clickcd = 15
+	swingdelay = 2
 	item_d_type = "blunt"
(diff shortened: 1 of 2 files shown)


Chief301:
While greatswords needed their mordhau touched, I agree, longswords are utterly reliant on  it and this is a heavy handed nerf to arguably one of the worst weapons in the game. It would be better to do what was done on SR and just reduce the greatsword’s force when using mordhau instead of an across the board nerf that makes it feel worse to use and makes longsword even worse than it already is. Please consider a reversion or adjust it to just be a nerf to the greatsword instead of a blanket nerf to longswords too.

PR #2226, opened 2026-07-27
Memcheck stuff + memory optimizations

Arturlang:
## About The Pull Request
additional memory debugging features + some z-level and turf related memory optimization,
major thing is that there's a config option for per-atom type and per-subsystem memory usage logging, and otherz's are now packed into z-levels based on how many z-levels up/down they need, separated far enough that even with XRAY you can't see it.
also changes the fillter turfs to be far more memory efficient

## Testing Evidence
loaded the game, checked the other sublevels existed

## Why It's Good For The Game
memory usage tools r good

## Changelog

shouldn't be user-facing


Arturlang:
dnm yet, i need to see if i can improve the windwos mem logging

PR #2230, opened 2026-07-27
AI Fix

Lezerchil:
## About The Pull Request

This is a combination of pulls done by Posshum.

It adds two new booleans to some mobs making them utilize special attacks or rmb attacks.
These two
```dm
	//If we utilize special attacks or not; All is handled within do_best_melee_attack() chain.
	var/special_attacker = FALSE
	//If we utilize our intents further outside of strong intent.
	var/smart_combatant = FALSE
``` 

I have some concerns if RW has additional mobs which CC doesn't. Those might break.
Though this should prevent that
```dm
		if(!did_we_change_intent) //Always default regardless.
			rog_intent_change(1)
``` 
If they do break spreading vars on them too should fix the issue.

I also have some concerns about this.
```dm
/mob/living/carbon/human/Initialize()
``` 
In RW it is usually written as Initialize(mapload). It might break mobs with new vars.

## Testing Evidence

I've taken a couple of vids just poking deranged knights in the face. They supposed to be smart and special.
No idea if it counts as evidence. I don't play the game.

https://github.com/user-attachments/assets/eb0b5624-1c64-4f0e-ab50-c9bccd62c343

https://github.com/user-attachments/assets/30c15f6d-9a57-41e0-a28b-31d558844269

## Why It's Good For The Game

People asked for it or so I heard. Smarter enemies are generally good. It makes combat with NPCs (hopefully) less trivial.

## Changelog

:cl:
add: Added 2 new booleans to manage AI behaviour
add: NPCs recieve a set of new actions they could take based on their designated behaviour.
balance: The following NPCs are now considered smart combatants: skeleton liches, skeleton soldiers, skeleton dreadnoughts, bog deserters, highwaymen, deranged knights, dungeon lizardmen, thieves.
balance: The following NPCs are now considered special attackers: militia, bog deserters, skeleton dreadnoughts, skeleton liches, orcs, sea raiders, dungeon lizardmen, drow, deranged knights.
:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/npc/_npc.dm
+++ b/code/modules/mob/living/carbon/human/npc/_npc.dm
@@ -888,14 +888,14 @@
 				apply_status_effect(/datum/status_effect/debuff/swapped_intent_npc) //45 seconds before we swap to a new weapon intent entirely.
 
 			if(special_attacker && prob(50) && !has_status_effect(/datum/status_effect/debuff/specialcd)) //Only if we use specials...
-				if(length(weapon_special_intents) > 1)
+				if(length(weapon_special_intents) >= 1)
 					if(possible_rmb_intents & /datum/rmb_intent/strong)
 						swap_rmb_intent(/datum/rmb_intent/strong)
 						try_special_attack(target)
 						return TRUE //We used our special intent on the target as soon as we could.
 
 			if(smart_combatant && prob(50)) // Only if we use rmb intents...
-				if(length(possible_rmb_intents) > 1)
+				if(length(possible_rmb_intents) >= 1)
 					if(!has_status_effect(/datum/status_effect/debuff/feintcd))
 						if(possible_rmb_intents & /datum/rmb_intent/feint && rmb_intent != /datum/rmb_intent/feint && prob(50))
 							swap_rmb_intent(/datum/rmb_intent/feint)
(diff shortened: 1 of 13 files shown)


RadiantFlash7:
Please fix the changelog before this is full merged, @Lezerchil

code/modules/mob/living/carbon/human/npc/_npc.dm
```diff
@@ -830,6 +836,15 @@
 	Weapon = get_active_held_item()
 	OffWeapon = get_inactive_held_item()
 
+	//Feint Riposte check before we do any further attacks to teach our enemy a lesson.
```
CameronWoof, line 839: What is the probability this gets used? Does it become effectively impossible to defend against smart NPCs because of this?

Lezerchil: Always, unless on a cooldown.

Lezerchil: No, you still can "defend" against NPCs. You just need to wait for the right moment. You cannot open up with a "defend" move and there's no guarantee that "defend" will always work in the middle of a fight.

Lezerchil:
> Please fix the changeling before this is full merged, @Lezerchil

Not sure what are you talking about. Aren't changelings part of an old code @RadiantFlash7 ?

RadiantFlash7:
> > Please fix the changeling before this is full merged, @Lezerchil
> 
> Not sure what are you talking about. Aren't changelings part of an old code @RadiantFlash7 ?

...autocorrect. Changelog. Still, Thank you for fixing it.

PR #2241, opened 2026-07-28
Unapologetic maul buff

Gug1234:
## About The Pull Request

- Standard Maul weapon defense increased from 3 when wielded (dogshit) to 6
- Grand Maul weapon defense increased from 6 to 9 when wielded
- Grand Maul force raised to 36, one above the grand mace. You are a stupid -int idiot who invested their statpack towards 15 strength; you shouldn't have your big dumb hammer end up being lesser than a boring mace, flail, etc. 
- Dwarven Maul and Spiked Maul weapon defense increased to 7 when wielded
- Fix Malum templar Maul by giving it the is_silver var

## Testing Evidence
Simple number changes only.

## Why It's Good For The Game
At expert maces, the standard maul against an expert weapon skill opponent had around a 40% parry chance. That's abysmal; you were better off running with dodge expert maul drakian than actually parrying with your big funny hammer. The maul also suffered from lower force and lacking range compared to the grand mace or general power potential of something like the steel flail. 
This aims to give mauls both the defense power and stopping power to be viable in the current combat landscape. Shouldn't become too too strong given heavy balance gets utterly stomped by anyone running dodge. 

## Changelog

:cl: Yuckuza
balance: Standard Maul weapon defense increased from 3 when wielded to 6.
balance: Grand Maul weapon defense increased from 6 to 9 when wielded. Grand Maul force raised from 34 to 36. 
balance: Dwarven Maul and Spiked Maul weapon defense increased to 7 when wielded.
fix: Malum templar maul now functions like a silver weapon as intended.
/:cl:

### DIFF
--- a/code/game/objects/items/rogueweapons/melee/blunt.dm
+++ b/code/game/objects/items/rogueweapons/melee/blunt.dm
@@ -658,6 +658,7 @@
 	desc = "Forged from the legacy of dwarven rock-hammers, this maul’s holy steel and divine runes grant it immense power. \
 	Unwieldy to those weak of arm or faith, its mighty blows have the strength to shatter both stone and skull alike."
 	icon_state = "malumhammer"
+	is_silver = TRUE
 	minstr = 8//Handled by the unique interaction below. Inverted to start, since they spawn with it, and funny stuff can happen.
 
 /obj/item/rogueweapon/mace/maul/grand/malum/pickup(mob/living/user)
@@ -668,7 +669,7 @@
 	..()
 
 //This thing is warded. For fluff. And because it's COOL, we give them silver blessings.
-//+1 DEF from it, too. For a total of 7 defence when wielded.
+//+1 DEF from it, too. For a total of 10 defence when wielded.
 /obj/item/rogueweapon/mace/maul/grand/malum/ComponentInitialize()
 	AddComponent(\
 		/datum/component/silverbless,\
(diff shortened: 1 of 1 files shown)


UntoldTactics:
I've modified my PR to not cause unintended consequences if this PR is merged. (Like making the kargrund maul have less defense than its parent object for no reason), though you could also technically fix the silver functionality of the kargrund maul in this PR without it leaving relevant scope and do it all in 1 PR rather than two.

Gug1234:
> I've modified my PR to not cause unintended consequences if this PR is merged. (Like making the kargrund maul have less defense than its parent object for no reason), though you could also technically fix the silver functionality of the kargrund maul in this PR without it leaving relevant scope and do it all in 1 PR rather than two.

done and done

UntoldTactics:
> > I've modified my PR to not cause unintended consequences if this PR is merged. (Like making the kargrund maul have less defense than its parent object for no reason), though you could also technically fix the silver functionality of the kargrund maul in this PR without it leaving relevant scope and do it all in 1 PR rather than two.
> 
> done and done

Closed my PR. Make sure to list those changes in changelog too!
My burden is lifted, I am FREE...

Seashine:
As far as I can see (I have no idea if this is an oversight or not) but _no_ templar gets silver weapons, so the kargrand maul being silver now is an outlier. I find this strange considering that the church is specifically supposed to fight heretics, but it seems to be the standard?

PR #2250, opened 2026-07-29
Hemophages no longer heal red and blue from blood.

2whatever2:
## About The Pull Request
As the title says, it removes Vital_Essence, which is given in 12u by blood when its processed by a hemophage. Which has these specific stats:

/datum/reagent/medicine/vital_essence/on_mob_life(mob/living/carbon/M)
	if(M.blood_volume < BLOOD_VOLUME_NORMAL)
```
		M.blood_volume = min(M.blood_volume + 5, BLOOD_VOLUME_NORMAL)
	if(!HAS_TRAIT(M, TRAIT_INFINITE_STAMINA))
		M.energy_add(10)
	var/list/wCount = M.get_wounds()
	if(wCount.len > 0)
		M.heal_wounds(5) // something between red and strong red. We cannot drink water 
	if(volume > 0.99)
		M.adjustBruteLoss(-1 * REM, 0)
		M.adjustFireLoss(-1 * REM, 0)
		M.adjustOxyLoss(-1, 0)
		M.adjustOrganLoss(ORGAN_SLOT_BRAIN, -1 * REM)
		M.adjustCloneLoss(-1 * REM, 0)
	..()
```
	For a point of reference, most max_energy goes to around 1100 to 1400 for combat roles.

## Testing Evidence
Dont need to test this its just a number change/deletion

## Why It's Good For The Game
Right, to break it down. Initially made this pr to nerf the healing, and maybe remove the energy gain, and add softlimits to how much you can hold, and making carrying blood in general remove it.

But then I realized.. "Hey, isn't Hemophage a Vice?"

currently, even 0.1u of blood gives you 12 Vitae Essence, which metabolizes at a rate of 1u per tick, the same as health potions and mana potions.

Sneak mode lets you sip from bottles, which is 1u for most containers, meaning a bottle gives you 50 sips of 0.1u blood for a total of 50 x 12 vitae essence, aka, 600u of blue and red.

Even if I nerfed and tuned this all down to make it almost worthless, or make it only so it heals you on bloodrinking only... This is _still_ a vice in the first place. Most vices simply just remove their mood debuff for treating it, while some just utterly ruin you (the stat vices/compliance, and a few others.)

Vices should stay vices, and if there's any bonus for taking it, they should be minimal, not to this extent though.

If a maintainer asks me to retain _some_ healing then I'm willing to make the healing only happen while you're sucking blood from a live person.

## Changelog

:cl:
balance: Hemophages no longer heal or regain energy. from drinking blood or dirty blood.
/:cl:


Aracino:
while you're here can you make it so that drinking blood actually satiates your hunger for longer than like 10 minutes. Shits extremely annoying, more so than any other vice.

Skyglazer:
Genuinely astonished that this was ever a thing. And astonished further that people would be downvoting this PR. Goodness gracious, what a world. Now I fully understand why the 'vice' was so popular on the sorts of characters I saw running it.

PR #2254, opened 2026-07-29
riding a saiga in deep water drains stamina

Xander3359:
## About The Pull Request
riding a saiga in deep water drains stamina

## Testing Evidence
Tested on local

## Why It's Good For The Game
Saigas should not be megaboats when it comes to moving in water.
This at least limits their potential to cross entire oceans for free.

## Changelog

:cl:
balance: Riding a saiga in deep water will now cost the same stamina as swimming normally would.
/:cl:


ariaworld:
Wait wouldn't it make more sense if the saiga was just slowed in the water?

Aracino:
> Wait wouldn't it make more sense if the saiga was just slowed in the water?

It would, just slow it to swim speed

annikaRU:
Keep in mind even if you run out of stamina the Saiga will continue moving so this effectively just does nothing

Aracino:
This is kind of a Why PR, as in, why not just have water slow down Saiga riders normally. Your mount is not a boat.

PR #2260, opened 2026-07-30
Admin PQ Adjustment Message Tweak

Shadow-Quill:
## About The Pull Request

The prompt that asks an admin how much PQ to adjust by has a hidden gate where, if you have the right permission, doesn't actually clamp how much PQ you can add/deduct.

This PR hides the allowed range if you're able to surpass it; it only changes how the message appears, it does not otherwise change the behavior of those verbs.

## Testing Evidence

image

Note no +ADMIN permission.

## Why It's Good For The Game

Incorrect information bad.

## Changelog

:cl:
admin: The prompt that appears when adjusting a player's PQ will no longer show the "allowed adjustment range" of -20 to 20 if you have the right permission to ignore that requirement anyways.
/:cl:


Shadow-Quill:
I got annoyed.

PR #2273, opened 2026-08-01
Gives LE Wilderness Guide Trait

ophaq:
## About The Pull Request
See title.

## Testing Evidence
One line change.
LEwildernessguide

## Why It's Good For The Game
The class that is meant for the wilds and is basically druid+ doesn't even prevent ambushes or triggering maneaters. If you are fighting in or roaming the wilderness, that gets annoying very quickly. I guess it also makes LE more useful as a bog wretch guide but some antags already prevent ambushes when traversing such as bandits anyway.

## Changelog
:cl:
add: Adds wilderness guide to LE so they don't trigger ambushes or maneaters.
/:cl:


wikimody:
This can be countered by trewalking with i belive Le has by default, and it effectivly buffs his enemies who cant escape as quickly when pursuing him and trigering a ambush

Xander3359:
> This can be countered by trewalking with i belive Le has by default, and it effectivly buffs his enemies who cant escape as quickly when pursuing him and trigering a ambush

This actually has an aura effect that prevents others from triggering ambushes as well.
Mind you it has limited range, but it still covers the screen and this is pretty reasonable to give.

wikimody:
> > This can be countered by trewalking with i belive Le has by default, and it effectivly buffs his enemies who cant escape as quickly when pursuing him and trigering a ambush
> 
> This actually has an aura effect that prevents others from triggering ambushes as well. Mind you it has limited range, but it still covers the screen and this is pretty reasonable to give.

Thats why i consider it a nerf, LE deals with abmbushes much better then his enemies, and both benefit from that trait

PR #2275, opened 2026-08-01
[PORT] Improvements/bug/fixes/changes from AP Mage Edition

RadiantFlash7:
## About The Pull Request
We have fallen severely behind AP on a variety of content/features/fixes/improvements all the way back to January.
This PR Ports a slew of mage-related PRs from January to late March relating to magic changes, additions bugfixes and tweaks. Notably, it does not include the spellblade PR in an attempt to reduce the amount of changes in one PR.  I fully intend to look at porting spellblades, and then Mage Gameplay loop 2/3 after that, likely individually. At some point after all that is handled, I intend to tackle Economy 2/3 if it's not yet done

The following PRs are included in this port

- null for cast time reduction https://github.com/Azure-Peak/Azure-Peak/pull/5589
- Fix of lingering light effects on mobs https://github.com/Azure-Peak/Azure-Peak/pull/5581
- cooldown reduction runtime fix https://github.com/Azure-Peak/Azure-Peak/pull/5669
- another fix of potential moblights and casting bugs https://github.com/Azure-Peak/Azure-Peak/pull/5669
- spell button examines https://github.com/Azure-Peak/Azure-Peak/pull/5844
- Visible spell cooldown timers https://github.com/Azure-Peak/Azure-Peak/pull/5730
- Telegraphs penalties for wearing unacceptable armor https://github.com/Azure-Peak/Azure-Peak/pull/6303
- Removes Aerosolize https://github.com/Azure-Peak/Azure-Peak/pull/5740
- Fire Cascade/firewalker https://github.com/Azure-Peak/Azure-Peak/pull/5130
- clickdelay improvements for spells from https://github.com/Azure-Peak/Azure-Peak/pull/5793

## Testing Evidence
image
image

## Why It's Good For The Game
The majority of these are bugfixes/improvement PRs. Those are self explanatory.
Spell cooldown timers are nice to have for mages to actually see what is going on. Examining a spell to see various stats and changes that occur from stats/int are important as well, so information isn't gated behind code-diving.
Fire Cascade/firewalker are good additions to the mage spell list (With plenty more pending with the further mage PRs Ported).

The most controversial change is removing of Aerosolize.  In truth, this spell is problematic. It's description and usage specifically encourage players to make the most lethal cocktail possible in a container, and then use it to rapidly kill anyone thats within the gas for two seconds. Taking a vein from AP's PR, It functionally incorporates the worst effects of TG's lethal chem gas mixes, with very few of the upsides you would find otherwise.

[...]

PR #2278, opened 2026-08-01
[PORT] Spellblades rework/ Guard Deflecting magic

RadiantFlash7:
## About The Pull Request
This PR continues the work #2275 started, porting the Spellblade rework done by [WeNeedMorePhoron](https://github.com/WeNeedMorePhoron)
It also ports over the action of raising your guard to block/deflect spells.
Dependant on 2275 to go through as it builds on the changes and tweaks in that PR.

## Testing Evidence
WIP

## Why It's Good For The Game
I'm slowly working on modernizing our magic system to be somewhat in line with AP's to a degree. To that end, starting back on missing changes from Feb/march is where we're at right now.
The spellblade rework provides unique spells/weapon usage for those roles that isn't just mage with a sword/mace.
Out of this PR, the guard option is the most controversial. Magic lacked any counter play beyond a shield, or antimagic ring, leaving it more or less entirely disjointed from the martial side of combat. I'm a smidge skeptical on blocking spells, but if someone pulls up a guard too early, a mage can just can't their spell cast and wait till it runs out- or use a more spammable spell.

## Changelog

:cl:
add: Spellblade, 3 unique subtypes
add: Guard action on defend intent will now block arcane/holy magic spells on hit (Using up the guard in the process)
/:cl:


Pidgey-chan:
I like the idea of parrying spells, but keep in mind that mages can spit out spells far faster than the cooldown for defend intent and that a lot of the problematic spells (repulse) wouldn't be parryable. We should also look into bringing back mouthgrabs.

RadiantFlash7:
> I like the idea of parrying spells, but keep in mind that mages can spit out spells far faster than the cooldown for defend intent and that a lot of the problematic spells (repulse) wouldn't be parryable. We should also look into bringing back mouthgrabs.

Thats very much intentional. Guard isn't intended to make people immune to magic permanently.

Kroto-Ker:
I'm not sure I like the addition of more reaction-time based mechanics to combat, especially around something like this. I think a better solution would be possible, though I'm not wholly opposed to this one.

RadiantFlash7:
> I'm not sure I like the addition of more reaction-time based mechanics to combat, especially around something like this. I think a better solution would be possible, though I'm not wholly opposed to this one.

Personally I was a bit iffy on it to begin with.
But it matches up with use of the guard mechanic for martial means, and ties martial mechanics a bit more to magic ones. Which is generally good, rather than magic mechanics existing in a vacuum.
That said, guard has a hefty cooldown, so it really only serves a purpose to pull if you identify someone pulling an arcane spell with a higher cast time and dangers effects. Of which, the magic user can cancel their cast, and switch to a more spammable option to trigger the guard sooner to follow up with another option after.

Pidgey-chan:
Martial parries with guardup gives one guaranteed parry probably among a string of 90% parries; parrying spells gives one guaranteed spell parry and then 0% for all other spells through the guard cooldown.

RadiantFlash7:
Going to close this. I'm not super satisfied with how spellblades ended up with it, and the guard deflect is a bit controversial. We'll look at doing magic systems later.

PR #2282, opened 2026-08-02
Brigandine and children all have leather studded protection stats

Gug1234:
## About The Pull Request

- swap brigandine armor var from ARMOR_PLATE to ARMOR_LEATHER_STUDDED

## Testing Evidence
Simple var edit.

## Why It's Good For The Game
Gives them a niche. Right now every brigandine type armor piece uses studded leather apart from the medium brig, which is functionally just a medium cuirass with better coverage and integ.

This gives it and coat of plates a better niche as the jack of all trades armor line between leather and plate.Brigandine vambraces (the arms) were recently changed to leather_studded for the same reason.

## Changelog

:cl: Yuckuza
balance: Brigandine and subtypes now use ARMOR_LEATHER_STUDDED across the board. In laymen's terms: 80 Blunt, 80 Slash, 60 Stab, 40 Pierce ki
/:cl:

### DIFF
--- a/code/modules/clothing/rogueclothes/armor/brigandine.dm
+++ b/code/modules/clothing/rogueclothes/armor/brigandine.dm
@@ -6,7 +6,7 @@
 	icon_state = "brigandine"
 	blocksound = SOFTHIT
 	body_parts_covered = COVERAGE_ALL_BUT_LEGS
-	armor = ARMOR_PLATE
+	armor = ARMOR_LEATHER_STUDDED
 	prevent_crits = list(BCLASS_CUT, BCLASS_STAB, BCLASS_CHOP, BCLASS_BLUNT, BCLASS_TWIST)
 	allowed_sex = list(MALE, FEMALE)
 	nodismemsleeves = TRUE
@@ -72,7 +72,7 @@
 	icon_state = "brigandine2"
 	blocksound = SOFTHIT
 	body_parts_covered = COVERAGE_FULL
-	armor = ARMOR_PLATE
+	armor = ARMOR_LEATHER_STUDDED
 	prevent_crits = list(BCLASS_CUT, BCLASS_STAB, BCLASS_CHOP, BCLASS_BLUNT, BCLASS_TWIST)
 	allowed_sex = list(MALE, FEMALE)
 	nodismemsleeves = TRUE


vincentthemongrel-commits:
Finally someone made a fix to that porting AP's brigandine stats.
THANK YOU.

Pidgey-chan:
This makes the medium brig and coat of plates kind of garbage?

Gug1234:
> This makes the medium brig and coat of plates kind of garbage?

Makes them better; the studded leather statline is like mini blacksteel. Very even spread. Brigandine and coat won't get mulched by blunt and still has highly competent slash and stab.

Pidgey-chan:
> > This makes the medium brig and coat of plates kind of garbage?
> 
> Makes them better; the studded leather statline is like mini blacksteel. Very even spread. Brigandine and coat won't get mulched by blunt and still has highly competent slash and stab.

Noooo, why would I want light armor defenses? 60 stab is garbage (for a non-dodging medium/heavy armor build). A dagger doesn't even have to pick you, they can just swiftly stab chest. Spears hitting the sweetspot are getting at least 20 through (at 10 st, usually more!).

The only upsides here, blunt and piercing, are covered by the magical kevlar linen padded gambeson, the meta pick (only viable) underlayer.

Let me put it this way. With a padded gambeson under plate, you have... a layer with 90 blunt, 100 slash, 80 stab, and 80 piercing.

With a padded gambeson under studded leather, you have 90 blunt, 80 slash (still enough probably cutting AP is garbage), 60 stab (bad!!!!!) and 80 piercing.

With mail under studded leather, you have 80 blunt, 100 slash, 80 stab, and 40 piercing (bad!!).

UntoldTactics:
Considering the Brigandine is medium armor I don't honestly see the benefit of this. Not only does this basically entirely remove an option for arm coverage medium armor that also covers chest (and often the highest integrity medium armor overall), this basically makes them...what?

Higher integrity studded leather armor that covers the arm. That's what. It's...not really jibing with what I feel like it'd do.

Note the coat of plates is also HEAVY armor. It's not something you can compare with the brigandine due to that. Whilst the concern of blunt is being spammed on the chest, the fact of the matter is there's quite a lot of players that still use sharp cutting weapons and wearing such a modified brigandine is begging to be getting smacked around by the average greatsword/polearm enjoyer.

The other brigandine pieces are genuinely their own niche because they fit inbetween full leather options and full steel options. The brigandine bracer has the integrity of iron bracers (202 in game), but the stats of studded leather armor. It doesn't have quite the integrity of say, heavy leather bracers, but has better overall overages. It doesn't have the 270 in game integrity of steel bracers, but obviously has superior blunt and piercing protections.

Brigandine  wouldn't really...fit any specific niche, here. It's not just about making them stand out, but the fact that the brigandine is one of the highest integrity medium armor pieces and one of the few readily accessible ones with arm coverage. Scalemail is its leg-covering counterpart but that only has 270 in game integrity vs the brigandines 315 in game. It's not a real competition, there.

Tldr; Brigandine is one of the better medium armor items and without a 'superior' plate armor it doesn't make sense. The trade-off of studded leather is better armor coverage for lesser integrity than plate counterparts. (generally speaking) The issue here is the brigandine IS the plate counterpart, and there is no better medium armor item aside the blacksteel cuirass or maybe the fluted cuirass (covers legs not arms)

Coat of plates is heavy armor, and thus invalid in the conversation. It also doesn't have that much more integrity than a regular brigandine, anywho. Maybe bringing it down to medium armor could create the interesting dynamic, I suppose. Brigandine as studded armor tier with the coat of plates replacing what the brigandine currently is. The issue is that this would either require the coat of plates buffed a little, or the brigandines integrity nerfed a little. The current difference between the two is about 50 integrity points. Consider in game that brigandine bracers and steel bracers is about 68 integrity point difference.

Further any roles that start with the brigandine would want to be shifted to starting with the coat of plates, to keep their current armor dynamics. We don't need say, the marshal wearing studded armor when one day they were clad in steel.

vincentthemongrel-commits:
so not merged? damn why

Gug1234:
> so not merged? damn why

#2445

PR #2295, opened 2026-08-03
Gives Drow Dark Vision

skippymcspeederton-dot:
## About The Pull Request
* Gives the Underdark's most iconic race the ability to see in it!

## Testing Evidence
Simply adds one trait to the Dark Elves, should be fine

## Why It's Good For The Game
Drow give others a racism stress moodlet when being examined, and what is their saving grace? They aren't held up by some cobweb?

With this change, they have a stronger identity and truly belong in the Underdark and the shadows as they're renowned to be.

## Changelog

:cl:
balance: Added Dark Vision to Dark Elves
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/roguetown/elf/elfd.dm
+++ b/code/modules/mob/living/carbon/human/species_types/roguetown/elf/elfd.dm
@@ -20,7 +20,7 @@
 	individual, such as a kinder heart fleeing from a brutal society that scorns them for their \
 	less cruel nature. However, not every dark elf seen on the surface can be safely assumed as \
 	kind, for some leave the Underdark simply to find their own greater heights of power.<br>\
-	(+1 Intellect, +1 Perception, Webwalker Trait)"
+	(+1 Intellect, +1 Perception, Webwalker and Dark Vision Traits)"
 
 /*
 	Former RT Desc: These guys were undead which doesn't really fit considering now you have a ton of them walking around.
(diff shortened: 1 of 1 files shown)


Pidgey-chan:
Discrimination cannot be said to exist when Drow can freely be nobility or other power roles, and even if it did discrimination is no cause for racial imbalance - dark vision is flavorful for Drow, however.

This should be accompanied with some stat loss.

skippymcspeederton-dot:
> Discrimination cannot be said to exist when Drow can freely be nobility or other power roles, and even if it did discrimination is no cause for racial imbalance - dark vision is flavorful for Drow, however.
> 
> This should be accompanied with some stat loss.

Drow have stress_examine = TRUE, which means that just by examining them in-game, your character will receive stress. 

Yeah they can absolutely be in power roles, but this is a negative trait that they can't get rid of and it isn't even communicated to the player when picking this race. You're talking about the concept of racism, but in this game there are mechanics for it.

Compared to the wood elf who DOESN'T have stress_examine = TRUE, who can almost instantly escape into trees and run along leaves safely and with extra speed. This is a trait that can save your lyfe and escape encounters multiple times in one round.

I think Darkvision with the Drow's current stat line up serves as a more or less equal balance when you take into account the racism mechanics. Webwalker is an interesting bit of flavor, but may as well not exist with how little use it sees and how easily it is to counter webs.

LittleEvil9:
I would love it if Drows had true dark vision only to be semi blinded while suffering stat loss whenever its day.

legalillegaldoc:
> Discrimination cannot be said to exist when Drow can freely be nobility or other power roles, and even if it did discrimination is no cause for racial imbalance - dark vision is flavorful for Drow, however.
> 
> This should be accompanied with some stat loss.

drow had darkvision before, and it got swapped for webwalker.

drow are by no means an overpowered race and you can just go zizo for darkvision, i don't personally think they need a stat debuff especially when other races get far better passives and stats

LittleEvil9:
> drow are by no means an overpowered race and you can just go zizo for darkvision, i don't personally think they need a stat debuff especially when other races get far better passives and stats

I agree, but I didn't meant darkvision as we have it with Zizo or Noc. I meant true darkvision in the sense of vampire or deadite vision where the whole screen is mostly at the same brightness level, hence a stat debuff when its day to compensate.

DorfGuy1:
yeah please do. People stress out when they look at you.

DorfGuy1:
> I would love it if Drows had true dark vision only to be semi blinded while suffering stat loss whenever its day.

flavorful but a bit much.

DorfGuy1:
This does not work quite right, I think there is a bug. I joined as a drow architecht, and it did not work

PR #2305, opened 2026-08-04
Jar helm, symbol of avarice, and throwing arm trait.

Gug1234:
## About The Pull Request

- Adds Symbol of Avarice, a new helmet dropped rarely by mimics on death. Raises fortune by +5 but lowers constitution by -5 when worn.
- Adds the jar helm, a pottery wheel crafted and cooked helmet that breaks easily but gives you the throwing arm trait when worn.
- Adds the throwing arm trait, which grants all thrown objects +0.5 throw speed, +1 throw range, and +2 throw force. Only accessible via the jar helm. 

Additionally:
- The symbol of avarice makes mimic noises when hit, and mimic death noise when broken.
- blocksound var can now be given any sound or list of sounds to use as the armor block sound rather than just the 4 plate, chain, cloth, etc. strings.

## Symbol of Avarice:
mimicdrop8
mimicdrop7
mimicdrop6
mimicdrop5
mimicdrop4
mimicdrop3
mimicdrop2
mimicdrop1

## Jarhelm:

jar7
jar6
image_2026-08-04_140038505

jar4
jar3
jar2
jar1

https://github.com/user-attachments/assets/d7e047fc-5f29-4b43-a0eb-7d560b7ff469

jartest

## Helmet sounds/blocksound changes

https://github.com/user-attachments/assets/5d68e28c-299b-4585-b0db-89ded22bfe02

## Why It's Good For The Game
Two silly helmets for our silly roleplay game.
Jar helm is something to maybe seek out the potter for if you want to minmax throwing weapons. 
Mimic head is decently rare, and fortune isn't important enough of a stat for the helmet to become meta. 

## Changelog

:cl: Yuckuza
add: Added the symbol of avarice, a rare drop from mimics that grants 5 fortune in exchange for -5 constitution.
add: Added the jar helm, a pottery craftable helmet that grants the throwing arm trait when worn.
add: Added the throwing arm trait, granting +0.5 throw speed, +1 throw range, and +2 throw force to any thrown item.
/:cl:


Yeetus223:
let me solo her

EvenInDeathIStillServe:
soul is back

code/__DEFINES/roguetown.dm
Xander3359, line 270: This in general might be due for a better refactor to handle more cases, but for now this is probably fine like this.

code/game/atoms_movable.dm
```diff
@@ -592,6 +592,9 @@
 		return
 	return hit_atom.hitby(src, throwingdatum=throwingdatum)
 
+/atom/movable/proc/get_effective_throwforce(datum/thrownthing/throwingdatum)
+	return throwforce + (throwingdatum?.bonus_throwforce || 0)//bonus is typically 2 extra force unless the var is reused elsewhere
```
Xander3359: Why is this an AM proc instead of a throwing datum proc?

code/modules/clothing/rogueclothes/headwear/helmet/heavy_helmet.dm
```diff
@@ -831,3 +831,62 @@
...
+	var/active_item = FALSE
+
+/obj/item/clothing/head/roguetown/helmet/heavy/mimic/equipped(mob/living/user, slot)
+	. = ..()
+	if(slot == SLOT_HEAD)
+		active_item = TRUE
+		to_chat(user, span_hypnophrase("Dead mimic flesh envelops your head, slippery, cold, and wet. The beast's hunger washes over you; you feel starved and emaciated, as if something has sapped your CONSTITUTION; however, in its place is left a gnawing, greedy avarice coupled with FORTUNE enough to sate it."))
+		user.change_stat(STATKEY_LCK, 5)
+		user.change_stat(STATKEY_CON, -5)
+	return
```
Xander3359: Should be changed to an early return.

```
if(slot != SLOT_HEAD)
  return
```

code/modules/clothing/rogueclothes/headwear/helmet/heavy_helmet.dm
```diff
@@ -831,3 +831,62 @@
...
+
+/obj/item/clothing/head/roguetown/helmet/heavy/mimic/dropped(mob/living/user)
+	..()
+	if(!active_item)
+		return
+	if(active_item)
```
Xander3359: This doesnt need to be checked because you already have the early return above.

code/modules/clothing/rogueclothes/headwear/helmet/heavy_helmet.dm
```diff
@@ -831,3 +831,62 @@
...
+	. = ..()
+	if(slot == SLOT_HEAD)
+		active_item = TRUE
+		to_chat(user, span_hypnophrase("Dead mimic flesh envelops your head, slippery, cold, and wet. The beast's hunger washes over you; you feel starved and emaciated, as if something has sapped your CONSTITUTION; however, in its place is left a gnawing, greedy avarice coupled with FORTUNE enough to sate it."))
+		user.change_stat(STATKEY_LCK, 5)
+		user.change_stat(STATKEY_CON, -5)
+	return
+
+/obj/item/clothing/head/roguetown/helmet/heavy/mimic/dropped(mob/living/user)
+	..()
```
Xander3359: . = ..()

code/modules/clothing/rogueclothes/headwear/helmet/heavy_helmet.dm
```diff
@@ -831,3 +831,62 @@
...
+	if(!active_item)
+		return
+	if(active_item)
+		to_chat(user, span_hypnophrase("The vile corpse pulls free with a squelch. Your head is left wet, glossed slick with the creature's mucus... You feel your vigor return!"))
+		user.change_stat(STATKEY_LCK, -5)
+		user.change_stat(STATKEY_CON, 5)
+		active_item = FALSE
+	return
```
Xander3359: Return at the end does nothing.
Unless you are returning something other than `.` or stopping the proc for some reason (early returns)

Xander3359 (changes requested):

code/game/atoms_movable.dm
Gug1234: Hacky code to ask the item itself, rather than the throwing datum, how much damage to do. Your suggestion is cleaner. I was drinking Fernet and store-brand Dr. pepper while making this shit mostly copying patterns from TG and previous statbuffing items like the rings. 

Resolved as of latest commit: the helper lives in throwing.dm

code/__DEFINES/roguetown.dm
Gug1234: Can't think of where we expand it; what else uses get_armor_sound and the blocksound var?

CaydenT:
As funny as the jar helm is ....... Why? Why, specifically, does it boost throwing? Its not even magical right?

Why throwing???? I don't think this will be a balance problem but its

Like

Far and away too silly and nonsensical.

I'm not offended by this just very perplexed. Its a bit "lolsorandum xD"

Gug1234:
> As funny as the jar helm is ....... Why? Why, specifically, does it boost throwing? Its not even magical right?
> 
> Why throwing???? I don't think this will be a balance problem but its
> 
> Like
> 
> Far and away too silly and nonsensical.
> 
> I'm not offended by this just very perplexed. Its a bit "lolsorandum xD"

Because it boosts throwing items in Elden Ring.

PR #2269, opened 2026-07-31
Skulls for the Madstar—Vorpal for Viscious Axe 

Gug1234:
## About The Pull Request

- Adds vorpal var to the Graggar armament vicious greataxe.

## Testing Evidence
1-line var addition.

## Why It's Good For The Game
Currently, there are no antag weapons with vorpal. The var is limited to Necran Templar Axe Respite, Dungeoneer Terminus Est, and Astratan Templar Executioner Sword Solar Judge.

I can think of no better patron or weapon deserving of this wonderful, wonderful var than Graggar's big funny axe. 

## Changelog

:cl: Yuckuza
add: Added vorpal to the Vicious Greataxe, decapitate the weak, skulls for the madstar!
/:cl:

PR #2306, opened 2026-08-04
Revert "Skulls for the Madstar—Vorpal for Viscious Axe "

AllCorruptingWyrm:
Reverts Rotwood-Vale/Ratwood-2.0#2269

After a discussion with a few staff members, and me reaching out to some players, and reading the original notes: Vorpal was not supposed to be given out--but it was. This reverts the PR.


AllCorruptingWyrm:
image
For context.

I do not think that its rare if anyone can sacrifice lux for it to get it via ritualist.

Pidgey-chan:
Decaps never should have been hugboxxed to begin with. Bring stakes and drama back to combat

PR #2310, opened 2026-08-05
Heretics actualy see eachother

wikimody:
## About The Pull Request
Addded examine text for assendant worshiper outside of mathiosians (only they had one)
I also added the seer check to the symbols as it wasn't present also

## Testing Evidence
3 ifs, compiles

## Why It's Good For The Game
I was in the midle of coding for my heresiath PR, when i realized only mathiosians had examine text for non seers. I belive it is advertised as part of being a heretic that you do see other heretics, and it just wasnt true for anyone except them.

If somone has better idea of examine text feel free to suggest them, I wrote what seemed to be fitting.

Heretic seer just didn't have the symbols for some reson?

## Changelog

:cl:
fix: Assendant worshipers from the gods other then mathios now can actualy recognize eachother (only mathiosians could before)
fix: Heretic seer now sees the heretical symbols of other heretics
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/examine.dm
+++ b/code/modules/mob/living/carbon/human/examine.dm
@@ -1208,15 +1208,18 @@
 /// Same as get_heretic_text, but returns a simple symbol depending on the type of heretic!
 /mob/living/proc/get_heretic_symbol(mob/examiner)
 	var/heretic_text
+	var/seer = FALSE
 	if(HAS_TRAIT(src, TRAIT_DECEIVING_MEEKNESS))
 		return
-	if(HAS_TRAIT(src, TRAIT_COMMIE) && HAS_TRAIT(examiner, TRAIT_COMMIE))
+	if(HAS_TRAIT(examiner, TRAIT_HERETIC_SEER))
+		seer = TRUE
+	if(HAS_TRAIT(src, TRAIT_COMMIE) && (HAS_TRAIT(examiner, TRAIT_COMMIE)||seer))
 		heretic_text += "♠"
-	else if(HAS_TRAIT(src, TRAIT_CABAL) && HAS_TRAIT(examiner, TRAIT_CABAL))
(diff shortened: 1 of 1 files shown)


LittleEvil9:
Gonna be really annoying when someone goes "Aren't you X worshipper, why aren't you helping me?" despite never having done anything to hint that.

wikimody:
> Gonna be really annoying when someone goes "Aren't you X worshipper, why aren't you helping me?" despite never having done anything to hint that.

Maybe, but that was already possible with the relations menu, since for some reson it has that implemented properly

Elli-Skala:
TECHNICALLY  the bug is that non math heretics can see others in the menu

This feature was removed by AP when we added faith salute (to buff xylxians) and we got this version when we did the last AP parity. right before azure peak merged the removal in the contact menu

coffeecoffeecoffee:
I don't think they _should_ be able to recognize eachother automatically for doing nothing at all, actually.

wikimody:
// ASCENDANT CULTIST TRAITS (all of them recognize each other)
#define TRAIT_COMMIE "Blessing of Matthios" //recognized by bandits as an ally
#define TRAIT_CABAL "Of the Cabal" //Zizo cultists recognize each other too
#define TRAIT_HORDE "Anointed" //Graggarites also recognize each other
#define TRAIT_DEPRAVED "Fallen" //Baothans also recognize each other

wikimody:
I cant say for sure whether they shoudl or shoudnt see eachother, but it was definitly meant to work that way

wikimody:
Either way even if this PR gets rejected i will make a separate one with the 2 bug fixes i found along the way, just without the examine text change

RadiantFlash7:
I'm of the opinion that heretics shouldn't identify each other in the known persons menu, rather then making heretics visible to each other.
Back in the day it was just Zizo worshippers and matthites that could, but I suppose it got crept further into all heretics.

RadiantFlash7:
> Either way even if this PR gets rejected i will make a separate one with the 2 bug fixes i found along the way, just without the examine text change

I would appreciate you doing so, though at this time i'm inclined to deny this PR.

PR #2313, opened 2026-08-05
The Lash: Dungeon Calling—Dungeonmaster Trait grants bonus chance to apply whip scarring wounds 

Gug1234:
## About The Pull Request

- TRAIT_DUNGEONMASTER adds a used *=2 multiplier to wound/scarring crit probability. Works the same way TRAIT_NUTCRACKER does with wound/cbt (AKA, it doubles the chance for the wound to roll).
- Aimed intent now grants an explicit bonus to wound/scarring as it does for other bclass wounds

## Testing Evidence
Math changes reusing existing logic shape from TRAIT_NUTCRACKER. 

## Why It's Good For The Game
Permanent scarring wounds are a punishment you dont see often. More often than not, dungeoneers who try to apply these wounds end up causing bleedout on the victim before ever having the opportunity to give them their flogging scars. 

This rectifies that, allowing dungeoneers to more consistently apply permanent scarring to prisoners without having to flog them for minutes, heal, then flog some more to try and roll the crit. 

Having nasty, painful scarring on your back that causes mechanical pain is a good way to incentivize someone to not rush back into a fight. 

## Changelog

:cl: Yuckuza
balance: TRAIT_DUNGEONMASTER, possessed by the dungeoneer and slave master, doubles the probability of whip scarring wounds on targets, the same way TRAIT_NUTCRACKER does for CBT wounds. 
balance: Aimed intent now grants a 10% bonus to the probability of whip scarring wounds. 
/:cl:


code/modules/surgery/bodyparts/bodypart_wounds.dm
```diff
@@ -290,8 +290,13 @@
...
 		used = round(damage_dividend * 20 + (dam / 3))
-		if(user && istype(user.rmb_intent, /datum/rmb_intent/strong))
-			dam += 10
+		if(user)
+			if(istype(user.rmb_intent, /datum/rmb_intent/strong))
+				dam += 10
+			if(HAS_TRAIT(user, TRAIT_DUNGEONMASTER))
+				used *= 2
+			if(istype(user.rmb_intent, /datum/rmb_intent/aimed))
+				used += 10
```
Xander3359: Should switch this order so that the +10 is also multiplied by 2 for extra benefit

code/modules/surgery/bodyparts/bodypart_wounds.dm
```diff
@@ -380,6 +385,10 @@
 		if(user)
 			if(istype(user.rmb_intent, /datum/rmb_intent/strong))
 				dam += 10
+			if(HAS_TRAIT(user, TRAIT_DUNGEONMASTER))
+				used *= 2
+			if(istype(user.rmb_intent, /datum/rmb_intent/aimed))
+				used += 10
```
Xander3359: same here

Xander3359 (changes requested):

code/modules/surgery/bodyparts/bodypart_wounds.dm
Gug1234: seems strong but bet

zoktiik:
i kinda think something like this shouldn't just be restricted to dungeoneer. what if you fw whip intents/combat intent

Gug1234:
> i kinda think something like this shouldn't just be restricted to dungeoneer. what if you fw whip intents/combat intent

Permanent scarring is really strong; everyone having access to higher chances for one wound in the game you cannot heal is ill-advised. We want dungeoneers putting this on, ideally, someone who is being punished with flogging that brands them and gives them wound pain the rest of the round. What we don't want are heretic spies with crazy strength buffs giving everyone perma wounds because of a virtue or something.

zoktiik:
Oh ok, I was confusing scarring wounds with something cosmetic.

PR #2325, opened 2026-08-07
Equalizes psydonic grand mace defense value with regular grand mace

UntoldTactics:
## About The Pull Request
As said on the tin, equalizes the Psydonic Grand Mace's defense values with the regular grand mace.
Basically a copy paste of the bonus that the grand mace gets, given to the Psydonic Grand Mace.

## Testing Evidence
image

## Why It's Good For The Game
The psydonic grand mace should not be inferior to a regular grand mace. It unironically ends up with 8 defense (vs grand mace 9) when wielded even with the blessing bonuses, due to lacking the notable wdefense bonus that the grand mace gets but the psydonic does not.

## Changelog

:cl:
balance: Psydonic Grand Mace now has the same wielded defense bonus as the Grand Mace.
/:cl:

### DIFF
--- a/code/game/objects/items/rogueweapons/melee/blunt.dm
+++ b/code/game/objects/items/rogueweapons/melee/blunt.dm
@@ -460,6 +460,7 @@
 	force_wielded = 35
 	minstr = 12
 	smelt_bar_num = 2
+	wdefense_wbonus = 5
 	is_silver = TRUE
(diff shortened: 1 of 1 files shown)


RadiantFlash7:
A relatively small oversight. None of the other silver weapons are worse at blocking then their counter parts. Thanks for the PR.

PR #2323, opened 2026-08-06
Deserttown Update

RadiantFlash7:
## About The Pull Request
This PR contains a host of additions, fixes and some new content for desert town
The list of the current changes is below:
----
- Fyritius now spawns randomly in grass tiles at a low chance, and has hard spawns at the forgotten Oasis
- purifying flasks now purify bath water into regular drinking water
- a massive host of decorative desert structures, including temple objects
- Implementation of dunes. Line of sight blockers that act like grass and can trigger ambushes. Good for making quick get aways, but be careful of ambush potential.
- Deep Dunes ambush list overhauled
- Threat level of deep dunes reduced, minimum threat level increased.
- new npc raider outfit, dunewell raiders
- Antlion mob w/ tambeable, saddleable variants
- Ifrit mob- Fire beast. Quite toasty
- Ifrit Matriach, Boss of a dungeon, very dangerious
- 3 new types of Sandworms (hatchlings, stalkers, elders) Sandstorms bury underground and charge their opponents- as long as they have sand to burrow into.
- New Ifrit temple dungeon. Some adventurers appear to have cracked open a sealed prison. Oops.
- a variety of new POis on the main level of the map
- Fleeting Repose (Mirage Oasis)- a set of tents around a healing spring oasis. This oasis teleports at down between 4 locations. It will not teleport players or items with it- it's a mirage after all.
- Obelisks with ancient glyphs, readable with an int stat check. Find out some lore, or hints for adventuring.
- obelisks with a gem enlaid, restore energy to nearby human mobs during the day, drain energy from nearby mobs at night.
- walking on Deserttown tiles now have a chance to leave tracks
- Dust devils no longer spawn in  town
- A variety of minor fixes throughout desert town, adjusting names on doors and missing tiles/correcting things.
- sandstorm goggles have been given to the Azebs and can be bought from the merchant. Wearing sandstorm goggles removes the wrap overlay from being in a sandstorm. best paired with a hijaab (hijaabs prevent energy loss from being in a sandstorm
- quicksand- A pile? Puddle? Of quick sand, stepping onto it sinks you into it. standing in it slowly drains your energy (blue bar), players can resist to remove themselves out of it- though completing the task of freeing ones self with take up a substantial amount of stamina on completion

-----------------------
This PR is already huge, but I have 3 remaining things I want to add to desert town eventually.  I'll be keeping these three to a separate PR.
That would be bring a town crier varient for the desert town to run the fighting/gladiator pit, with the option of fighting mobs for coin.

[...]

PR #2328, opened 2026-08-07
Changes default probability of leaving a track

Heheniy:
## About The Pull Request
Changes default probatility of leaving a track from 0 to 10, mostly to fix DT not having tracks on most of the map

The author of map can revisit probabilities for every tile type later

## Testing Evidence
varedit

## Why It's Good For The Game
Kinda a fix?

## Changelog

:cl:
balance: default probability of leaving a track on turf 0 -> 10; will mostly affect desert town for now
/:cl:


RadiantFlash7:
I'm going to deny this as it has implications for other tiles. That, and the issue is fixed in #2323.

PR #2341, opened 2026-08-07
Unapologetic Disciple Skin Buff

Gug1234:
## About The Pull Request

- Changes disciple skin armor var to ARMOR_GNOLL_STANDARD

## Testing Evidence
One line var change

## Why It's Good For The Game
Disciple skin presently is actively worse than the virtue tough skin, covering less. 
For the guy purpose built for tanking hits with unarmed parries, his armor options were woefully underpowered compared to what any other grappler can get. 

This makes the tank monk actually tanky, while still having a clear weakness in his hands, head, feet, and from being out speeded by 90% of the enemies you'd expect the inquisition to be engaged with. 

It's a better way, and frankly more fun, way to buff them that isn't just giving disciple master wrestling (see prior PR) 

## Changelog

:cl: Yuckuza
balance: Disciples skin significantly buffed to stand apart from the tough skin virtue armor and give the class a much needed boost.
/:cl:

### DIFF
--- a/code/modules/clothing/rogueclothes/armor/regenerating.dm
+++ b/code/modules/clothing/rogueclothes/armor/regenerating.dm
@@ -108,7 +108,7 @@
 	I AM LIVING ON A FUCKING CORPSE. HE IS THE WORLD, AND THE WORLD IS ROTTING AWAY. \
 	HEAVEN CLOSED ITS GATES TO US, LONG AGO.' </br>'YET, HIS CHILDREN PERSIST; AND AS LONG AS THEY DO, SO MUST I. \
 	HAPPINESS MUST BE FOUGHT FOR.'"
-	armor = list("blunt" = 30, "slash" = 50, "stab" = 50, "piercing" = 20, "fire" = 0, "acid" = 0) //Custom value; padded gambeson's slash- and stab- armor.
+	armor = ARMOR_GNOLL_STANDARD//unique instead of worse version of the virtue tough skin
 	prevent_crits = list(BCLASS_CUT, BCLASS_BLUNT)
 	repair_time = 20 SECONDS
 	max_integrity = 300


MickDick1:
Unlike gnolls disciples still have armor that isn't just their skin on top.

To clarify since it was not revealed in the PR, gnoll skin values are:
ARMOR_GNOLL_STANDARD list("blunt" = 70, "slash" = 95, "stab" = 90, "piercing" = 80, "fire" = 40, "acid" = 0)

This is also on a class that gets critical resistance as a trait.

CameronWoof:
> Unlike gnolls disciples still have armor that isn't just their skin on top.
> 
> To clarify since it was not revealed in the PR, gnoll skin values are: ARMOR_GNOLL_STANDARD list("blunt" = 70, "slash" = 95, "stab" = 90, "piercing" = 80, "fire" = 40, "acid" = 0)
> 
> This is also on a class that gets critical resistance as a trait.

These stats are disgustingly overtuned for anything other than a gnoll (which should be picking out specific targets who opted in). 

Do **not** apply this to anything else. It's literally three? Four? times better than the current line?

UntoldTactics:
To note: the ONLY chest armor the disciple can get besides the skin armor itself is the monk robe which is a padded gambeson tier item. Requires legendary skincraft so it NEEDS a tailor.

That being said I concur and the armor values should probably be a bit more sane. That also being said, the integrity is only 270 in game for the skin armor, so whilst strong, it isn't as tanky as a gnolls in pure integrity.

An unarmed disciple also has to retain a very notable weakness on their hands. The glove armor slot item they get that boosts unarmed damage has a piddling 80 integrity or so, and a smart enemy (especially a grabbed enemy) can obliterate them by slashing that a lot.

Personally I'd suggest something along the padded armor tier line or leather armor tier line, rather than using the values of an antagonist armor, though with all I've said in mind.

Ephemeralis:
I'd rather we just restored unarmed parry functionality rather than buffing every unarmed class to the gills to accommodate for its yawning absence.

Gug1234:
gonna try a few different things on my local, and come back with some realistic numbers/tweaks

PR #2350, opened 2026-08-08
[Balance] Disables Wretch Necromancer

IronDragoon:
## About The Pull Request
Very basic PR, just disables the Wretch Necromancer subclass. 

## Testing Evidence

Ran, cannot select necromancer. Everything else works as expected. 

## Why It's Good For The Game

Currently we are struggling with significant issues with this specific subclass. With it being available to Wretches (which are almost always in round), it means most of the time it will be present in 9/10 rounds. We have found the class incredibly difficult to balance, and has led to repeated large conflicts with minimal RP due to their ability to summon large numbers of skeletons. 

We are drafting plans to move the Necromancer to a seperate minor antag role, but as this is seen as a pressing concern, it is being disabled right now to prevent further issues in the meantime. 

## Changelog

:cl:
balance: Wretch Necromancer is currently not selectable in round, pending a rework of the subclass to a full blown seperate minor antagonist
/:cl:

### DIFF
--- a/code/game/objects/structures/mineral_doors.dm
+++ b/code/game/objects/structures/mineral_doors.dm
@@ -1275,6 +1275,6 @@
 	keylock = TRUE
 	grant_resident_key = TRUE
 	resident_key_type = /obj/item/roguekey/townie// should be every wretch class - ideally we can get resident_role to accept lists but until then this'll do
-	resident_advclass = list(/datum/advclass/witch, /datum/advclass/wretch/licker, /datum/advclass/wretch/deserter, /datum/advclass/wretch/deserter/maa, /datum/advclass/wretch/berserker, /datum/advclass/wretch/hedgemage, /datum/advclass/wretch/necromancer, /datum/advclass/wretch/heretic, /datum/advclass/wretch/heretic/spy, /datum/advclass/wretch/outlaw, /datum/advclass/wretch/poacher, /datum/advclass/wretch/plaguebearer, /datum/advclass/wretch/pyromaniac, /datum/advclass/wretch/vigilante, /datum/advclass/wretch/blackoakwyrm, /datum/advclass/wretch/antipope, /datum/advclass/wretch/ancientchampion)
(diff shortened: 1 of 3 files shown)


CameronWoof:
Necromancer is the most common abuser of dark crystals, but any Zizite with Ritualist can do the same thing that Necromancer does. The only difference is that Necro spawns with a crystal already in their inventory. 

I just opened a PR this morning that halves the number of sentient minions a dark crystal can support (from 2 to 1) and moves Necromancer down to a single slot (down from 2) which should help curb Necro a bit and also stop the current Necro players from just moving to a new class and doing the same thing.

UntoldTactics:
Heresiarch can basically act as a necromancer+ if you ask me, btw. As the above comment says, due to how dark crystals work it's not even the class alone. It's just most convenient for necromancer to do, but any other zizite can amass a decent horse. (heretic and heresiarch being best for it since they have ritualist of course and access to lacriama)

IronDragoon:
These are reasonable concerns, I believe for now we will just make this one but will see what maints decide upon in discussion with senior admins

Ephemeralis:
Necromancer could probably be solved by having a 'summon budget' split between the skeleton NPCs they can spawn and the player-driven ones they can get from the black crystal. The main issue with them is the time they get to snowball if left alone for longer than like, half an hour.

Capping their max player skeletons at like 4 and their NPC skeletons at like 6 or 8 should give them enough breathing room to be a threat without being the round-dictating thing they currently are.

AzuriteEchelons:
> Necromancer could probably be solved by having a 'summon budget' split between the skeleton NPCs they can spawn and the player-driven ones they can get from the black crystal. The main issue with them is the time they get to snowball if left alone for longer than like, half an hour.
> 
> Capping their max player skeletons at like 4 and their NPC skeletons at like 6 or 8 should give them enough breathing room to be a threat without being the round-dictating thing they currently are.

This assumes you fight about 5-6 people in a skirmish, when in reality its the entire town + mercs/adv bots

Shell-Knight:
Outright content removal is bad bad bad and we should fix fix fix instead

AllCorruptingWyrm:
we should fix necromancer as shell said.

RadiantFlash7:
Necromancer has long since gotten away from the original intent when I tweaked it way back when. The flood of skeletons with a delay nightly has lead to them needing to be dealt with early, or players suffering from a multiskeleton swarm. This- is not the intention of the wretch role.

That said, We've come up with an alternative option to this, and will be closing this PR in favor of it.

PR #2361, opened 2026-08-09
Removes map vote tally

IronDragoon:
## About The Pull Request

This is not a PR from 'staff', rather that personally I am not a fan of the map tally system. Currently it does little of use beside mean that a small handful of people vote DT every round and it will eventually just force that map, often during peak hours when often only 5-10 people voted for it. I'm not a fan of this system, and so this removes it.

## Testing Evidence

Tested locally, I ran some votes and changed maps, it seems to work as expected.

## Why It's Good For The Game

Personally I think holding a vote then choosing the least voted option purely based on random other votes from players who may not even be playing at the time is silly

## Changelog

:cl:
del: Removes the map vote tally system
/:cl:

### DIFF
--- a/code/controllers/configuration/entries/general.dm
+++ b/code/controllers/configuration/entries/general.dm
@@ -520,26 +520,3 @@
 
 /datum/config_entry/flag/reopen_roundstart_suicide_roles_command_report
 
-/// The minimum number of tallies a map vote entry can have.
-/datum/config_entry/number/map_vote_minimum_tallies
-	default = 1
-	min_val = 0
-	max_val = 50
-
-/// The flat amount all maps get by default
-/datum/config_entry/number/map_vote_flat_bonus
-	default = 5
-	min_val = 0
-	max_val = INFINITY
-
-/// The maximum number of tallies a map vote entry can have.
-/datum/config_entry/number/map_vote_maximum_tallies
-	default = 200
-	min_val = 0
-	max_val = INFINITY
-
-/// The number of tallies that are carried over between rounds.
-/datum/config_entry/number/map_vote_tally_carryover_percentage
-	default = 100
-	min_val = 0
-	max_val = 100
--- a/code/controllers/subsystem/map_vote.dm
+++ b/code/controllers/subsystem/map_vote.dm
@@ -1,4 +1,3 @@
-#define MAP_VOTE_CACHE_LOCATION "data/map_vote_cache.json"
 #define SS_INIT_SUCCESS 2
 
 SUBSYSTEM_DEF(map_vote)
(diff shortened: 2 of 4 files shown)


RadiantFlash7:
Is it a bit less then ideal during lowpop hours, Yes.
Is it the best option we have to prevent rockhill 24/7? Also yes. The voting method functionally prevents the tyranny of the majority from rolling constantly.

A potential option would be to cap the maximum amount of tallies lower, so player voting has more an impact, but that’ll be exceptionally hard to balance for high pop hours

coffeecoffeecoffee:
THANK CHRIST.
Players from hours ago should by no means be dictating the map that we end up playing 9 hours later.

This was implemented with the hopes that the "attention" of being forced to play these other maps would get their numerous flaws fixed.
**Those changes haven't come**. I have said it and will say it again; the map should be _fun to play first, BEFORE_ expecting people to play on it. If it can't win in a vote, _then maybe some things need to be adjusted with the map!_
The "tyranny" of actual democracy is a significantly better system than this. _Let people vote to play the map they actually want to play._ If that's rockhill 24/7 - **then maybe the other maps need some actual work done first.**

albsh08-code:
> Is it a bit less then ideal during lowpop hours, Yes. Is it the best option we have to prevent rockhill 24/7? Also yes. The voting method functionally prevents the tyranny of the majority from rolling constantly.
> 
> A potential option would be to cap the maximum amount of tallies lower, so player voting has more an impact, but that’ll be exceptionally hard to balance for high pop hours

Don't get me wrong, but if the people want rockhill, they should get rock hill, just because someone voted 6 hours ago to play Dt and most likely isn't playing the Round that gets effected, should mean the people who played dun world or Dt don't have a say over someone who isn't currently gonna play the next round, kinda silly.

AllCorruptingWyrm:
ngl i'd rather have Dunn world/rockhill swap and no DT.

Pidgey-chan:
> Is it a bit less then ideal during lowpop hours, Yes. Is it the best option we have to prevent rockhill 24/7? Also yes. The voting method functionally prevents the tyranny of the majority from rolling constantly.
> 
> A potential option would be to cap the maximum amount of tallies lower, so player voting has more an impact, but that’ll be exceptionally hard to balance for high pop hours

How is votes of people not even in the round counting remotely fair

If I vote for Desert Town, go to bed, and 2 rounds later when I'm asleep my tally leads to Desert Town being played by people who vote for Rockhill, how is that desirable for me or them or anyone involved? Tallies are sooo fucking stupid

Skyglazer:
Please god, let this go through. I'm tired of the forced DT rounds. I like the occasional Dun World we get with the tally system, but the DTs do not justify it.

NotGoldenglow:
How about having tallies expire after, say, 2 rounds? This greatly caps them within the lifetime of those who'd want to see DT/DW, while also preventing the 'I voted 16 hours ago' deal.

coffeecoffeecoffee:
> How about having tallies expire after, say, 2 rounds? This greatly caps them within the lifetime of those who'd want to see DT/DW, while also preventing the 'I voted 16 hours ago' deal.

How about letting the people vote for what map they want to play on next, and respecting their wishes?

zoktiik:
Remove it or nerf it. I get what it's trying to solve but it needs a HARD nerf

IronDragoon:
I have reworked this into a new PR.  https://github.com/Rotwood-Vale/Ratwood-2.0/pull/2376

PR #2365, opened 2026-08-10
Fixes Fortify's 50% buff applying twice, Fix's surgery tools causing damage

RadiantFlash7:
## About The Pull Request
Embedded tools with theSURGICAL_TOOL flag no longer cause injury in patients every life tick. This should make surgery much faster as it won't frequently divert healing ticks to 2 damage generated on the limb with the tools.

This PR also fixes fortify's fifty percent buff applying twice, leading to a proper 1.5x heal rate instead of 2.25x it was currently doing.

## Testing Evidence
image

## Why It's Good For The Game
I want to fix our current healing before we make any more balance tweaks, so we have accurate data to balance around.

## Changelog

:cl:
fix: embedded surgery tools no longer damage on tick
fix: miracle heals no longer have a fortify 1.5x buff apply twice if healing someone who's been fortified

/:cl:


code/modules/mob/living/carbon/life.dm
```diff
@@ -211,6 +211,9 @@
 /mob/living/carbon/handle_embedded_objects()
 	for(var/obj/item/bodypart/bodypart as anything in bodyparts)
 		for(var/obj/item/embedded as anything in bodypart.embedded_objects)
+			if(embedded.item_flags & SURGICAL_TOOL)
+				to_chat(src, span_danger("[embedded] in my [bodypart.name] hurts!"))
```
Xander3359: This will chat spam the fuck out of you if you have multiple embedded surgical tools (like retractor + hemostat)

\+ It goes off every process. Oh yeah, we gonna chat spam the everloving shit out of you.

roguetown.dme
Xander3359, line 1: Check your DME

Xander3359 (changes requested):

code/modules/mob/living/carbon/life.dm
RadiantFlash7: ...Ah hell. The other ones are a probability. You're right. I'll fix that right quick

PR #2368, opened 2026-08-10
SLOOC Bug Fix, LOOC Ban Separation, No LOOC while Dead

hugmeorelse:
## About The Pull Request

Someone should double-check the code and then TM. I was not able to test this fully due to limitations.

As discussed and requested by players, this disables LOOC after death. Yet, it allows some flexibility. Furthermore, players can be safely LOOC banned, while allowing SLOOC usage. That way there is always an ability from a player to communicate if need be.

**Death Restrictions**

- LOOC is disabled after death once more.
- SLOOC still works while you are dead in your body, so a corpse can still talk to whoever is nearby.
- Admins are exempt only while aghosted. An admin whose character dies and ghosts normally is blocked like any other player until they aghost.

**LOOC and SLOOC Ban Separation**

- A LOOC ban no longer blocks SLOOC. So, existing rows keep role = 'LOOC' and are simply no longer queried by the SLOOC path.
- No SLOOC ban role was added. Server ban remains the permanent escalation for SLOOC misuse.
- Added MUTE_SLOOC, independent of MUTE_LOOC in both directions. Wired into the player panel mute row and into cmd_admin_mute so the mute is logged with the right channel name.
- SLOOC chat lines now carry two links, [MUTE SLOOC] and [MUTE LOOC], each toggling only its own channel.

**Bugs Fixed**

- Fixed the @ prefix path double sanitizing the message. 
- Moved do_subtlelooc's sanitize to the first line of the proc, matching do_looc
- The SLOOC target picker now only lists mobs with a client. NPCs and disconnected bodies used to appear as valid targets. No more of that!
- The target dialog was titled "Subtle Emote". It now says "SLOOC".

## Testing Evidence

It compiles and runs. Tested it using some local workarounds to avoid using the DB and you can ban LOOC usage without banning SLOOC usage. You can no longer LOOC while dead. The SLOOC and LOOC temporary mute buttons work as intended. Also, it seems to block out targets that do not have a client.

## Why It's Good For The Game

LOOC while being dead cause unnecessary disruption, however, we allow them to SLOOC if needed.

Also, bug fixes good.

## Changelog

:cl:
del: LOOC no longer works after death. SLOOC still works while your body is on the floor, but ghosts lose both.
admin: LOOC bans and LOOC mutes no longer apply to SLOOC, and SLOOC has its own temp mute on the player panel.
admin: SLOOC lines in chat now carry separate MUTE SLOOC and MUTE LOOC links.
fix: Apostrophes and quotes no longer show up as &#39; when sending SLOOC through the @ prefix.
qol: The SLOOC target list no longer offers NPCs or disconnected players.
qol: The SLOOC target window is no longer titled "Subtle Emote".
/:cl:

### DIFF
--- a/code/__DEFINES/admin.dm
+++ b/code/__DEFINES/admin.dm
@@ -7,6 +7,7 @@
 #define MUTE_ADMINHELP	(1<<3)
 #define MUTE_DEADCHAT	(1<<4)
 #define MUTE_LOOC		(1<<5)
+#define MUTE_SLOOC		(1<<6)
 #define MUTE_ALL		(~0)
 
 //Some constants for DB_Ban
--- a/code/modules/admin/admin.dm
+++ b/code/modules/admin/admin.dm
@@ -158,7 +158,8 @@
 		body += "<A href='?_src_=holder;[HrefToken()];mute=[M.ckey];mute_type=[MUTE_PRAY]'><font color='[(muted & MUTE_PRAY)?"red":"blue"]'>PRAY</font></a> | "
 		body += "<A href='?_src_=holder;[HrefToken()];mute=[M.ckey];mute_type=[MUTE_ADMINHELP]'><font color='[(muted & MUTE_ADMINHELP)?"red":"blue"]'>ADMINHELP</font></a> | "
 		body += "<A href='?_src_=holder;[HrefToken()];mute=[M.ckey];mute_type=[MUTE_DEADCHAT]'><font color='[(muted & MUTE_DEADCHAT)?"red":"blue"]'>DEADCHAT</font></a> | "
-		body += "<A href='?_src_=holder;[HrefToken()];mute=[M.ckey];mute_type=[MUTE_LOOC]'><font color='[(muted & MUTE_LOOC)?"red":"blue"]'>LOOC</font></a>\]"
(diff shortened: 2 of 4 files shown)


NotGoldenglow:
The ONLY issue I have with this is that, uhm. For better or worse, we've had prefbreak bans for necrophillia more than once which is not gucci. This prevents people from telling others off. I'd say keep SLOOC on after death, just not general LOOC.

YakiAttaki:
Nah, take out the bit about no LOOC while dead. That's needless; you should be allowed to communicate while dead because genuinely, people will dump you in the healing ward and then entirely forget about you. Then you can't communicate outside of SLOOC?
Bad, don't like. Take that part out.

Yeetus223:
everything is fine but the no looc when dead, both for the freaks and the people that wanna ask questions after a lethal fight so they can improve.

hugmeorelse:
> The ONLY issue I have with this is that, uhm. For better or worse, we've had prefbreak bans for necrophillia more than once which is not gucci. This prevents people from telling others off. I'd say keep SLOOC on after death, just not general LOOC.

SLOOC after death is enabled as noted in the PR.

hugmeorelse:
@Xander3359 Tested it out locally using some local workarounds to avoid having to interact with a DB.

CaydenT:
Is SLOOC range still really short? It may be worth increasing

PR #2371, opened 2026-08-10
Banditjak Part I

kimesik:
## About The Pull Request
A bunch of Bandit class changes.

- Changes to various skill levels. Most notable are:
- - Knave finally gets Legendary in Climbing (on par with Adventurer-Thief), Expert in Tracking and Athletics and Expert in crossbows, bows and slings regardless of their chosen subclass.
- - Rogue Mage gets Journeyman in Athletics (on par with Adventurer-Sorcerer-Spellslinger).
- - Pioneer gets Expert in Climbing and Swimming, because they are a siege engineer who's supposed to overcome obstacles first and create a safe path for their allies.
- Sellsword no longer has subclasses. They get Expert in almost all weapons and start with The Mercenary subclass's equipment. They are supposed to be a versatile medium armour fighter you can equip with just about anything, like MAA Footman.
- Iconoclast gets Minor Devotion regen, down from Major. Iconoclast never runs out of Devotion either way, but this change makes them consistent with Wretch-Heretic, which is good.
- Now all Bandit classes are correctly stated. They all get a weighted stat total of +10 by default, +13 if they have a visible bounty. There are no longer odd exceptions. Unless I missed some.

## Testing Evidence
image

## Why It's Good For The Game
More consistently balanced Bandits, that properly compare to Adventurer and Wretch equivalents.

## Changelog

:cl:
balance: Bandit classes skill adjustments (too many to list).
balance: Bandit classes stat rebalance (everyone is weighted +10 now, or +13 if visible bounty).
balance: Iconoclast devotion regen is Minor, down from Major.
balance: Bandit-Sellsword no longer has subclasses. They get Expert in almost every weapon type instead.
/:cl:

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/antag/Iconoclast.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/antag/Iconoclast.dm
@@ -15,7 +15,7 @@
 	subclass_stats = list(
 		STATKEY_STR = 2,
 		STATKEY_WIL = 2,
-		STATKEY_LCK = 1,
+		STATKEY_LCK = 2,
 		STATKEY_CON = 2, //We have a total of +10 in stats. +13 if we have a visible bounty.
 	)
 	subclass_skills = list(
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/antag/Iconoclast.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/antag/Iconoclast.dm
@@ -13,10 +13,10 @@
 		TRAIT_RITUALIST,
 		)
 	subclass_stats = list(
-		STATKEY_STR = 2,// LETS WRASSLE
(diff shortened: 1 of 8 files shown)


Pidgey-chan:
I think stat imbalances like rogue mage having a lower total or iconoclast being 1 or 2 less than hedge knight are fine. Magic and miracles are boons in a whole nother class from armor proficiency or weapon skill.

kimesik:
> I think stat imbalances like rogue mage having a lower total or iconoclast being 1 or 2 less than hedge knight are fine. Magic and miracles are boons in a whole nother class from armor proficiency or weapon skill.

While I agree that mages and miraclists should have lower stats than plain fighters, this isn't the case for any other faction or class (with the very notable exception of Wretch-Fighter-Warcaster), so I am sticking to +10/+13 for Bandits and not taking the responsibility for restatting a few dozens of magical and hybrid classes in this codebase.

PR #2376, opened 2026-08-10
Stops DT from counting tallies

IronDragoon:
## About The Pull Request

After my last PR had discussion I propose this. 

Tallies still count for RH and DW, however not for DT. I recommend we TRY this and see how things go

## Testing Evidence

Runs and works locally

image

## Why It's Good For The Game

Personally I think holding a vote then choosing the least voted option purely based on random other votes from players who may not even be playing at the time is silly

## Changelog

:cl:
del: Stops DT from counting in map tallies
/:cl:

### DIFF
--- a/code/controllers/subsystem/map_vote.dm
+++ b/code/controllers/subsystem/map_vote.dm
@@ -37,9 +37,16 @@ SUBSYSTEM_DEF(map_vote)
 /datum/controller/subsystem/map_vote/proc/write_cache()
 	rustg_file_write(json_encode(map_vote_cache), MAP_VOTE_CACHE_LOCATION)
 
+/datum/controller/subsystem/map_vote/proc/should_cache_map_tally(map_name)
+	return map_name != "Desert Town"
+
 /datum/controller/subsystem/map_vote/proc/sanitize_cache()
 	var/max = 200
 	for(var/map_name in map_vote_cache)
+		if(!should_cache_map_tally(map_name))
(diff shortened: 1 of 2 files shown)


zoktiik:
I don't know if we should be making the system not work at all for just this specific map, it's not a good engineering solution. I don't exactly like that I'm apologizing for Desert Town, but I find this very weird and needlessly snowflakey.

IronDragoon:
this was suggested as a compromise by other admins. The main problem with DT is that a few tally votes are adding up to then force it to be played by those who are not on. This means Dun world will still get lots of play when it often gets lots of votes, where DT usually only gets 5 or so per round.

zoktiik:
I understand that, yes. But a solution like this threatens to be made permanent without further action. If we are guaranteed further action that would remove this stopgap, then okay, sure, we can do it.

Pidgey-chan:
Better to just get rid of tallies or use the only counts for 1-2 rounds thing. Desert isn't my favorite but it isn't fair to make it the only one not included in the system

IronDragoon:
that is what the last PR I put up did...

CaydenT:
Can we atleast recognize that tallied votes from over four rounds (TWELVE HOURS!!) should fall off? Some small group of 7-8 from half a day ago that are voting for a map probavly aren't here anymore when the map gets forced.

coffeecoffeecoffee:
The other PR was much better.

PR #2378, opened 2026-08-11
Psydon miracles buff - Arbalist little buff

zBxnny:
## About The Pull Request

Psydon martial miracles from t3 and t4 _(absolver ones are NOT touched at all)_ are as of right now, rendered quite useless, I've asked inquisition and psydonite players and played a lot of it myself. And honestly I could never find a time where using these miracles would be useful. (Because they just pretty much fuck you up)

They're totally soulful, but balance wise, killing yourself in the middle of the fight or even before it starts is not ideal.

Did a little buff to the arbalist as well.

## Testing Evidence

Works on my end without issue. It's just value changes with some optimization on code to adjust blood by Artur's recommendation.

https://github.com/user-attachments/assets/a5d67820-0e25-4ec6-ba52-eee7771a83f9

## Why It's Good For The Game

In general, I've tweaked the brute damage you take and blood volume you give. This way you can play around with being psydon's most enduring soldier without instantly having to yield because you crippled your leg mid-fight with the press of a button. (You can oneshot your limbs even at full hp just by using some miracles as it is right now)
Good spells are still pretty damaging, but without actually crippling or rendering you useless as soon as the fight starts.

The arbalist as of right now feels better literally just running with a normal crossbow than using the sauterelle, this is due to how slow the reloading speed is, originally, we were supposed to be able to bless bolts to turn them into holy bolts, they would hit harder than they do now, but since that idea got abandoned, might as well just tweak it and make it reload just a tiny bit faster to make it usable in a fight.

## Changelog

Psydonic Retribution:
Blood volume -100 to -50.  An empowering hit that does more damage to you than the enemy even if you debuff them is never a good trade.

Psydonic Inspire: 
Blood volume -200 to -130. Half of your blood for a little buff that only lasts two minutes is never good, you'll die to a single bleeding in a minute or two.

Sacrosanctity:
Apply brute damage 200 to 100. The idea was good, but the damage trade is too much for just blood, 200 is usually enough to leave you bordering critpain depending on if you have 15 con or less.

Inviolability:
Apply brute damage 300 to 200. The damage enough is a lot and can cripple one limb instantly, lowering to 200.

Syonacrum:
Blood damage taken -400 to -200
Apply brute damage 300 to 250. At 300, it's enough to literally cripple limbs, which renders the miracle only useful if you have the absolver ready to take damage or in bed.

Sauterelle (Arbalist Unique Crossbow):
Reload speed from 8 seconds to 6.

> If anything feels too much, tell me so, my idea is to make these miracles *useful* again, right now they're just a meme to have in your bar and might even fuck you up if you click on them accidentally.


Charopos:
Good, but I'd recommend increasing the debuff time for Retribution as well. As it, it's still completely worthless as a debuff - at 30(!) seconds in exchange for a huge amount of bloodloss. This bloodloss, which lasts way longer and has a much more debilitating effect than what you stand to gain from it (-3 WIL, -1 CON to your foe) - it's not something that makes much a difference, even if you do use it more regularly w/ a lower cost.

wikimody:
Migt be out of scope, but can you give arbalest specyficly a recepture for making his heavy and special bolts? Currently they are unobtanium outside of new arbalest spawning in, you cant even buy them from inqusition vendor. if you use up all of your starting ones the crosbow becomes only a way to deliver your 1 lux boltbefore you have to swith to normal one

DogExtrordinaire:
I think it'd also be cool to give sacrosancity to all the psyheretics

PR #2384, opened 2026-08-11
Reworks Tally System to no longer be global tallies

RadiantFlash7:
## About The Pull Request
This PR makes it so the Tally system now tracks tallies per player, and adds it to their vote specfically. It prevents players from 3 rounds ago who aren't playing presently from influencing the current map.
TLDR: When person A votes for rockhill, and Person B votes for dunworld, Person B's next vote specifically for dunworld will be worth 2 votes. 
If dunworld is still not selected, worth 3 votes, and so on. When dunworld is selected, This vote value counter resets back to 1.
This extra value caps at making a vote worth 10 votes.

## Testing Evidence
it's admittedly somewhat hard to test in a solo test server, but so far everything seems to be functioning properly.

## Why It's Good For The Game
One of the main complaints about the tally system is how players who are likely not even playing are influencing the round map pick. This has the side effects of having players from old rounds over-emphasizing their vote when they are no longer present.
Switching it from a global tally, to a per person tally resolves this issue neatly.

## Changelog

:cl:
refactor: Tally Subsystem is no longer global, but per person's individual mapvote.

/:cl:


Pidgey-chan:
I still don't like tallies but this is a big improvement.

coffeecoffeecoffee:
Yeah, this is alright. I am wholly against the tallies system, but this is at least tolerable.

code/controllers/subsystem/map_vote.dm
```diff
@@ -1,58 +1,74 @@
...
-	/// admin override flag
+	// admin override flag
 	var/admin_override = FALSE
 
-	/// vote finalized flag
+	// vote finalized flag
 	var/already_voted = FALSE
 
-	/// selected map config
+	// selected map config
```
Xander3359, line 21: why are you getting rid of autodoc

code/controllers/subsystem/map_vote.dm
```diff
@@ -63,6 +79,47 @@ SUBSYSTEM_DEF(map_vote)
...
+		if(cfg.config_min_users && connected_players < cfg.config_min_users)
+			continue
+		if(cfg.config_max_users && connected_players > cfg.config_max_users)
+			continue
+
+		valid_maps += map_id
+
+	return valid_maps
+
+// Returns a player's currently banked bonus for a given map (0 if none).
```
Xander3359, line 116:
```suggestion
/// Returns a player's currently banked bonus for a given map (0 if none).
```

Xander3359 (approved):

CaydenT:
Oh thank god. This is genuinely awesome. Thank you for this. It makes it so legitimate victories for a map via voting are actually possible without letting people troll-vote half a day ago to get an undeserved map in there. This is peak!

GrayRachnid:
Can it just take my vote from the character menu preference instead of having to vote every single time? Some times I get caught up in the LOOC discussion when the round ends and forget to vote, having the tally only matter if you vote consecutively makes it... Important to vote.

Genevesh:
This is... pretty bad. That limits DT and Dun to be played maybe once every 5 Rockhill maps or so. Do we really need to go over why map rotation is really fucking important for both server's health and development of the other maps?

If you're going to listen to whiners because they bitch and moan like infantile fucking midgets, then I can call a bunch of people who DON'T do that, because we're all adults, to bitch and moan louder than them just to prove a point.

RadiantFlash7:
> This is... pretty bad. That limits DT and Dun to be played maybe once every 5 Rockhill maps or so. Do we really need to go over why map rotation is really fucking important for both server's health and development of the other maps?
> 
> 
> 
> If you're going to listen to whiners because they bitch and moan like infantile fucking midgets, then I can call a bunch of people who DON'T do that, because we're all adults, to bitch and moan louder than them just to prove a point.

First off- tone. Please don't insult other players/commenters.

Secondly,
I agree map rotation is important- but I also think people voting for it should be the ones to get it, rather than a few rounds later when they're offline. The current changes build up to a total of 10 tallies per player. In three rounds if you have 15 people voting for DT in a row, you'll end up with 45 votes for DT. Which, I think is a significant enough numbers bump.

PR #2396, opened 2026-08-12
Fix masterworks being unable to gain the polish buff. Polish_bonus and glint lost on full break rather than <25% integ.

Gug1234:
## About The Pull Request

- Polishing cream now checks for a polish_bonus (i.e., 3 force wielded, 2 force one-handed, 10% extra integrity) on the target item, making masterworks and blessed weapons that have the metal glint but not the bonus no longer block polishing attempts.
- Changed polish bonus and glint loss from 25% integrity to a full item break. Easier to intuit for players. Broken item = need to repair and repolish. 

## Testing Evidence

https://github.com/user-attachments/assets/e8377520-3880-465a-8f06-30f76c2984b5

## Why It's Good For The Game
Punishing someone by denying them their polish buff if the weapon is a masterwork is bad and feels unintended.

## Changelog

:cl: Yuckuza
fix: blessed silver and masterwork weapons that have a metal glint are now able to be polished with cream for the associated force and integrity buffs. No more avoiding making masterworks because it denies your squire from buffing the weapon.
balance: polish and glint loss now occurs when an item breaks rather than when below 25% integrity.
/:cl:

### DIFF
--- a/code/game/objects/items.dm
+++ b/code/game/objects/items.dm
@@ -1753,6 +1753,7 @@ GLOBAL_VAR_INIT(rpg_loot_items, FALSE)
 	return "<br><b><u>THERMAL RESISTANCE:</u></b><br>" + jointext(out, "<br>")
 
 /obj/item/obj_break(damage_flag)
+	lose_polish()//call to remove polish bonus on armor/weaps when broken. lives in /blacksmith/items.dm
 	..()
 
 	update_damaged_state()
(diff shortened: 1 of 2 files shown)


Elli-Skala:
I was always under the impression masterworks *were* suppose to get the polish buff (but it was bugged forever) as a incentive to be able to charge abit more for master works

Gug1234:
> I was always under the impression masterworks _were_ suppose to get the polish buff (but it was bugged forever) as a incentive to be able to charge abit more for master works

All masterwork does as apply polish 4 and the glint component which is what the polish cream checks to see if something has been polished. What it doesn't do is add the buffs using cream the normal way does. My change makes the masterwork glint cosmetic so you still need to use cream and the trait for your gamer buff.

PR #2401, opened 2026-08-13
Vote Tally Decay

IronDragoon:
## About The Pull Request

Changes the tally system, so that vote tally's only last for one round before they decay

## Testing Evidence

Run and tested locally

## Why It's Good For The Game

Had a lot of requests and discussion around this. Most players, whether fans of DT or not will surely agree that votes from upward of 9-12 hours ago should really not be having an effect on the current map vote. 

## Changelog

:cl:
balance: map vote tally's only persist for the next round before decaying
/:cl:

### DIFF
--- a/code/controllers/configuration/entries/general.dm
+++ b/code/controllers/configuration/entries/general.dm
@@ -538,7 +538,7 @@
 	min_val = 0
 	max_val = INFINITY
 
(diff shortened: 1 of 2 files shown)


IronDragoon:
did not see this PR, this is a better solution https://github.com/Rotwood-Vale/Ratwood-2.0/pull/2384

PR #2400, opened 2026-08-13
makes ztratocaster unbreakable on the ground

ceeetchell:
## About The Pull Request

one line change to remove the CAN_BE_HIT flag from the ztratocaster so people can't just walk up to you and break your guitar
## Testing Evidence
1 line + compiles

## Why It's Good For The Game
i don't even know why i put this flag on it, i think i was being sloppy with copying over the dreamwalker axe
ztratocaster is a 1 of a kind item that's destruction neuters the herald of progress, it shouldn't be able to be broken, like avantyne.

## Changelog

:cl: mushy
balance: Ztratocaster can now no longer be broken on the ground.

/:cl:

PR #2428, opened 2026-08-17
creates flag INDESTRUCTIBLE_ITEM and gives it to ztratocaster

ceeetchell:
## About The Pull Request
rectifying my fuck up on #2400 , i realized why it has CAN_HIT, cuz the repair ability relies on that same flag. awesome! oops!

by maintainer request instead of having the ztratocaster's indestructibility be an override of obj_destruction() unique to the ztratocaster, this pr introduces the global flag INDESTRUCTIBLE_ITEM that makes the item.. indestructible. it retains all other previous hit interactions.

also means that this flag can be used in the future for whatever, yay!

## Testing Evidence
compiles, tested it by beating the shit out my ztrato, didn't get destroyed but broke. didn't record it though

## Why It's Good For The Game
#2400 

## Changelog

:cl: mushy
fix: Fixes the ztratocaster being able to be broken (for real this time)
fix: Fixes the ztratocaster not being able to be repaired
/:cl:

### DIFF
--- a/code/__DEFINES/obj_flags.dm
+++ b/code/__DEFINES/obj_flags.dm
@@ -14,6 +14,7 @@
 #define BLOCK_Z_IN_DOWN			(1<<10) // Should this object block z falling from above?
 #define BLOCK_Z_IN_UP			(1<<11) // Should this object block z uprise from below?
 #define IGNORE_SINK				(1<<12)
+#define INDESTRUCTABLE_ITEM		(1<<13) // Item cannot be destroyed, allows for all other interactions
 
 // If you add new ones, be sure to add them to /obj/Initialize as well for complete mapping support
 
--- a/code/__DEFINES/obj_flags.dm
+++ b/code/__DEFINES/obj_flags.dm
@@ -14,7 +14,7 @@
 #define BLOCK_Z_IN_DOWN			(1<<10) // Should this object block z falling from above?
 #define BLOCK_Z_IN_UP			(1<<11) // Should this object block z uprise from below?
 #define IGNORE_SINK				(1<<12)
-#define INDESTRUCTABLE_ITEM		(1<<13) // Item cannot be destroyed, allows for all other interactions
+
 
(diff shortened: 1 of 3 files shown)


code/__DEFINES/obj_flags.dm
```diff
@@ -33,6 +33,7 @@
 #define SURGICAL_TOOL			(1<<12)	//Tool commonly used for surgery: won't attack targets in an active surgical operation on help intent (in case of mistakes)
 #define DREAM_ITEM				(1<<13) //Otherworldly items from Abyssor's dream. Tend to have special effects!
 #define FRESH_FOOD_ITEM			(1<<14) // Currently only used for fresh meat from butchering to differentiate it from stockpile magic meat.
+#define INDESTRUCTABLE_ITEM		(1<<13) // Item cannot be destroyed, allows for all other interactions
```
Xander3359:
```suggestion
#define PREVENTS_DESTRUCTION (1<<13) // Item cannot be destroyed, allows for all other interactions
```

Would be a better name, is more clear about what it does

Xander3359: This is also a pretty powerful trait because it also can't be tossed into acid or burnt away.

ceeetchell: name updated; i think it's fine for this case to avoid a situation where a summon zinstrument ritual is needed. it should be applied very sparingly to anything in the future

but, alternatively, acid+lava reduction can be restored by just moving the conditional below them, if needed

rintezukathearmlessartist-beep:
Had a report in game this might make it unable to be hit with a hammer to repair it, as a note.

Not sure if anything else has been merged recently that could be causing this behavior.

ceeetchell:
> Had a report in game this might make it unable to be hit with a hammer to repair it, as a note.
> 
> Not sure if anything else has been merged recently that could be causing this behavior.

if this is true i'm going to lose it

ceeetchell:
tested it and repairing should work, thank god

PR #2429, opened 2026-08-17
Cuffing Refactor, Overhaul, and Fixes with Improved Logging

hugmeorelse:
## About The Pull Request

This ended up being more than I intended. I just wanted to add logs for cuffing people. Then, in the process of testing logging, I come across another bug. Then, as I fix that bug and make sure it's all good, I come across two more bugs. This is so stupid. 

For example, leg cuffs had several different calls, each one individually needing slowdown, alert, and overlay being implemented each time along with writing the log. Literal psychopath code.

**Anyways, there are probably edge cases that I didn't cover. So, please, let me know if something breaks horribly. Better to fix now rather than later. My mind went numb at a certain point.**

**Main Change**
- All leg restraint handling now goes through one piece of code (`set_legcuffed`). Nothing else touches the leg slot directly.
- Restraints now simply state how much they slow you instead of each one manually switching the effect on and off in two different places and hoping they match. Some people are masochists I guess.
- The "restrained" alert used to live in the human sprite code, which is why only humans ever got one. It's been moved somewhere every creachure can reach.
- Deleted a bunch of unnecessary code. 
- Struggling out of restraints, and giving up, are now logged.

**Bugs Fixed**
- Burn or delete a rope tied to someone's legs and they stayed slowed **forever**. This SHOULD stop happening now, but I couldn't be fucked to test out every edge case.
- Slipping out of leg restraints threw a runtime error in theory.
- Dropping a spare rope from your hands removed the slowdown from a *different* rope tied to your legs, so you could walk normally while still visibly tied up. ???
- Leg restraints put on by an admin, or through the strip panel, did nothing at all. No slowdown, no alert.
- Non-human characters never got the alert, so it wasn't easy for them to struggle out of it.
- If you got netted, shook the net off, then got tied with a rope, the net's timer would expire half a minute later and silently untie the rope.
- The admin log didn't read correctly
- Nets, the weird Graggar net spell, admin equips and every single removal wrote nothing to the log at all.
- You can (in theory) no longer start tying someone up and then walk away. Both arm and leg tying now check you're still next to each other. **By the way, I couldn't test this one locally. So, this will have to be tested live.**

**Overhaul**
- Strength now scales all the way to 20 instead of maxing out at 15. Remember when only a legendary beggar roll or maniac could break chains? Well, yeah, let's have insta-chain breaks be actually special. 
- Chains take roughly two minutes at average strength, one minute at 15, twelve seconds at 19, and only break instantly at 20.
- Untying your own legs when your hands are free now takes about four seconds, the same speed someone else could untie you. You're not stupid. You know how to use your hands.
- Pressing resist again now gives up on struggling, instead of telling you that you're already struggling. Bystanders see you stop.
- Giving up is instant. It used to be blocked for ten seconds by the cooldown that the struggle itself had charged.
- Getting free no longer leaves you frozen and unable to act. That cooldown was charged when you *started*, so anyone escaping quickly came out free but couldn't use their hands for a while.
- If something went wrong mid-struggle, a restraint could get permanently stuck as un-resistable. It now recovers on its own in theory. Again, I'm not testing every edge case.

**P.S.**

The whole telekinetic chaining problem isn't just with chains. The problem seems to rest with the do_mob shit. Apparently, RT characters have latent telekinetic powers. Who knew? Way out of scope for this PR.

## Testing Evidence

It compiled and worked for most things that I tested, but I honestly got tired of hopping on and off to test out one bug fix, only to find another bug.

## Why It's Good For The Game

Read above.

## Changelog
cl:
fix: Leg restraints no longer leave you slowed forever after the rope is destroyed somehow
fix: Slipping out of leg restraints no longer throws a runtime (in theory)
fix: Dropping a rope from your hands no longer removes the slowdown from a different rope tied to your legs.
fix: An expiring net no longer unties whatever restraint you are wearing at the time.
fix: Non-human creachures can now see and click the restrained alert to struggle free.
fix: Leg restraints applied by admins or through the strip panel now actually work.
fix: You can no longer start tying someone up, walk away, and still have it work.
balance: You now need 20 strength to break out of restraints instantly instead of 15. 
balance: Escapes can be interrupted by stuns and movement again at every strength below 20.
balance: Untying your own legs with your hands free is now as fast as someone else untying them
qol: Pressing resist again now gives up on struggling out of restraints
qol: Getting free of restraints no longer leaves you unable to use your hands for several seconds after
admin: Leg restraints now log every time they go on or come off, including nets, spells and admin equips. Thrown nets record who threw them (in theory).
admin: Struggling out of restraints, and giving up on it, are now logged.
refactor: Leg restraint handling is now in one place instead of spread across several.
/:cl:

### DIFF
--- a/code/datums/status_effects/rogue/debuff.dm
+++ b/code/datums/status_effects/rogue/debuff.dm
@@ -179,19 +179,6 @@
 	effectedstats = list(STATKEY_SPD = -5, STATKEY_WIL = -2)
 	duration = 30 SECONDS
 
-/datum/status_effect/debuff/netted/on_apply()
-		. = ..()
-		var/mob/living/carbon/C = owner
-		C.add_movespeed_modifier(MOVESPEED_ID_NET_SLOWDOWN, multiplicative_slowdown = 3)
-
-/datum/status_effect/debuff/netted/on_remove()
-	. = ..()
-	if(iscarbon(owner))
-		var/mob/living/carbon/C = owner
-		C.legcuffed = null
-		C.update_inv_legcuffed()
-		C.remove_movespeed_modifier(MOVESPEED_ID_NET_SLOWDOWN)
-
 /atom/movable/screen/alert/status_effect/debuff/sleepytime
 	name = "Tired"
 	desc = "I should get some rest."
--- a/code/game/objects/items.dm
+++ b/code/game/objects/items.dm
@@ -89,8 +89,9 @@ GLOBAL_VAR_INIT(rpg_loot_items, FALSE)
 	var/edelay_type = 1 //if 1, can be moving while equipping (for helmets etc)
 	var/equip_delay_other = 20 //In deciseconds, how long an item takes to put on another person
 	var/strip_delay = 40 //In deciseconds, how long an item takes to remove from another person
-	var/breakouttime = 0 // greater than 15 str get this isnstead
+	var/breakouttime = 0 // str 20 breaks out on this instead of struggling for slipouttime
 	var/slipouttime = 0
+	var/legcuff_slowdown = 0 //movespeed slowdown while worn as legcuffs, 0 = none
 
 	var/list/attack_verb //Used in attackby() to say how something was attacked "[x] has been [z.attack_verb] by [y] with [z]"
 	var/list/species_exception = null	// list() of species types, if a species cannot put items in a certain slot, but species type is in list, it will be able to wear that item
--- a/code/game/objects/items/rogueitems/ropechainbola/net.dm
+++ b/code/game/objects/items/rogueitems/ropechainbola/net.dm
@@ -9,6 +9,7 @@
 	w_class = WEIGHT_CLASS_SMALL
 	icon_state = "net"
 	slipouttime = 2 SECONDS //ideally you're using this to catch a dodger, not in the middle of combat
+	legcuff_slowdown = 3
 	gender = NEUTER
 	throw_speed = 2
 	var/knockdown = 0
@@ -24,9 +25,7 @@
 	if(iscarbon(loc))
 		var/mob/living/carbon/M = loc
 		if(M.legcuffed == src)
-			M.legcuffed = null
-			M.remove_movespeed_modifier(MOVESPEED_ID_NET_SLOWDOWN, TRUE)
-			M.update_inv_legcuffed()
+			M.set_legcuffed(null)
 			if(M.has_status_effect(/datum/status_effect/debuff/netted))
 				M.remove_status_effect(/datum/status_effect/debuff/netted)
 		forceMove(M.loc)
@@ -39,17 +38,16 @@
 /obj/item/net/throw_impact(atom/hit_atom, datum/thrownthing/throwingdatum)
 	if(..() || !iscarbon(hit_atom))//if it gets caught or the target can't be cuffed,
 		return//abort
-	ensnare(hit_atom)
+	ensnare(hit_atom, throwingdatum?.thrower)
 	// Nets always fall off after 10 seconds resist or not, so that the advantage it brings you is limited
 	// Being hit by a net and instalossing isn't fun for anyone because removing can be interrupted
 	addtimer(CALLBACK(src, PROC_REF(remove_effect)), 10 SECONDS, TIMER_OVERRIDE|TIMER_UNIQUE)
 
-/obj/item/net/proc/ensnare(mob/living/carbon/C)
+/obj/item/net/proc/ensnare(mob/living/carbon/C, mob/user)
 	if(!C.legcuffed && C.get_num_legs(FALSE) >= 2)
 		visible_message("<span class='danger'>\The [src] ensnares [C]!</span>")
-		C.legcuffed = src
 		forceMove(C)
-		C.update_inv_legcuffed()
+		C.set_legcuffed(src, user)
 		SSblackbox.record_feedback("tally", "handcuffs", 1, type)
 		to_chat(C, "<span class='danger'>\The [src] entraps you!</span>")
 		C.Knockdown(knockdown)
--- a/code/game/objects/items/rogueitems/ropechainbola/rope.dm
+++ b/code/game/objects/items/rogueitems/ropechainbola/rope.dm
@@ -11,6 +11,7 @@
 	throw_range = 3
 	breakouttime = 5 SECONDS
 	slipouttime = 1 MINUTES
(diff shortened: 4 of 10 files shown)


Plum-but-gayer:
All very good fixes but not very fond of getting rid of chainsnapping being mixed in all this. 

Having 15 go from instant to a full minute (effectively useless) is a bummer. Chainsnapping is already pretty rare and pretty easy to beat (simply kill the person who chainsnaps) and serves as counterplay for people who abuse chains in combat instead of an end-of-combat restraint tool. 20 is also truly rare for an ability that, while neat, has extremely situational usefulness.
 
 Otherwise great stuff.

X404ErrorX:
Maybe 20 for insta-break is too much, but yeah 15 STR can be reach by most of the normal roles, maybe 18 could be an good stat for chainsnapping, to be honest high STR its really eazy to get right now, so 20 its not that bad anyways.

Chainsnapping at 15 STR was an mechanic of the days when XII was really high str stat, now even levys can get higher numbers.

hugmeorelse:
> Maybe 20 for insta-break is too much, but yeah 15 STR can be reach by most of the normal roles, maybe 18 could be an good stat for chainsnapping, to be honest high STR its really eazy to get right now, so 20 its not that bad anyways.
> 
> Chainsnapping at 15 STR was an mechanic of the days when XII was really high str stat, now even levys can get higher numbers.

I'll probably lower it to 18 maybe. We'll see.

RasbeRissa:
Let me preface this annoying wall of text by saying that the bugfixes are phenomenal - this solely is aimed at changing the chain snap threshold, which in my personal opinion should probably be atomized - it's a balance adjustment tucked into a balance and logging PR.

Getting 20 Strength is pretty unheard of for the most part, especially since with strength softcap on force scaling being 15 there isn't much of a point unless you're one of the few types who can get Strength Unbound - and declare yourself instantly valid by doing so. Even 15 strength isn't as easy as people are making it out to be for the average vale dweller - getting 15 strength requires you to be one of the classes that gets +2-3 Strength, pick a statpack that gives +1-2 Strength, and also possibly be a race that gives you +1 Strength, depending on if you compromised elsewhere. Bloodloss also rapidly guts your strength.

So the people who could chainsnap reliably (+3 from Class, enabling 15 with just statpack or pack+race) were:
ADVENTURERS/MIGRANTS:
Amazon - Plate-kini Subclass
Barbarians
The Czwarteki Hussar
Gronnic Itinerants
BANDITS:
Brigand
Iconoclast - Max Bounty
CHURCH - IN THE CHURCH SPECIFICALLY:
Templar
Martyr
GARRISON:
City Guard - Maul Subclass
Heavy Knight
Retainer - Henchman Subclass (side note i have no idea what this class is I've been gone a week)
MERCENARIES:
Atgeveri Shaman
Black Oak - Guardian Subclass (15 Strength Capout - Must take Muscular to reach 15 due to racial restriction)
Fjall Járnklæddur, or as code refers to them: Gronn Heavy.
Hired Thug
Oathmarked - Drakian Exclusive. Starts at 14 Strength as a result.
TOWNERS:
Thug - Big-man Subclass
WRETCHES - I am not going to factor Omnistat at this time:
Ancient Champion
Berserker
Lunacy Embracer

And that's our classes. Now, admittedly, there are a LOT of classes that give +2 Strength, and wretches specifically are able to get +1 Omnistat by going with a full recognized bounty, which makes their list specifically much larger. I'm not going to list them all. This comment is already a small wall with just the above list. However, the above list is a list of classes that can reliably hit the 15 Strength to chain snap. All classes with +2 Strength instead of +3 would need to get +3 Total Strength to hit 15 - and that means taking the Muscular Statpack (or getting a lucky Fated Roll) and being a race with +1 Strength as a racial, like the Half Orcs, the Drakians, or an Aasimar taking Strength.

This also doesn't take into account potential buffs, of course. Sure. People COULD buy a strength potion, or get a mage to help them. But these are situational circumstances and probably shouldn't be considered when weighing "Should chain snapping require more strength than it already does?" due to various other factors such as how long said buff lasts or in the case of Strength Unbound, making you very much the focus fire target of everyone ON TOP of making people decide to kill you for chain snapping.

Another thing, and this is a big one: Bloodloss. It is EXTREMELY rare to chain someone who is healthy. You usually only chain them after they've lost a fight, which usually means they've lost a good amount of blood. Which means the bloodloss debuff, which means they have less strength already - and by making it 20 strength, also means realistically speaking we'll never see a chain break again.

TL;DR

Despite the sheer size of this comment, I am actually not taking a stance for or against changing the strength requirement for chain breaking beyond my opinion that making it 20 is unreasonable. I am, however, suggesting that this be an atomized change rather than lumped into this PR.

Gug1234:
When I read "strength scales up to 20" I was worried you uncapped the damage scaling. I like this change now that I understand you scaled chain snapping up to 20.

I was looking and didn't see anything specific but did you touch the pillory breaking strength check as well?

usernameasterisk:
refactor pr

look inside

stealthy nerf

image

Danyak3:
I Believe it needs Rescaling, I Get that chain-breaks were abit too easy back in the day and you want them to be special but now it sort of screws over strength greatly.

I Believe that still 20 Being a instant chain-break is fine, But I believe the Threshold for instant chain breaks threshold should be 19 at minimum.  18-17 Should be a Minute that breaks down Chain Durability. (Only Strength 17-18 can affect chain durability.)  This would give a Bone to those who can't get their Strength at that high degree but still gives 17-18 strength characters a bone and a reason to run that much strength as opposed to not even bothering and focusing on perception and Intelligence/Willpower which is King for Fighting. It would also make sense Aswell because someone who has a strength of 17-18 could not be allowed to just walk unescorted by someone who is only mildly weaker than them it would make sense narratively. It would even be awesome narratively that for those that are seasoned warriors (Combat Aware Guards/Wardens) should get a Tooltip about that one person that was just chained to basically say that it would be a good idea to have 2 people to keep that chained Strength behemoth restrained.....OR HE (or she or whatever) WILL Break out eventually.

As it stands right now due to how binary chains are, you either have the strength to break out of them or your just done for.... Adding Chain Durability for those that have strength 17-18 (Basically picking the Struggler Stat block) would make things less sucky for those that haven't hit the Peak but still built for it and make having that much strength at the Cost of Perception fair....and feel nice. 

Or maybe just have Strugglers be allowed to wear down chains with regular resisting then if that's simpler.... but something for those that have monstrous Strength, but not peak Strength would be nice.

PR #2430, opened 2026-08-17
Toiler Adjustments for RW

AllCorruptingWyrm:
## About The Pull Request

Flatly: Toiler has morphed into something I'm not happy with. This is the safe option before the nuclear option. The intent behind Toiler was that it was supposed to be a wretch role that acts as an evil mastermind for the other wretches, but it swiftly has been used as a frontlining position, and that isn't what I'm OK with.

This PR shelves some stats and removes the Inhumen Buff miracle for that reason, as it also detracts away from the Willworking/Miracle work of Graggarites and it just doesn't fit.

This PR will also set toiler slots to 0

Saying that, the alternative will be making this role like AP's Munitioneer.

## Testing Evidence

n/a

## Why It's Good For The Game

This role has become something I never wanted it to be.

## Changelog

:cl:
add: Added more things
del: Removed old things
qol: made something easier to use
balance: rebalanced something
fix: fixed a few things
map: added/modified/removed map content
spellcheck: fixed a few typos
code: changed some code
refactor: refactored some code
admin: messed with admin stuff
server: something server ops should know
/:cl:

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/wretch/wretchedtoiler.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/wretch/wretchedtoiler.dm
@@ -9,7 +9,7 @@
 	cmode_music = 'sound/music/cmode/antag/combat_thrall.ogg' //We don't have the original .ogg and I'm lazy. You get thrall music now. Change it if you want.
 	class_select_category = CLASS_CAT_TRADER
 	extra_context = "Choose between 2 options: being an EVIL mastermind or a WRETCHED servant" //choose between master and servant
-	maximum_possible_slots = 1 
+	maximum_possible_slots = 0
 
 	// balance isn't real i picked these bc they're funny
 	subclass_stats = list(
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/wretch/wretchedtoiler.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/wretch/wretchedtoiler.dm
@@ -13,8 +13,9 @@
 
 	// balance isn't real i picked these bc they're funny
 	subclass_stats = list(
-		STATKEY_STR = -1, //YOU ARE WRETCHED!!!
-		STATKEY_CON = -2, //AND YOU WILL +TOIL+!!!!!!!!!!!
+		STATKEY_STR = -4, //YOU ARE WRETCHED!!!
+		STATKEY_CON = -4, //AND YOU WILL +TOIL+!!!!!!!!!!!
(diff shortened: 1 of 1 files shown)


Venithefox:
I think these are good changes so far. this is just a battle of balance philosophy though. 

so you, tzula want to make it so toiler is a pure support. which is what I personally agree with. and a support antag is a good thing regardless of "the direction" of wretch. wretch is a free-form role and after playing both mastermind and servant toiler extensively I think I can give some feedback. but first.

tzula: support antag good, should give up combat ability for making others more powerful
radiantflash: thinks antags should be able to fight on their own which toiler can already do. and I also agree with that though not to the same extent.

mastermind plays more traditionally like any other antagonist role. though I am of the opinion it's more of a direct upgrade to rogue mage? like yes you trade T3 for T2 but you can also heal yourself and give yourself whatever other bonuses of your patrons miracles which is more than enough to make up for the loss of AOE spells. not to mention it's a rogue mage that can make themselves better armour, repair said armour and starts with a scrying orb.

servant doesn't have many issues, it is afterall what a toiler is supposed to be. one who toils. yes legendary mining enchanted pickaxe is funny but it's...not the best. massive swing delay makes it so the only time this will ever be useful for combat is if someone is grabbing. even with swift+haste.

my suggestion is having toiler be separate from wretch. yes they're in the wretch grotto but toilers aren't really held up to the same standard of antagonism as any other wretch. your purpose does conflict with the philosophy of "you must antagonize." I personally disagree that a toiler making things for wretches ISN'T antagonism. because you are directly being a force multiplier making a difference whether through buffs or equipment that turns tides of battles. but considering toiler isn't a "traditional" antagonist, I believe it should be its own role like gnolls for example. 

another suggestion, the first one I actually made when toiler was first added actually. was incurring a MASSIVE speed and con penalty. toiler needs stamina for buffs so I don't think WIL should be reduced. however they are a SUPPORT and being weak and unable to escape combat makes them focus a lot more on being defensive and supportive. the entire point of the role. sure toilers can still frontline if they had like -5 con and -3 speed or something but you need to weigh that commitment because you aren't getting out if you do. alternatively give mastermind -5 speed and give servant -5 con. a mastermind is a calm, calculated controller of the battlefield and as such they should be surrounded by those who will protect them. whereas a servant is (in my head canon at least, the larp is real.) a slave, perhaps a normal servant. and as such toil so much as to destroy their own bodies constitution. both mastermind and toiler DO need their willpower and in fact probably could do with MORE willpower because stamming out using all your abilities is annoying.

kimesik:
0 slots means the class isn't selectable AT ALL
If you want UNLIMITED slotcap, you put -1 instead.

PR #2432, opened 2026-08-17
Weep can now refill blood, even without wounds

Shadow-Quill:
## About The Pull Request

Absolver's weep (and, by extension, the Stigmata's) can now be used on an unwounded target to refill their blood.

## Testing Evidence

Test instance where a human was injured via the health panel, with weep used to refill blood, wound, then both at once.
When used on someone who needed neither blood nor had wounds, the cast was reverted.

## Why It's Good For The Game

Weep is strange in that it *has* to be used when someone has a wound to refill their blood.
This makes it slightly more versatile, while still having the same cost. (Of losing your own blood.)

## Changelog

:cl:
balance: Absolver's WEEP can now be used to refill the blood of someone who does not have wounds.
/:cl:

### DIFF
--- a/code/modules/spells/roguetown/oldgod.dm
+++ b/code/modules/spells/roguetown/oldgod.dm
@@ -38,44 +38,53 @@
 		revert_cast()
 		return FALSE
 
+	// To allow the spell if the target needs blood, but has no wounds, we need these.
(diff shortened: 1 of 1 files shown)


Shadow-Quill:
superseded by #2472

PR #2434, opened 2026-08-18
Moves devotion verbs to IC tab category

Shadow-Quill:
## About The Pull Request

The "Check Devotion" and "Give Prayer" verbs are the only ones that are given to the Cleric verb tab. I'm moving it to IC, because it doesn't make overmuch sense to give an category *just* for these two. Also, does anyone use the verbs?

This only affects where the verbs show up in the list; hud interaction (ie clicking your devotion bar) is unchanged.

## Why It's Good For The Game

Less categories, I guess?

## Changelog

:cl:
del: The "Check Devotion" and "Give Prayer" verbs have been moved to the IC tab.
/:cl:

### DIFF
--- a/code/controllers/subsystem/rogue/miscprocs.dm
+++ b/code/controllers/subsystem/rogue/miscprocs.dm
@@ -229,7 +229,7 @@
 
 /mob/living/carbon/human/proc/devotionreport()
 	set name = "Check Devotion"
-	set category = "Cleric"
+	set category = "IC"
 
 	if(!devotion)
 		return FALSE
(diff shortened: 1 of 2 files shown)


wikimody:
Give prayer is nesesary for any class with actualy runs out of devotion, so anyone but acolite. Having them in separate tabs helps have quick acess to them instead of trying to find them though multiple options

Shadow-Quill:
> quick acess

Left clicking the devotion bar (or middle-clicking a worn cross) is a shortcut to start the praying cycle.

zBxnny:
Good change, as staff that has 25 categories, the less I have, the better.

PR #2443, opened 2026-08-18
Bush & Grass fixing

NPC1314:
## About The Pull Request

QoL focus around grass/bush/thornbush/weepwine/kneestingers

Made the mistake of local testing. Found many things abitrary or wrong with grass & bush objects and now I got to fix them.
Note for future: There are 2 vars that decide how things layer. PLANE is the superior level, LAYER the lower.
Changing the LAYER does little to fix issues if its due to the offending object having the "wrong" PLANE.

* Tall bushes and thorn bushes were once move blockers for a reason. Can´t be arsed to restore them since it would require remapping probably, but at least thorn bushes got mechanics added so its not just broken and a weird decal.
* Bushes no longer 0 impediment when running east-west and full move block north-south. Instead can move in any direction through it, but you get stuck for a little bit in the bush. EDIT: Currently moved to a move-delay  thing, since immobilize is also baked into stuns and rider knockoffs and whatnot. Low delay, STR dependent.
* Thorn bushes work as above, but also if not kneestinger immune you risk getting a thorn scratch (likely) or impale(10% chance). Curse of Dendor makes this effect worse, and disables kneestinger immune effect.
* Grass now layer under objects and mobs properly. No more fighting invisible badgers in grass. Less PITA to pick up that club you lost during training on the dummy.
* Kneestingers layer a little bit better, was under too many things.
* Rustle of grass crossing slightly less sharp. Was annoyingly loud.
* Weeping vines were supposed to be spawned by evil trees, but its mapped all over. Currently just a decal that can add a little damage. Added a less dramatic slowdown effect like the bushes, but ignored by KNEESTINGER IMMUNE. Makes them a little better for druids ofc, but makes them make more sense. CURSE OF DENDOR makes them slow you more.
* Reeds got 4 new variations of sprites so now 8 total, all very close to the originals. The slight edits make them conceal mobs behind them a tiny bit less, while also improving the ease of harvesting them.
* Racoon and Mucrab mobs and subtypes now get a small mob size instead of human sized. Also now allow moving over them. Smaller animal sizes affect the negative effects of the thorns etc a bit now. Raccons keep density since default hostile and throwing weapons should work on them. Shame the code ties thrown weapons together with being a tileblocker but it is what it is.
* A type of bush inherited wrong integrity, fixed.
* During testing I accidentally stepped on glasses. Lord have mercy. Added a non-earzape sound for it instead, tiny reduction on loudness for the normal glass smash, still ear-pain loud, just a tiny bit less so. Also removed pointless duplication in inheritance for the spectacles sounds.

https://github.com/user-attachments/assets/954b2508-c73e-45e7-9927-da3028608339

## Testing Evidence

Tested locally, will obviously need a couple days TM to ensure theres no unexpected issues, particulary since the layering standards seem messy theres bound to be something unaccounted for. Most normal interactions tested for naturally.
Pr1

https://github.com/user-attachments/assets/fd0cbe8d-48c4-41ed-9db1-e6a7f6105e4b

## Why It's Good For The Game

Slowdown for bushes instead of them working like weird railings makes it less cancer to move about, and makes more sense.
The layering things are pure QoL, as is the minor weed edit.
Making thorn bushes especially, and partly weepwines not be just decals is good
Overall fixes various minor issues and makes forest walking a little less shitty.

## Changelog

:cl:
add: Replaced old way bushes impact mobility to a slowdown lessened by STR
add: Thornbushes got mechanics now, work like above with extra thorns even if walking. LCK helps against getting stung.
add: Weepwine added a weak slowdown lessened by STR
add: Grass and Kneestinger objects layering fixed
code: Integrated Curse of Dendor and Kneestinger immunity better into the gameplay with Thornbushes and Weepwine
sound: slightly less loud grass crossing
sound: spectacles new breaksound. Tiny reduction in glass shatter volume.
image: minor modifications to weeds for QoL
fix: mudcrabs and subtypes got correct animal size and density removed so you dont get bodyblocked by a frog
fix: one sidetype of bush had crazy 150 integrity (thorn bush got 35 for reference) due to incorrect inheritance, set to 30
/:cl:


code/game/objects/structures/roguetown/rogueflora.dm
```diff
@@ -565,8 +574,8 @@
 /obj/structure/flora/roguegrass/bush/CanPass(atom/movable/mover, turf/target)
 	if(istype(mover) && (mover.pass_flags & PASSGRILLE))
 		return 1
-	if(get_dir(loc, target) == dir)
-		return 0
+//	if(get_dir(loc, target) == dir)	disabled to allow movement of mobs through bushes
+//		return 0
```
noelle-lavenza: we have version control, don't comment this out, just remove the entire override. it will never return FALSE so there's no need to override it at all

code/game/objects/structures/roguetown/rogueflora.dm
```diff
@@ -763,7 +772,38 @@
...
+				if(!HAS_TRAIT(src, TRAIT_PIERCEIMMUNE))
+					var/obj/item/bodypart/BP = pick(H.bodyparts)
+					var/obj/item/natural/thorn/TH = new(src.loc)
+					BP.add_embedded_object(TH, silent = TRUE)
+					BP.receive_damage(10)
+					to_chat(H, span_danger("\A [TH] impales my [BP.name]!"))
+			else
+				var/obj/item/bodypart/BP = pick(H.bodyparts)
+				to_chat(H, span_warning("A thorn [pick("slices","cuts","nicks")] my [BP.name]."))
+				BP.receive_damage(10)
```
noelle-lavenza: please use early returns here, e.g.
```dm
	if(!isliving(AM))
		return
```
and similar

code/modules/events/spacevine.dm
```diff
@@ -297,13 +297,21 @@
...
-		var/mob/living/M = crosser
 		M.adjustBruteLoss(5)
 		to_chat(M, "<span class='warning'>I nick myself on the thorny vines.</span>")
+	if(istype(crosser))
+		if(HAS_TRAIT(crosser, TRAIT_CURSE_DENDOR))// cursed by Dendor means longer tangle. Sucks.
+			M.Immobilize((36)-M.STASTR)
+			M.adjustBruteLoss(5)
+			to_chat(M, "<span class='warning'>I nick myself on the thorny vines.</span>")
+		else if(!isvineimmune(crosser))
+			M.Immobilize((18)-M.STASTR)
```
noelle-lavenza: we already checked `istype(crosser)` above, please pull that out
also, this should never run for non-living mobs anyway, so you should do an early `if(!isliving(crosser))` return IMO. also if you move the immobilization time into a variable you can reduce the number of duplicated lines.

plus, as written, this means someone who's cursed has a 23% chance of getting "I nick myself on the thorny vines." twice, which i don't think is intentional

code/game/objects/structures/roguetown/rogueflora.dm
```diff
@@ -472,8 +478,9 @@
 	max_integrity = 100
 	destroy_sound = "plantcross"
 	climbable = FALSE
-	dir = SOUTH
+	dir = NONE // Impassable in this direciton
```
noelle-lavenza: i don't think this is correct, direction is often used for visual stuff. also, you removed bushes blocking movement in CanPass anyway, so this comment makes no sense

noelle-lavenza (changes requested):

code/game/objects/structures/roguetown/rogueflora.dm
NPC1314: True, remnant of first draft, removed.

code/modules/events/spacevine.dm
NPC1314: Had to refactor the proc to do that, sorted.

code/game/objects/structures/roguetown/rogueflora.dm
NPC1314: Ehh it was this that needed to be redone not he one below, point is, sorted.

code/game/objects/structures/roguetown/rogueflora.dm
NPC1314: deleted

NPC1314:
Also added a fix for a bush integrity and a couple small critter size adjustments, see edited description.

Skyglazer:
The bush immobilization seems a little troublesome. I've been playing high-STR characters for the majority of the time it's spent TMed so far, so it's actually personally affected me quite little- but I'm hearing from friends who play low-STR characters that it's an absolute pain- might be worth considering reducing it a bit, or letting the formula take SPD into account as well- or perhaps whichever's highest between your SPD and STR, or something like that?

SneakySlinky:
The PR describes it as 'slowdown' but an outright immobilize (and around 2-3s if you're not pumping STR at that) is anything but; 

We have ways to enforce slower movement moving from turf to turf (such as how stuff like moving around on dirt/mud works, and where stuff like longstrider benefits)

Perhaps large bushes should also fall into that sort of movement-affecting behavior, if so desired?... Adding onto the movement slow unless you have means to reduce/ignore it

CameronWoof:
The bush stuff is pretty miserable for gameplay, everything else is fine.

CameronWoof:
image

This PR also seems to make flowers not layer above other foliage, which makes them look... sad.

DorfGuy1:
The changes to racoons makes you unable to hit them with thrown objects. Take this bug report seriously. don't make me use d*scord

DorfGuy1:
Also the bushs interaction with mounts is PURE ass

thevaliantknife:
Please change the bush mechanic. It provides nothing but an annoyance, and causes people to funnel even harder. A slowdown is fine. A hard stun is not.

P-roxii:
Yeah, the full stun is just really annoying. There are so many bushes and you are going to end up having to walk through them all the time. Pretty please revert it or just make it a slow down, such as how those without longstrider are slowed down running through the bog.

NPC1314:
Immobilize proc obviously got too many drawbacks based on the testing,  I'll try something else in a bit and see if it does the trick

envoyin:
I truly wonder the point of the bush stun.

NPC1314:
Changes::
Switched the immobilize to a movedelay thing instead for normal bushes, less annoying.
Raccoons huge again so they can be javelined. For default hostile creatures its the least worst tradeoff.
Herbs +0.01 layering so they still purdy.

fisherfridrich:
I have to agree with a lot of the people here that the stun is atrocious... If you don't play a str build, you get stuned for way too long. This also makes the Dryad with its weep wines a menace, as it stuns you for a long while as a mage and then rushes in and beats you up with nearly no way of fleeing anymore. Dryads are not meant to wipe a solo mage once the vines hit you, but rather punish you through bleed and nicks from the vines for moving. Please change this to a slowdown, I beg of you.

envoyin:
why has the bush stun returned.

DorfGuy1:
> I have to agree with a lot of the people here that the stun is atrocious... If you don't play a str build, you get stuned for way too long. This also makes the Dryad with its weep wines a menace, as it stuns you for a long while as a mage and then rushes in and beats you up with nearly no way of fleeing anymore. Dryads are not meant to wipe a solo mage once the vines hit you, but rather punish you through bleed and nicks from the vines for moving. Please change this to a slowdown, I beg of you.

I think maybe weep vines should get the bush stuns.

PR #2452, opened 2026-08-19
Rockhill inquisition manor fortification

MickDick1:
## About The Pull Request

The following are the inquisition manor now after changes:

image
image
image
image

## Why It's Good For The Game

If you've played Inquisition, you already know what this is about.
The manor has 50 different low effort break-in spots and no alarms and -and-and-and...
Basically you cannot leave the manor ever or else some chucklefuck rushes to steal the relics they have and a key typically in less than 2 minutes the millisecond the inquisition team leaves.

It is in my personal opinion that the inquisition should have one of the more locked down locations on the map that isn't a separate Z level hidey hole. Of all the 'factions'. It should be them. They're a team of at most 5 guys assuming no randoms show up and sign on. They are expected to move as a team. They cannot visit the manor every 2 minutes in the game... I'd even give them an airship if I had the chops and time right now. But that will be a project for later, possibly.

Right now, the changes as they are made. Simply add more time to breaking in. I'm sure someone will find the most meta-efficient way to break in even with the padding. But at least it won't be a 2 minute smash and grab anymore. And a silverlining to the not-unlikely-scenario where an intruder mines... they leave behind material that can be used to rebuild. So. There's that.

Anyways. Literally every time this happens every member of the inquisition groans IRL. It's high-key annoying as shit that this happens **every single round** and there's basically nothing you can do about it.

Only rockhill for now. If it's accepted I can see about unshittifying some of the other map's manors but I need more experience with them first. Rockhill is the only one I've really played on.

## Changelog

:cl:
map: Fortified the inquisition manor greatly
/:cl:

### DIFF
--- a/_maps/map_files/rockhill/rockhill.dmm
+++ b/_maps/map_files/rockhill/rockhill.dmm
@@ -1047,11 +1047,13 @@
 /turf/closed/wall/mineral/rogue/decowood,
 /area/rogue/indoors/town/church)
 "alI" = (
-/obj/structure/chair/stool/rogue{
-	pixel_x = 14;
-	pixel_y = 13
+/obj/structure/fluff/walldeco/psybanner{
+	pixel_x = -32
 	},
-/turf/open/floor/rogue/herringbone,
+/obj/effect/decal/cobbleedge{
+	dir = 5
+	},
+/turf/open/floor/rogue/blocks/paving,
 /area/rogue/under/cave)
 "alL" = (
 /obj/structure/fluff/statue/small{
@@ -1841,10 +1843,15 @@
 /turf/open/transparent/openspace,
 /area/rogue/outdoors/town/rockhill)
 "atU" = (
-/obj/structure/fluff/walldeco/psybanner{
-	pixel_y = 29
+/obj/structure/rack/rogue/shelf,
+/obj/item/flashlight/flare/torch/lantern{
+	pixel_y = 32
 	},
-/turf/open/floor/rogue/herringbone,
+/obj/item/flashlight/flare/torch/lantern{
+	pixel_y = 32
+	},
+/obj/machinery/light/rogue/candle/blue/r,
+/turf/open/floor/rogue/concrete,
 /area/rogue/under/cave)
 "aue" = (
 /obj/structure/fluff/railing/border{
@@ -2538,11 +2545,12 @@
 /turf/open/floor/rogue/herringbone,
 /area/rogue/indoors/town/manor/rockhill)
 "aBD" = (
-/obj/structure/rack/rogue/shelf,
-/obj/machinery/light/rogue/candle/r,
-/obj/structure/fluff/psycross/psycrucifix/stone,
-/turf/open/floor/rogue/herringbone,
-/area/rogue/under/cave)
+/obj/machinery/light/rogue/candle/blue{
+	pixel_y = 24
+	},
+/obj/structure/fluff/statue/gargoyle,
+/turf/open/floor/rogue/blocks,
+/area/rogue/indoors/inq/basement)
 "aBG" = (
 /obj/structure/stairs/stone{
 	dir = 4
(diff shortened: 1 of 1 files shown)


Loosenail:
Pretty much any location that has a high reward for antags to raid... should have high security. It also always felt stupid that you're functionally better off passing heretics to the Keep or >Church rather than the single faction that is supposed to handle them. I'm sure the weak points in Inquisition Manor were meant to give antags options to break their buddies out if they get caught... but more often the weak points are just relentlessly abused by either bored antags who want to fragbait, shitters, or gnolls with no intention of rescuing anyone.

The design of Inquisition means they're going to be out of their manor for a long time at a time. It just doesn't make sense to make them less secure than places like the Keep or Church where people are pretty much guaranteed to be present at all times.

I like this new change. Sticking the manor more into the cliff itself means there's still ways to get in through mining, but that requires a bit of actual effort rather than just climbing a little.

Jamdrawers:
Maximally Sauceless.

Perhaps we ought to reduce the players capability to break into things more generally rather than reconstructing every part of the map to resemble a maximum security prison.

MickDick1:
The alternative would be like... Quintupling integrity values in a snowflake manner for the inquisition manor's 50 different ways to casually break in.

The inquisition's HQ being a fortified compound built into deep nearby mountain rock is cool in my opinion anyways. On Rockhill no less.
I didn't intend for this to be much in the way of some uniquely clever or pretty solution. It's simple. I know that. It's obvious. It also is what it is, in the current state of affairs round-to-round for the inquisition players.

(This said you're ofc free to take your own crack at solving the problem. IDC as long as "something" gets done. WHICH HAS NOT HAPPENED FOR HALF A YEAR!!!)

UntoldTactics:
> Maximally Sauceless.
> 
> Perhaps we ought to reduce the players capability to break into things more generally rather than reconstructing every part of the map to resemble a maximum security prison.

This.
Advanced climbing, all the various tools and abilities to break into places without nerfing any of the thief classes has generally left it so that even the most basic thief is a hypercompetent mission impossible sneak. 

I like that we're improving the security of places and I concur the Inquisition should be THE hardest place to get into, if you can at all (I frankly subscribe to the concept of a secret base or elsewise), but otherwise it is worrying to see the keep and other places becoming total forts...

Especially when none of it matters because a single club or mace can still break down as many stone walls as a dude needs to get into a place.

sippykot:
> > Maximally Sauceless.
> > Perhaps we ought to reduce the players capability to break into things more generally rather than reconstructing every part of the map to resemble a maximum security prison.
> 
> This. Advanced climbing, all the various tools and abilities to break into places without nerfing any of the thief classes has generally left it so that even the most basic thief is a hypercompetent mission impossible sneak.
> 
> I like that we're improving the security of places and I concur the Inquisition should be THE hardest place to get into, if you can at all (I frankly subscribe to the concept of a secret base or elsewise), but otherwise it is worrying to see the keep and other places becoming total forts...
> 
> Especially when none of it matters because a single club or mace can still break down as many stone walls as a dude needs to get into a place.

yeah... walls are more or less useless right now unless they go to the top z level because people can just casually climb over

DogExtrordinaire:
Hmm... I think this one's a little too much? If you were to ask me I'd take notes from Azure's new inquis manor in that it primarily makes it harder to just climb and walk right in but doesn't actually make it that much harder to directly assault or blow up. it primarily does this by making the fences two z-levels high in most places, but also adds a balcony accsessible to normal orthos so they can, you know, get up on their own roof.

it's a little basic, but the revamp earlier that just added bars to our windows was also...mostly enough.

RadiantFlash7:
I am currently planning on a massive overhaul to rockhill. This includes accounting for climbing skills, because rockhill was never designed with it in mind.
That said- Burying the inquisition manor in rock is probably a bit excessive, truthfully. Somethings could use some tweaking, to prevent climbing up to break in, but I don't think completely burying it in mountain side is the way to go.

sippykot:
> I am currently planning on a massive overhaul to rockhill. This includes accounting for climbing skills, because rockhill was never designed with it in mind. That said- Burying the inquisition manor in rock is probably a bit excessive, truthfully. Somethings could use some tweaking, to prevent climbing up to break in, but I don't think completely burying it in mountain side is the way to go.

can always be changed later during that overhaul
right now its cbt to defend

RadiantFlash7:
Closing this for now. While the inquisition manor ought be more defensive- this is a bit excessive. We also have a huge rockhill remap in the pipeline within the next month.

PR #2455, opened 2026-08-19
Temperature Tweaks

NPC1314:
## About The Pull Request

I like the temperature system in general. In practice it got some parts thats feels unfinished, tweaks that were needed left undone. This PR tries to fix that.

* Temp increases from firestacks reduced and capped. For max heat death you need warm weather or turf effects too, not just firestacks. Reason - gameplay: quickly maxing up heatstroke on all in fireball range is not good design IMO. This makes it slightly less cancer to bring back people who been left burning for a while, and keeps players in the fight a little bit longer vs fire.

* Bumping into mobs while on fire, and walking over burning mobs, has chance to transmit fire reduced to 50% (from 100)
A little less feeling of being drenched in gasoline, while keeping the potential for chaotic fire spreading, only not quite so fast.

* The temp damage over time on several weathers tweaked a little to allow some more exposure time without being immediatly crippling

* Frostbite dont occur any more. Max low temp still gives you hypothermia, and the healing of hypothermia no longer instant once you heat up, it got a 50% to fire each minute, to give some short/medium-term consequences of the exposure.
Reason - Frostbite is terribly implemented, inflicting unique wounds on players with no clear feedback how to fix that also messes up your screen slightly etc is not great. If reworked it could work, or simply flesh out hypothermia a bit, but in this state its better to not have it. Fleshing out hypo is the better choice anyways.

* Braziers got the ability to warm your hands readded. Modest heat boost after 4 second interaction. Reason - its minor, and theres way too few easy to grasp ways to interact with the temp system.

## Testing Evidence

Tested locally by fireballing and weatherspamming the little dudes.
temp

## Why It's Good For The Game

Doesn´t solve every issue concerning temp, and the values might need tweaking again a little in the future, but overall it solves several glaring issues. Needs TM for tweaks of course, local testing worked well.

## Changelog

:cl:
add: brazier heat hands on r-click (+temp)
del: hypothermia no longer progress to frostbite
qol: several weathers tweaked average temp per tick effect
balance: firestacks slower to raise temp and caps out before the max on their own, still give mega heatstroke but less time to cool down
code: bumping into mobs while on fire, and walking over burning mobs now 50% chance to spread not 100
/:cl:

### DIFF
--- a/code/game/objects/lighting/rogue_fires.dm
+++ b/code/game/objects/lighting/rogue_fires.dm
@@ -53,7 +53,7 @@
 			L.visible_message(span_info("[user] starts to warm their hands."), span_info("You warm your hands."))
 			if(do_after(L, 4 SECONDS, target = src))
 				if(L.bodytemperature < BODYTEMP_NORMAL_MIN)
-					L.adjust_bodytemperature(50)
+					L.adjust_bodytemperature(35)
 		return
 
 
--- a/code/game/objects/lighting/rogue_fires.dm
+++ b/code/game/objects/lighting/rogue_fires.dm
@@ -46,6 +46,17 @@
 				icon_state = "[base_state]0"
 			return
 
+/obj/machinery/light/rogue/firebowl/attack_right(mob/user)	// warm your hands a little at a more accessible spot than fireplaces
+	if(isliving(user))
+		var/mob/living/L = user
+		if(on)
+			L.visible_message(span_info("[user] starts to warm their hands."), span_info("You warm your hands."))
+			if(do_after(L, 4 SECONDS, target = src))
+				if(L.bodytemperature < BODYTEMP_NORMAL_MIN)
+					L.adjust_bodytemperature(50)
+		return
+
+
 /obj/machinery/light/rogue/firebowl/off
 	icon_state = "stonefire0"
 	base_state = "stonefire"
--- a/code/game/objects/lighting/rogue_fires.dm
+++ b/code/game/objects/lighting/rogue_fires.dm
@@ -53,7 +53,7 @@
 			L.visible_message(span_info("[user] starts to warm their hands."), span_info("You warm your hands."))
 			if(do_after(L, 4 SECONDS, target = src))
 				if(L.bodytemperature < BODYTEMP_NORMAL_MIN)
-					L.adjust_bodytemperature(35)
+					L.adjust_bodytemperature(20)
 		return
 
 
--- a/code/game/objects/lighting/rogue_fires.dm
+++ b/code/game/objects/lighting/rogue_fires.dm
@@ -53,7 +53,7 @@
 			L.visible_message(span_info("[user] starts to warm their hands."), span_info("You warm your hands."))
 			if(do_after(L, 4 SECONDS, target = src))
 				if(L.bodytemperature < BODYTEMP_NORMAL_MIN)
-					L.adjust_bodytemperature(20)
+					L.adjust_bodytemperature(10)
 		return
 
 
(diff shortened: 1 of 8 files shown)


code/modules/mob/living/living.dm
```diff
@@ -1659,8 +1659,7 @@
  */
 
 /mob/living/proc/on_fire_stack(seconds_per_tick, datum/status_effect/fire_handler/fire_stacks/fire_handler)
-	adjust_bodytemperature(((fire_handler.stacks)) * 0.5 * seconds_per_tick)
-
+	adjust_bodytemperature(((fire_handler.stacks)) * 0.1 * seconds_per_tick)
```
RadiantFlash7: Changing this won't really affect much player facing. As you already editted, handle_fire in datum/species is what will adjust temperature from fire for most human mobs.

code/modules/mob/living/living.dm
```diff
@@ -1708,6 +1707,9 @@
 	if(HAS_TRAIT(spread_to, TRAIT_NOFIRE) || HAS_TRAIT(src, TRAIT_NOFIRE))
 		return
 
+	if(prob(50))	// 50% chance you don´t catch fire from a bump or walking over a burning corpse. Cause people don´t often bathe in gasoline.
```
RadiantFlash7, line 1710: This has been a significant sore point for players, one player being on fire now running into others causing 5 players to be on fire at rapid pace. It might be worth considering tweaking the amount of firestacks being transfered as well if this proves insufficient, but we can give it a try.

code/game/objects/lighting/rogue_fires.dm
```diff
@@ -46,6 +46,17 @@
 				icon_state = "[base_state]0"
 			return
 
+/obj/machinery/light/rogue/firebowl/attack_right(mob/user)	// warm your hands a little at a more accessible spot than fireplaces
```
RadiantFlash7, line 49: I originally added this feature as being an intuitive way to heatup. However- When it was removed, Kroto had set up braziers to heat people up passively- at RAPID rate, just by standing next to it. I'm unsure if clicking on the bowl is even really needed at this point- Though I still think it's nice enough to have, we can try re-adding it.

RadiantFlash7 (commented):
These are fairly well reasoned changes. A couple of things I wanted to point out, but for the most part I'm happy enough to give this a try testmerging.

NPC1314:
I´ll do some more tweaking if the testing works out well, based on feedback here, in a day or two
(snowstorm nerf lessened, firebowl bonus halved so its not too quick)

code/game/objects/lighting/rogue_fires.dm
NPC1314: Maybe its more of a me issue, I just instinctually clicked the braziers and it annoyed me when nothing happened. Ill observe the passive effect more closely next test I run.
Clicking stuff and getting a feedback "warmed hands" is kinda flavorful though, can be worth to keep just for that.

code/modules/mob/living/living.dm
NPC1314: Its a bit crude but should give a inkling if its the right path. Only transfering say, one or two stacks is a good idea too, or some sort of hook for a stat LCK? or being splashed with flammables or somethin too. Ill look into it

code/modules/mob/living/living.dm
NPC1314: Thanks, will remove this unneeded edit in next pass

code/game/objects/lighting/rogue_fires.dm
thezblah: Honestly, the passive warming IS stronger and faster than old handwarming was but the UI is slow to update and hard to notice if you don't know what you're looking for (the temp indicator crystal gets a little fire icon on it to indicate you're warming up).  Overall the handwarming had better feedback, gave you a more active option to combat the cold, and was also very flavorful.

CaydenT:
Holy shit i love you

PR #2448, opened 2026-08-19
Allows Fluvians to be Underdweller mercenaries

Pidgey-chan:
## About The Pull Request

Adds Fluvians to the list of allowed species for Underdweller mercs.

## Why It's Good For The Game

The in-game map references a Fluvian population in the Underdark city of "Mercuriam", therefore it is reasonable for Fluvians to become Underdwellers - at least as much as verminvolk anyway.

## Changelog

:cl:
add: Fluvians can now be Underdwllers.
/:cl:

PR #2462, opened 2026-08-20
Revert "Allows Fluvians to be Underdweller mercenaries", uses erroneous lore from Origins

SneakySlinky:
Reverts Rotwood-Vale/Ratwood-2.0#2448, as it relies on preceding lore that does not exist on this server, mistakenly believed to be the case due to the origins PR that references "Mercuriam", which is lore from another server mistakenly transplanted here.


animalsandsuch:
bad pr.

AlucardHellsingChapel:
Ngl I would love if our lore was changed to the one where fluvians worship saints cocoon (pestra)

Kroto-Ker:
This lore was explicitly approved by the lore head at the time, and I have not heard anything otherwise since.

PR #2473, opened 2026-08-20
Mounted knight grand mace > steel mace

Pidgey-chan:
## About The Pull Request
Changes the mounted knight's grand mace + longbow option to steel mace + longbow instead.

## Why It's Good For The Game
Currently, the mounted knight's grand mace does not spawn because it can no longer go in the belt slot like it could when the loadout was introduced. While this could be fixed by setting the grand mace to spawn in hand, it would probably make more sense for the loadout to spawn with a smaller mace that they can put on the belt while using the bow - currently they would have to clear out a back slot for a greatweapon strap to stow their grand mace or drop it to use their ranged weapon.

## Changelog

:cl:
balance: changes the mounted knight's longbow and grand mace option to longbow and steel mace
/:cl:

### DIFF
--- a/code/modules/jobs/job_types/roguetown/garrison/knight/knight_mounted.dm
+++ b/code/modules/jobs/job_types/roguetown/garrison/knight/knight_mounted.dm
@@ -66,10 +66,10 @@
 				beltr = /obj/item/quiver/arrows
 				beltl = /obj/item/gun/ballistic/revolver/grenadelauncher/bow/recurve
 				H.adjust_skillrank_up_to(/datum/skill/combat/polearms, SKILL_LEVEL_MASTER, TRUE)
-			if("Grand Mace + Longbow")
+			if("Steel Mace + Longbow")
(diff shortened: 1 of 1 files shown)


coffeecoffeecoffee:
Unfortunately this will almost certainly result in this option almost never getting picked compared to the other options, but if that's the only option, then it does make sense. It's barely better than just removing it though.

Pidgey-chan:
> Unfortunately this will almost certainly result in this option almost never getting picked compared to the other options, but if that's the only option, then it does make sense. It's barely better than just removing it though.

Shit, I'll probably pick it. I don't think a steel mace is a bad weapon considering mounted knight has the versatility and Per to use a longbow - which is a _very_ good weapon especially with their town buff up. A steel warhammer was also considered, but they won't be using shields and it felt a bit too on-meta, as it were. I dunno. You can suggest something else.

Xander3359:
It's fine I suppose?

PR #2495, opened 2026-08-23
Port azure combat accuracy balance changes

Gug1234:
## About The Pull Request
Without getting math-jargony, combat accuracy (how likely you are to hit the body zone you have targeted on the doll) has been changed:
- A mob being grabbed or prone no longer has all accuracy checks bypassed on hit. Instead, a 10% bonus to accuracy is applied for passive grab, 20% for aggro grab or better (i.e., pin), and a 30% bonus for being prone. No more 5 perception grapple chuds suddenly gaining pinpoint accuracy after grabbing someone.
- A mob that is prone attacking a standing target's feet or legs no longer bypasses all accuracy checks. Instead, a 5%  bonus is applied.
- Attacking exposed/vulnerable targets, such as those off balance, feinted, baited, riposted, stamina stunned, etc., grants 20% bonus to accuracy
- Pure unarmed punch intent has a 10% bonus, much the same as short-length weapons. 
- Expert Pugilist unarmed attacks no longer bypass accuracy checks against legs and feet; instead, they get a 30% bonus.
- The perception bonus to accuracy is now capped at 40%, which is reached at 15 perception. This applies in both melee and ranged combat
- Very precise face hits are now extra difficult. A 24% penalty is applied when targeting eyes, ears, nose, mouth, and skull. 
Additionally, ranged attacks against face precise zones are clamped at a max accuracy of 30%. No more cheesy 2-tap pinpoint longbow crits as warden. 

Some practical math stuff
If you are aiming hands with a ranged weapon, the current math comes out to:
50% hand, 25% arm, 25% chest.

New math:
40% hand, 24% arm, 36% chest.

Targeting an arm with a ranged weapon currently grants two 75% rolls, effectively ~94%.
New math makes that one roll at 55%

Attacking a face zone with a ranged weapon using the new math comes to 30% face, 21% head, 49% chest.

## Testing Evidence
Character used has 12 PER

https://github.com/user-attachments/assets/f012490f-cfb7-4444-9576-d90b8211ffc4

new prone

new grab

## Why It's Good For The Game
- Reduces the reliability of face-stabbing and face targeting with ranged weapons
- Reduces the 'need' for stacking multiple face coverages as it's notably harder to hit
- Removes guaranteed limb accuracy; I don't need to explain why this is bad. Grappling and conchecking so you can get free 100% accurate hits on someone despite having 7 PER is dumb; this fixes that while still rewarding grapples and hits on prone targets.

## Changelog

:cl: Yuckuza
balance: Targeting face carries a heavier accuracy malus on fully-able player targets.
balance: Guaranteed limb accuracies on hits in certain situations now replaced with high % bonuses
balance: Pure unarmed punch intent has a 10% bonus, much the same as short-length weapons. 
balance: Attacking exposed/vulnerable targets, such as those off balance, feinted, baited, riposted, stamina stunned, etc., grants a 20% bonus to accuracy.
balance: The perception bonus to accuracy is now capped at 40%, or 15 PER.
/:cl:


Pidgey-chan:
Makes open-faced helmets more viable so sure thumbs up.

code/modules/mob/living/combat/accuracy_checks.dm
```diff
@@ -99,49 +128,76 @@
...
 	// Aiming very precise part has a chance of hitting the parent limb instead.
 
+	// We only want these limits vs Players
+	if(zone_type == PRECISE_FACE_ZONE && !mind)
+		zone_type = ULTRA_PRECISE_ZONE
+
 	switch(zone_type)
 		if(ULTRA_PRECISE_ZONE)
-			chance2hit -= RANGED_ULTRA_PRECISE_HIT_PENALTY
+			chance2hit += RANGED_ULTRA_PRECISE_HIT_PENALTY
```
Wrackbang, line 176: The operations here are inconsistent with how it's handled above. I would change the constant to a positive integer and revert the operations to a subtraction assignment, similar to how it's structured for melee math, as that's slightly more intuitive on a first read. I'm not sure where else these constants get used so double check that when you flip the sign.

code/modules/mob/mob_helpers.dm
```diff
@@ -45,6 +45,28 @@
...
+		if(BODY_ZONE_PRECISE_NOSE)
+			return TRUE
+		if(BODY_ZONE_PRECISE_MOUTH)
+			return TRUE
+		if(BODY_ZONE_PRECISE_EARS)
+			return TRUE
+		if(BODY_ZONE_PRECISE_SKULL)
+			return TRUE
+		if(BODY_ZONE_HEAD)
+			if(check_head)
```
Wrackbang: Seems like this optional argument is never used by callers? The function argument and condition on this line should get removed.

code/modules/mob/living/combat/accuracy_checks.dm
```diff
@@ -99,49 +128,76 @@
...
 	// No matter what, 5% chance to hit the zone. No benefit from overaccuracy (unlikely)
 	var/zone_type = ranged_zone_difficulty(def_zone)
-	var/chance2hit = final_accuracy
+	var/chance2hit = final_accuracy + get_zone_ranged_hit_bonus(def_zone)
 	// If you aim very precisely, you take -25 on hit chance, and then no matter what, it is clamped at 50%
 	// If you aim precisely (at limb), -10, 75% max.
 	// Aiming very precise part has a chance of hitting the parent limb instead.
 
+	// We only want these limits vs Players
+	if(zone_type == PRECISE_FACE_ZONE && !mind)
```
Wrackbang, line 171: This function is called on the defender side? I think the `mind` check is inverted here and actually makes non-player mobs harder to target.

Gug1234: It's not, this just treats the percise face target zones on mindless carbons like hands and feet, instead of using the ACC_FACE_SUBZONE_PENALTY which is -24% which is JUST for player carbons

code/modules/mob/living/combat/accuracy_checks.dm
Gug1234: they are just local. I prefer them as is, makes the code comments next to the constant definitions up top read easier.

code/modules/mob/mob_helpers.dm
Gug1234: it's from the weapon binding stuff upstream, good catch.

wikimody:
I dont particulary like the acuracy cap, with the bow cap in place it now makes it so per past 15 is complelty usless

Gug1234:
> I dont particulary like the acuracy cap, with the bow cap in place it now makes it so per past 15 is complelty usless

per past 15 means you can still look 2 screens ahead among other things.

also lol, ranged players deserve this and more

Genevesh:
> per past 15 means you can still look 2 screens ahead among other things.
> 
> also lol, ranged players deserve this and more

Eh, I'm with the port author here, statmaxxing in all shapes and forms is stupid and shouldn't be a thing. We're already getting strength and speed and intmaxxers, we don't need ANOTHER stat added to our combatslop.

Tekkomeka:
It surely took a while to touch the most broken (in terms of code) stat in the game. But it's also surely a good change. Accuracy is one of the hugest issues for combat.

I'm too lazy to check if the 11 PER oversight was fixed yet or not. I didn't play for a while, but there was a line where you were supposed to fail checks, but it was missing return statement, so, if you had 11 PER+ you were literally getting a second roll for success. _That's why you could see people running eagle beak with 10 PER failing all head attacks, while 11 PER characters almost never missed lmao, despite having penalty for blunt type even._

DorfGuy1:
> It surely took a while to touch the most broken (in terms of code) stat in the game. But it's also surely a good change. Accuracy is one of the hugest issues for combat.
> 
> I'm too lazy to check if the 11 PER oversight was fixed yet or not. I didn't play for a while, but there was a line where you were supposed to fail checks, but it was missing return statement, so, if you had 11 PER+ you were literally getting a second roll for success. _That's why you could see people running eagle beak with 10 PER failing all head attacks, while 11 PER characters almost never missed lmao, despite having penalty for blunt type even._

Yeah uh, it seems like at ten per you miss A LOT

CaydenT:
Complaints about building above 15 in any stat are silly. I really don't think we should encourage turbo minmax in any capacity. I like that 15 is becoming more of a "target" stat and less of a "minimum" stat.

This is also a sensible grab nerf without ruining it, good stuff.

usernameasterisk:
> > I dont particulary like the acuracy cap, with the bow cap in place it now makes it so per past 15 is complelty usless
> 
> per past 15 means you can still look 2 screens ahead among other things.
> 
> also lol, ranged players deserve this and more

not mad just saying that maybe instead of heavily nerfing all archers something should be done about mages who can land you in crit for 3-5 fireballs, no matter the zone, armor, pain resist, critresist trait etc
said "casters" are archers+ now, who is immune to mouthgrab (just wear any mask, not even armored one), can break grabs just by spamming fireball or spitfire (they are considered explosions) into their enemy and got BIG arsenal of diverse combat tools in form of their spells

ngl it's kinda ridiculous how nerdy bookworms can easily outperform specialized warriors in their own art - battles and wars. even ravox should fear noc if he isn't fireproof.

PR #2445, opened 2026-08-18
Add ARMOR_BRIGANDINE define. B+, A+, C, B+. Sidegrade to plate rather than reskin.

Gug1234:
### New dedicated armor rating for brigandine/splint gear. 
newbrig

Per armor_defines.dm:_"Better blunt and arrow padding than plate, but sword stabs and above will pen. Best light armor gets for melee. Medium/heavy classes should still wear maille under it!"_

**Blunt 70 | Slash 90 | Stab 40 | Pierce 70**

Additionally, a new ARMOR_INT_CHEST_PLATE_BRIGANDINE_WEIGHT_MODIFIER define is added to add or subtract 60 integ from heavy brigandines (coats of plates) and light brigandines, respectively. 

Also changed psyalmist helmet from ARMOR_LEATHER to ARMOR_STUDDED since the description mentioned it being studded with blessed coinage or something. 

> [!IMPORTANT]
> All integ values shown are the raw total integ, which isn't what's displayed in game. The in-game integ is 90% of the total real integ. Because of our horrible shitcoded codebase, armor continues to protect PAST its 90% break threshold (which can be observed when you "break" armor but it still blocks 1 more hit; comparatively, peels allow damage through the moment they're successfully executed because they aren't a bugged mess.)

70 blunt will NOT prevent blunt chip, which is hardcoded to only mitigate at blunt protection values >80. Wear a gambeson if you dont want chip damage.

| Armor | Blunt | Integ | Effective blunt integ | Hits |
|--------|--------|--------|--------|--------|
|Steel scalemail | 10 | 300 | 226 damage | 7 strikes|
|Lightweight brigandine | 70 | 340 | 374 damage | **12 strikes**|
|Steel fullplate | 10 | 500 | 377 damage | 12 strikes|
|Brigandine | 70 | 400  | 440 damage | **14 strikes**|
|Coat of plates | 70 | 460  | 506 damage | **16 strikes**|
|Gilded fullplate | 50 | 700  | 668 damage | 21 strikes|
|Blacksteel fullplate | 80 | 600 | 715 damage | 22 strikes|

90 slash protection and 100 slash protection behave identically against most sword strike intents. The only weapon intents getting through slightly easier would be chops from a vicious axe wielded by an unbound strength graggite.

Armor | Slash | Integ | Hits |
|--------|--------|--------|--------|
Brigandine | 90 | 400 | 13
Light brigandine | 90 | 340 | 11
Coat of plates | 90 | 460 | 14
[...]

PR #2496, opened 2026-08-23
Fix bug with projectiles passing their full scaled damage to wound and embed rolls when dealing any amount of penetrating damage.

Gug1234:
## About The Pull Request

- Added "armor" argument to projectile wound and embed call
- Projectiles no longer pass their entire scaled damage when doing any amount of penetrating damage.

## Testing Evidence
This is something I found in #2445 and went over there, below is the math from the prior markdown. 
living_defense.dm line 65 /mob/living/bullet_act(obj/projectile/P, def_zone = BODY_ZONE_CHEST) calls apply_damage()
```dm
if(!apply_damage(P.damage, P.damage_type, def_zone, armor))
``` 

species.dm line 2040 /datum/species/proc/apply_damage(damage, damagetype = BRUTE, def_zone = null, blocked, mob/living/carbon/human/H, forced = FALSE, spread_damage = FALSE) subtracts armor from damage:
```dm
damage = max((damage - blocked) * (1 - armor / 100), 0)
```

living_defense.dm line 69 runs the wound call, which OMITS "armor"
```dm
if(P.woundclass)
    check_projectile_wounding(P, def_zone)
```

Because blocked is null, carbon_defense.dm line 72 treats the argument as 0
```dm
/mob/living/carbon/check_projectile_wounding(obj/projectile/P, def_zone, blocked)
    var/obj/item/bodypart/BP = get_bodypart(check_zone(def_zone))
    if(BP)
        var/newdam = P.damage * (100-blocked)/100
        BP.bodypart_attacked_by(P.woundclass, newdam, ...)
```
bow scaled with 15 PER comes to 60; wound calc uses 60 regardless of armor block. 
P.damage = 60 so...
```dm
newdam = P.damage × (100 − 0) / 100
       = P.damage
       
 newdam = 60 x (100-0)/100
        = 60
```
> [!IMPORTANT]
>  Layman's terms: 10 HP through a brigandine from a 15-PER broadhead is treated as the full 60 damage during wound calculation. ANY PENETRATION can cause crit wounds quickly, regardless of actual raw damage. This is exacerbated by most armor lacking BCLASS_PIERCE crit protection.

## Why It's Good For The Game
Bugfix. Projectiles are no longer feeding wound math 40 force when only doing 4 real damage through someone's armor. This is already the case with check_projectile_dismemberment.

## Changelog

:cl: Yuckuza
fix: fixed projectiles passing their entire scaled damage when doing any amount of penetrating damage.
/:cl:

### DIFF
--- a/code/modules/mob/living/living_defense.dm
+++ b/code/modules/mob/living/living_defense.dm
@@ -70,7 +70,7 @@
 			if(P.dismemberment)
 				check_projectile_dismemberment(P, def_zone,armor)
 			if(P.woundclass)
-				check_projectile_wounding(P, def_zone)
+				check_projectile_wounding(P, def_zone, armor)
 
 			if(P.poisontype)// New proc for poisoning that respects if armor stopped damage from the projectile, by blocking or through reduction. Only called if poison type is defined.
 				if(!P.poisonamount)
@@ -81,7 +81,7 @@
 					if(P.poisonfeel)
 						M.show_message(span_danger("You feel an intense [P.poisonfeel] sensation spreading swiftly from the area!"))
 
-			if(P.embedchance && !check_projectile_embed(P, def_zone))
+			if(P.embedchance && !check_projectile_embed(P, def_zone, armor))
 				P.handle_drop()
 
 		else


Wrackbang:
> Because blocked is null, carbon_defense.dm line 72 treats the argument as 0

DML is an incredibly cursed language

RadiantFlash7:
I don't think I can count a detailed explanation as testing evidence- as helpful as it is- Though if it functions as you say, it's probably worth test merging.

Gug1234:
> I don't think I can count a detailed explanation as testing evidence- as helpful as it is- Though if it functions as you say, it's probably worth test merging.

I was digging and saw the same issue was fixed on Azure back during their armor rework. It's been fixed there for some time.

PR #2498, opened 2026-08-23
Increase phoenix guard activation from every 20s -> 60s

Gug1234:
## About The Pull Request

- Defines PHOENIX_GUARD_COOLDOWN as 60 SECONDS

## Testing Evidence
Number change.

## Why It's Good For The Game
Fewer free firestacks that interact with our temp system. AP parity.

## Changelog

:cl: Yuckuza
balance: Phoenix Guard cooldown raised from 20s -> 60s
/:cl:

### DIFF
--- a/code/datums/magic_items/greater_items/greater.dm
+++ b/code/datums/magic_items/greater_items/greater.dm
@@ -1,3 +1,5 @@
+#define PHOENIX_GUARD_COOLDOWN 60 SECONDS
+
 ///T3 Enchantmentsdatum
 /datum/magic_item/greater/lifesteal
 	name = "life steal"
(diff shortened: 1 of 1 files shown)


wikimody:
Its not very good as it is alredy, fire is not particularly dangerus unless you can spam it like mages can

Gug1234:
> Its not very good as it is alredy, fire is not particularly dangerus unless you can spam it like mages can

Coupled with temp, any amount of firestacks means you have to stop and cool down, or else suffer horrendous stamina consumption debuffs the rest of combat

wikimody:
That activates on the second temp stage, not the first. It takes a while to get to said second stage even while on fire. It goes: neutral - warm (does basicly nothing) - overheating (stamina hell)
asuming you pat out the flame within around 15 seconds from geting set on fire, and its not being constantly reaplied (mages again) you will stay on warm, with has very minimal effects

PR #2503, opened 2026-08-24
Gives Eastern Warrior (Adventurer class) TRAIT_NOPAINSTUN

kimesik:
## About The Pull Request
Gives Eastern Warrior (Adventurer class) TRAIT_NOPAINSTUN ("Enduring")

## Testing Evidence
A single line change. It compiles and works.

## Why It's Good For The Game
Right now Eastern Warrior is worse than any Warrior-type Adventurer class, because you are a light armour guy who doesn't get Expert Dodger, Critical Resistance or armour training.

This makes Eastern Warrior more palatable without turning him into Battlemaster-but-slightly-different.

## Changelog

:cl:
balance: Eastern Warrior gets Enduring.
/:cl:

### DIFF
--- a/code/modules/jobs/job_types/roguetown/adventurer/types/combat/foreigner/eastern.dm
+++ b/code/modules/jobs/job_types/roguetown/adventurer/types/combat/foreigner/eastern.dm
@@ -7,7 +7,7 @@
 	allowed_races = NON_DWARVEN_RACE_TYPES // Clothing has no dwarf sprites.
 	outfit = /datum/outfit/job/roguetown/adventurer/kazengun
 	class_select_category = CLASS_CAT_NOMAD
-	traits_applied = list(TRAIT_STEELHEARTED)
+	traits_applied = list(TRAIT_STEELHEARTED, TRAIT_NOPAINSTUN)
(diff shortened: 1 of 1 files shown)


coffeecoffeecoffee:
I really, really don't think we need more buffs to these classes when there are already loads of very good Kazengun mercenary classes and an extreme overabundance of what was supposed to be a mysterious and scarce far-away culture. I don't appreciate these adventurers, and I wish that Kazengun classes were more exclusive.

Plum-but-gayer:
nopainstun expert weapon adventurer... I was going to balancejak but realized they don't have crit resist like I remember them having(?) so this is basically battlemaster with less weapon choice who gets enduring instead of medium armor. I guess it's fine? Nopainstun is REALLY good though. But also I guess we hand it out to a bunch of races and virtues so if someone wanted to powergame they'd just use that.

CaydenT:
Nopainstun is good but probably not as good as any of the other defensive traits in actual combat. If anything it's more like QoL because it means no constant screaming and falling down when stunning your toes while adventuring.

That is to say, this is a cool change that makes them a bit unique.

PR #2513, opened 2026-08-26
AHelp logging to DB.

Lezerchil:
## About The Pull Request

This brings back db logging of ahelps to RW.
This thing requires rigorous testing from RW Staff.
Amongst obvious problems this thing have certain actions, e.g. Cake, which carry a massive chunk of HTML garbage with them into BD log.

## Testing Evidence

{9939E5C3-7CD4-475E-9D20-85237C2BE402}

## Why It's Good For The Game

Logging is generally good for situations when you have to recall certain events, such as disputes and reports.

## Changelog

It doesn't really change anything. It adds a few procs here and there 

:cl:
server: Ahelp recieves logging to DB
:cl:

### DIFF
--- a/code/controllers/subsystem/blackbox.dm
+++ b/code/controllers/subsystem/blackbox.dm
@@ -282,6 +282,25 @@ Versioning
 	key = new_key
 	key_type = new_key_type
 
+/datum/controller/subsystem/blackbox/proc/LogAhelp(ticket, action, message, recipient, sender)
+	if(!SSdbcore.Connect())
+		return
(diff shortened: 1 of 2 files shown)


Arturlang:
@Lezerchil why is this needed over to-file logging which we already have? from atlesat what i see with log_admin_private

Lezerchil:
Because Useroth asked me to.
Perhaps it is easier to deal with DB than whatever we have. Or he wants to double it in case of whatever. He didn't tell me I didn't ask.

PR #2518, opened 2026-08-27
fix for naledi clothing greyscale with boobs

RadiantFlash7:
## About The Pull Request
Disables boob detailing on naledi qaba and kandy.
This handles the issue for these select clothings atleast.

## Why It's Good For The Game
Boob detailing was stealing all of the color being applied by dying it, for approximately 2 pixels of colorization.
Now you can walk around with properly colored robes.

## Changelog

:cl:
fix: disabled bood detailing on qaba and kandy to stop it stealing all dye color from dyed naledi clothes
/:cl:


The-Sun-In-Splendour:
this seems to also be bugged on the hierophant robes

Edit: actually I think the hierophant robes are literally just the khandys but renamed for some reason..?

PR #2522, opened 2026-08-28
Fix DNR examine text checking the wrong mob

zoktiik:
## About The Pull Request
See changelog

## Testing Evidence
Compiled

## Why It's Good For The Game
Informational integrity

## Changelog

:cl:
fix: Fixed "This was their only chance at lyfe." not being appropriately displayed. It was checking if YOUR mob is "dead", not the DNR person, so examining the (living) Duke as a ghost would show that message.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/examine.dm
+++ b/code/modules/mob/living/carbon/human/examine.dm
@@ -398,8 +398,8 @@
 		// Shouldn't be able to tell they are unrevivable through a mask as a Necran
 		if(HAS_TRAIT(src, TRAIT_DNR) && src != user)
 			if(HAS_TRAIT(user, TRAIT_DEATHSIGHT))
-				. += span_danger("They extrude a pale aura. Their soul [src.stat == DEAD ? "was not" : "is not"] clean. This is it for them.")
-			else if(user.stat == DEAD)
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/carbon/human/examine.dm
```diff
@@ -398,8 +398,8 @@
 		// Shouldn't be able to tell they are unrevivable through a mask as a Necran
 		if(HAS_TRAIT(src, TRAIT_DNR) && src != user)
 			if(HAS_TRAIT(user, TRAIT_DEATHSIGHT))
-				. += span_danger("They extrude a pale aura. Their soul [src.stat == DEAD ? "was not" : "is not"] clean. This is it for them.")
```
RadiantFlash7, line 401: Src is explicitly the person being examined. Two lines above, it checks if they have the DNR trait and ensures src doesn't equal user.
 To my knowledge, this has been functioning fine?

zoktiik: It's just cleaner to nix it, the fix is in the other line (`user.stat` to `stat`).

RadiantFlash7: Fair enough.

PR #2524, opened 2026-08-28
Allows harpies to take flight and move down z levels while moving.

Xander3359:
## About The Pull Request
Cleans up some flight-related code
Allows harpies to take flight without having to stand still
Allows harpies to move down a z level while moving.
Makes moving down z levels double as fast.

NOTE: YOU STILL NEED TO STAY IN PLACE TO FLY UPWARDS

## Testing Evidence
Worked on my machine

## Why It's Good For The Game
Adds some QOL to make flight feel more fluid.
It might be a bit too much of a buff, remains to see. Though, harpies already have a con malus and if they are just con checked it cancels their flight, they still need to stay in place in order to move upwards so they can just be grabbed and pulled out of the channel.

## Changelog

:cl:
balance: Harpies can now move while taking flight.
balance: Harpies can now move down z levels faster while in flight, and can move while doing so
/:cl:

### DIFF
--- a/code/datums/keybinding/mob.dm
+++ b/code/datums/keybinding/mob.dm
@@ -183,56 +183,56 @@
 	category = CATEGORY_HUMAN
 
 /datum/keybinding/mob/fly_up/down(client/user)
-	if(iscarbon(user.mob))
-		var/mob/living/carbon/C = user.mob
-		if(C.flying)
-			var/turf/open/transparent/openspace/turf_above = get_step_multiz(C, UP)
-			if(C.canZMove(UP, turf_above))
-				var/athletics_skill = max(C.get_skill_level(/datum/skill/misc/athletics), SKILL_LEVEL_NOVICE)
-				var/stamina_cost_final = round((10 - athletics_skill), 1)
-				var/atom/movable/pulling = C.pulling
-				var/time_taken = 1.5 SECONDS
-				if(ismob(pulling))
-					stamina_cost_final *= 2 //double our stamina cost if we're pulling someone with us
-					time_taken *= 2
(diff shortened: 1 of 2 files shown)


hugmeorelse:
Curious to see how this goes, especially considering harpies have been historically considered fairly weak out of fears that they might be too overpowered with their unique abilities. 

I don't think this will push them over the edge, especially considering they can get con-checked when taking off from the ground. I think the only thing of concern is namely changing z-levels while in flight, while moving about. But it seems like it shouldn't be an issue if it's only while moving down a z-level?

You can only go down so much really. You just have to keep an eye out to see if it gets abused to the point harpies become the ultimate skirmisher. But with such a low constitution, I doubt it.

Genevesh:
I remind you that grappling a harpy that tries to take off will more than likely stop them from flying altogether - and despite being able to move while changing a z-level, their stamina is still not at all infinite. This also doesn't touch anything relating to the costs of holding someone, etc etc... overall this PR really is more QOL than an actual buff.

Let's... also be real, melee harpies are a myth, and ranged harpies are virtually untouched by this because whether you're casting (low wil by default for most spellcasters) or using ranged weapons (all require a free hand), you kinda HAVE to stop flying somewhere. 

Like the answer above said, you can only go down so much. This will let harpies sweep down like birds should be able to.

That's a pretty nice PR, Xander. NOW CODE HER LAYING AN EGG!

Heheniy:
so it is just a free escape button in a pursuit

Arturlang (approved):

Arturlang:
to clarify, this does not let you take off instantly, only lets you move during the do_afters

PR #2527, opened 2026-08-28
Assorted instrument & psyaltrist fixes

Ephemeralis:
## About The Pull Request

Simple stuff:

- **Instruments now remember the last song they've played, and if you use them while in combat mode, they automatically play your last played song without you needing to traverse any additional menus.**
    - I don't know how anyone has ever played a bard in combat before now without this.
- Instruments now have helpful text on their examine that explains the MMB audience-adding behavior and the song replaying described above, if you have inspiration.
- The psyaltrist now spawns a scabbard properly when picking a rapier as its main weapon.
- The psyaltrist now has the appropriate devotion regen rate for a T2 miracle class instead of having T1 capacity and regeneration worse than devotee. (250 -> 400 max devotion, 0.1 (weak) -> 0.5 (minor) regen). They're equal to an adventurer cantor with this change.
    - All psydonite miracles use a shit ton of devotion and at the prior values, a combat-active psyaltrist had to spend nearly **8 minutes** MMB channeling prayer to get enough devotion back to full, and could cast a whopping total of 4 miracles before it'd deplete.

## Testing Evidence

dreamseeker_7bjKjQETaJ

dreamseeker_pvuqL4gVQ0

## Why It's Good For The Game

Less control burden with an already VERY control-heavy class (bard) is great for everyone involved. All of these should be pretty self-explanatory, really.

## Changelog

:cl: Ephemeralis
qol: Instruments now play their last-played song immediately if used in combat, instead of needing you to navigate two dialogs while completely still.
qol: Instruments now have helpful examine text for characters with inspiration to explain the above change, and other QoL changes made recently.
fix: Fixed the psyaltrist's scabbard not spawning when picking the rapier loadout.
fix: Fixed the psyaltrist to use the proper devotion values for a T2 miracle class.
/:cl:

### DIFF
--- a/code/game/objects/items/rogueitems/instruments.dm
+++ b/code/game/objects/items/rogueitems/instruments.dm
@@ -247,6 +247,8 @@ GLOBAL_LIST_EMPTY(instrument_band_lobbies)
 	/// When TRUE, songs will loop (repeat) when they end. Off by default.
 	var/loop_enabled = FALSE
 
+	var/last_played // store the last played thing we have here to use in quick cmode toggle
+
 // Added null-guard on soundloop. During Initialize() the parent chain
 // may trigger equipped() before soundloop is assigned.
 /obj/item/rogue/instrument/equipped(mob/living/user, slot)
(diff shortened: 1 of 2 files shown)


wikimody:
I wrote something but then realized its already implemented, disregard this coment

PR #2528, opened 2026-08-28
Enables Furry Sprites to Use Eyeliner + Eyesocket Accessories (BOUNTY)

SirSwagmeyer:
## About The Pull Request

Adds eyeliner and eyesockets to the facial accessories tab in your character creation menu. This enables you to add eyeliner and eye sockets to your characters _without_ the snout, horns, etc getting in the way. 

## Testing Evidence

image
image

## Why It's Good For The Game

Commission job. Character customization accessibility is good.

## Changelog

:cl:
add: Added eyeliner and eye sockets to the character customization pool.

/:cl:


Gug1234:
Nice, could also have forced the face body marking to the top layer or at least above snouts, but this also works

Loosenail:
> Nice, could also have forced the face body marking to the top layer or at least above snouts, but this also works

I'm sure this is something you can do, but TBH it always made more sense to me for eye-related markings to be located in the accessory or face detail slot. Though providing them in both accessory and head markings allows people to still have stuff like earrings if they aren't using a snout that covers the eyeliner. And on the other hand, having them accessible in accessory section frees up a head marking slot for a lot of people.

SirSwagmeyer:
> Nice, could also have forced the face body marking to the top layer or at least above snouts, but this also works

You're likely right, but body markings are weirdly finnicky with that stuff, and the client specifically requested I provide an alternative via the face accessory slot.

PR #2540, opened 2026-08-30
AP #8507: gnoll satchels, basic gnoll skills, quiet gnoll speech

zoktiik:
## About The Pull Request
Port Azure-Peak/Azure-Peak#8507. I like the rationale expressed in that PR's thread, go see there for more details.

- If on SNEAK intent, gnolls produce a far quieter, more indistinct speech sound when they talk.
- Gnolls have their **right shoulder** slot opened up, and all classes spawn with a gnoll satchel equipped. The only difference is that the satchel does not show a sprite on the mob.
  - Left shoulder is the backpack shoulder. They can't equip backpacks.
- All Gnoll subclasses now get basic some basic stealth/survival skills. See the diff for more details.
  - Most notably, **Journeyman Sneaking** to Berserker, Knight, Templar. (All three previously had none; Shaman and Impure already have Expert Sneaking.) 
  

## Testing Evidence
Compiled. Azure tested it for me!

Update: Tested while also testing other gnoll PRs.

## Why It's Good For The Game
I sought out this PR because people were mentioning gnolls being able to use satchels on AP, and it makes a lot of sense. The other changes are really nice, simple fixes that I don't think would hurt to include. 

Gnolls should probably be able to have a conversation without all of China knowing they're here.

## Changelog

:cl:
balance: Gnolls now have baseline sneaking and better crafting skills. Gnolls now have an unlocked right back slot, as well as a satchel to go with it. Mind, the right back slot does not allow for larger storage like backpacks.
balance: On SNEAK intent, Gnolls produce a far quieter, more indistinct speech sound when talking.
/:cl:

### DIFF
--- a/code/modules/antagonists/roguetown/villain/gnoll/gnoll.dm
+++ b/code/modules/antagonists/roguetown/villain/gnoll/gnoll.dm
@@ -242,3 +242,9 @@
 	miss_text = "strikes the air!"
 	miss_sound = "bluntwooshlarge"
 	attack_verb = list("punches", "strikes", "tears")
+
+
+/obj/item/storage/backpack/rogue/satchel/gnoll 
+	name = "stained satchel"
+	desc = "A fetid sack fashioned into a storage accessory. Whatever's put there inevitably comes out twice the bloody."
+	mob_overlay_icon = null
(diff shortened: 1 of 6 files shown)


Kneesox44:
Praise be- the ability to hold more than a single needle.

PR #2549, opened 2026-08-31
Lets Bards of T2 or above play their instrument with an offhand occupied

ceeetchell:
## About The Pull Request

adds var/can_play_with_occupied_offhand to instruments.dm with a conditional to allow bards t2 and above (bards that use combat mechanics) to be able to play instruments in their offhand.

this is so that when you get con checked and drop your instrument, you don't have to deal with the clunky mess of having to pick up your instrument and weapon, sheathe your weapon (or leave it on the ground), start playing music, and then equip your weapon.

now, you can just pick up both, restart your song, and you're back in the fight.

## Testing Evidence

simple var addition + adding it to the conditionals, compiles, Works On My Machine™

## Why It's Good For The Game

clunky menus and non intuitive restrictions are bad and annoying

## Changelog

:cl: mushy

balance: Bards of T2 and above can now play instruments with an occupied offhand.

/:cl:


zoktiik:
what if you held onto your instrument when getting conchecked...

Ephemeralis:
What if we shot con-checking into the sun?

ceeetchell:
> What if we shot con-checking into the sun?

the world isn't ready for this truthnuke

DogExtrordinaire:
> what if you held onto your instrument when getting conchecked...

not dropping your weapons at all when tackled is how it works on azure. You can only drop weapons there by bait kicking or wall kicking

PR #2554, opened 2026-08-31
Dodge Fix for Simplemobs.

Lezerchil:
## About The Pull Request

I don't know what happened to simplemobs and why suddenly they've become undodgable, but I think I found a problem and the stupidest fix to it.
Before any calculations to a dodge there was a check if an attacker is a human (ishuman). Because of this simplemobs attacking a player triggered a dodge attempt which ended up in a runtime error.

I'm not sure why it broke now and not some time earlier because this code was written about 13 months ago.
I don't know if it will cause more problems down the line, but it seems like it is working.

## Testing Evidence

https://github.com/user-attachments/assets/29245e95-4dbe-4fdf-bb37-11f1090750c4

## Why It's Good For The Game

People asked me to fix it. I assume it is a good thing making dodge builds viable.

## Changelog

:cl:
fix: do_dodge should work on simplemobs too.
:cl:

### DIFF
--- a/code/modules/mob/living/combat/dodge.dm
+++ b/code/modules/mob/living/combat/dodge.dm
@@ -113,10 +113,9 @@
 	var/drained_npc = 5
 	if(ishuman(src))
 		H = src
-	if(ishuman(user))
-		UH = user
-		if (UH.used_intent)
-			I = UH.used_intent.masteritem
+	UH = user
+	if (UH.used_intent)
+		I = UH.used_intent.masteritem
 	var/prob2defend = U.defprob
 	if(L.stamina >= L.max_stamina)
 		return FALSE


RadiantFlash7:
I'm not actually sure this is fixing the issue at present. On my local, I'm able to dodge just fine. The runtime that seems to pop up on the live version of the server seems to be related to status_traits in the do_dodge proc.

@Lezerchil Can you confirm you can't dodge without your change on your local?

Lezerchil:
I mean I double checked. @RadiantFlash7 

https://github.com/user-attachments/assets/9855841a-295c-4a54-bcdd-06354feaf83b

I'm generally speaking not sure myself. What Runtime exception does? Does it ruin the whole proc or it force the current IF check to be FALSE?
Generally speaking how I looked at it. I took parry.dm because parry worked on wolf and dodge didn't. And I wanted to bring the two to one somewhat compatable form. But then I found a runtime error and I asked Useroth about it, and he confirmed it. I went to the line where it NULLs and read the arguments. So, the "UH" variable was null. I went up where it is defined as 
`If (ishuman(src))
UH=user`
I repalced it with what you see right now and it worked.

GrayRachnid:
I'm going to state the obvious and asked if this bug happens on a local without test merges.
it might be a test merge causing the bug if nobody can locate it.

Lezerchil:
It happens on my local, which I try to keep as close as possible to Ratwood main.
When I initially made https://github.com/Rotwood-Vale/Ratwood-2.0/pull/2230 I got a few reports from peeps, but then I locally tested it and it was generally fine. I cannot say anything about exceptions at the time.
I went to make a DB log for AHelps, then I got contacted on the matter again. I reset my local to ratwood and it was broken.

PR #2565, opened 2026-09-02
the great custom descriptorening

zeroisthebiggay:
## About The Pull Request

Custom descriptors. Code taken from AP for the most part, with little tweaks here and there because we're not quite AP in terms of code parity. The images speak for themselves.

## Testing Evidence

image

image

## Why It's Good For The Game

More character customization is GOOD. If people abuse this, we can kill them for it like how I imagine we already do if you abuse your two customs.. I see a lot of opportunities for funny and interesting things with this, plus it means noone has to ever add a descriptor ever again! Yay!

## Changelog

:cl:
add: custom descriptors for everything
/:cl:

### DIFF
--- a/code/datums/mob_descriptors/descriptors/body.dm
+++ b/code/datums/mob_descriptors/descriptors/body.dm
@@ -95,7 +95,7 @@
 
 /datum/mob_descriptor/body/custom
 	name = "Custom Body"
-	var/custom_index = 8
+	custom_index = 8
 
 /datum/mob_descriptor/body/custom/can_describe(mob/living/described)
 	return length(described.custom_descriptors) >= custom_index
--- a/code/datums/mob_descriptors/descriptors/body.dm
+++ b/code/datums/mob_descriptors/descriptors/body.dm
@@ -92,3 +92,14 @@
 
 /datum/mob_descriptor/body/burly
 	name = "Burly"
+
+/datum/mob_descriptor/body/custom
+	name = "Custom Body"
+	var/custom_index = 8
+
+/datum/mob_descriptor/body/custom/can_describe(mob/living/described)
+	return length(described.custom_descriptors) >= custom_index
+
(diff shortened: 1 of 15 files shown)


code/datums/mob_descriptors/descriptors/body.dm
```diff
@@ -92,3 +92,14 @@
 
 /datum/mob_descriptor/body/burly
 	name = "Burly"
+
+/datum/mob_descriptor/body/custom
+	name = "Custom Body"
+	var/custom_index = 8
```
Xander3359: At this point, it should just be a variable on `mob_descriptor`

code/modules/client/preferences_descriptors.dm
```diff
@@ -130,6 +150,59 @@
...
 
+/datum/preferences/proc/preview_descriptors(mob/user)
+	if(!COOLDOWN_FINISHED(src, descriptor_preview))
+		to_chat(user, span_warning("You must wait before previewing descriptors again."))
+		return
+	COOLDOWN_START(src, descriptor_preview, 5 SECONDS)
+	to_chat(user, span_notice("-- Preview of [real_name]'s descriptors --"))
+
+	// someone please fix this horror one day
+	var/mob/living/temp = new /mob/living(null)
```
Xander3359: Can't you just apply these to the client's preview mob?

Xander3359 (changes requested):

code/modules/client/preferences_descriptors.dm
zeroisthebiggay: Yes, actually. I ended up with something like that before I decided to just nearly one to one steal AP's code.
Minor side effect but more of an upside really, is that now you can see genital information since it's copying directly from the mannequin instead of making a features devoid dummy mob.

image

P-roxii:
Is it just me or does this not actually work despite being merged?

PR #2572, opened 2026-09-02
Removes the ability to create enchanting scrolls of Phoenix Guard or Frostveil

CameronWoof:
## About The Pull Request

Title. The items still exist, in case they're used in loot tables or something, and as rare loot they'd probably be fine. 

## Why It's Good For The Game

Enchanting has fundamental issues that I do not have the skill or knowledge to fix; namely, you can place the same enchantment on multiple items (or enchant the same item multiple times) and the effects stack. This should ideally be fixed at some point (and at that point, it would be reasonable to reintroduce these enchantments) but until such a time, armor enchants with offensive procs like this cannot exist. 

## Changelog

:cl: CameronWoof
balance: Phoenix Guard and Frostveil enchanting scrolls can no longer be produced.
/:cl:


Dagonlives:
I use these items all the time.  I think they are not a good idea to remain in module until other changes can be instituted to balance them more easily.

GrayRachnid:
Bringing the balance hammer down on Mages with removals without having longterms plans to balance them in an interesting way is a short-term solution that will cause long-term displeasure and hurt feelings towards the role and devs. Wretches were constantly hated on by the necromancer nerfs until Lich was re-enabled. I don't mind you removing these, but you're not balancing them- You're just removing what people see and complain about the most.

CameronWoof:
> Bringing the balance hammer down on Mages with removals without having longterms plans to balance them in an interesting way is a short-term solution that will cause long-term displeasure and hurt feelings towards the role and devs. Wretches were constantly hated on by the necromancer nerfs until Lich was re-enabled. I don't mind you removing these, but you're not balancing them- You're just removing what people see and complain about the most.

At some point in the future there might be a meeting between the people who recognize there are issues and the people with the talent to fix them - as of yet, that has not happened. Stopgap solutions are unfortunately all we have until something better comes along, and it's more important that these issues stop negatively impacting rounds than it is for mages to retain access to the huge glut of murder tools they currently have at their disposal.

GrayRachnid:
If your solution is short term then I'd rather you do nothing at all. Humans are meant to adapt, so do that instead. 

These are not major game-breaking disasters, they are an edge mages get over their melee counterparts. Let them exist. Just because mages are different in a strong way does not mean they should be struck down.

You're going to remove these and it'll push people to explore other enchantments. Then they'll make these other enchantments and suddenly you're removing those too in a few months time.

Removals removals removals. Most of your balance prs are removals. I'd rather see creativity from you rather than removals. You're removing options in a game that benefits from having a huge list of options.

GrayRachnid:
Kicking the can down the road for someone more talented is how features die.

coffeecoffeecoffee:
I don't think these are serious or pressing problems and I rarely see them ingame ever, at all. I think it has been weeks since I last saw anybody bothering with enchantments.

Trilaterall:
I personally think Phoenix Guard should just be completely removed from the game. You can’t counter it. I don’t see why an enchantment that engulfs you in flames for no other reason than you attacked them was a good idea

CameronWoof:
> If your solution is short term then I'd rather you do nothing at all. Humans are meant to adapt, so do that instead.
> 
> These are not major game-breaking disasters, they are an edge mages get over their melee counterparts. Let them exist. Just because mages are different in a strong way does not mean they should be struck down.
> 
> You're going to remove these and it'll push people to explore other enchantments. Then they'll make these other enchantments and suddenly you're removing those too in a few months time.
> 
> Removals removals removals. Most of your balance prs are removals. I'd rather see creativity from you rather than removals. You're removing options in a game that benefits from having a huge list of options.

This is a preferable solution to "do nothing and hope that someone eventually fixes it". 

If you haven't seen Phoenix Guard in a while, good! I envy you - it means your rounds have been better than some of the ones experienced by myself and others. The problem is not necessarily phoenix guard itself - some damage and firestacks once a minute is certainly an effective enchantment, but far from overpowered. The issue is that you can enchant phoenix guard onto 5 different items, even multiple times on each item, and then when you get hit someone is immediately set to maximum firestacks. 

It's an oversight in the enchanting system that has been brought up repeatedly for months and has not been fixed. I have waited for the long-term solution, it is not forthcoming. This is the stopgap. Hopefully it encourages someone to fix the problem, and if it doesn't, then at least we can save people the agony of dealing with it.

GrayRachnid:
> > If your solution is short term then I'd rather you do nothing at all. Humans are meant to adapt, so do that instead.
> > These are not major game-breaking disasters, they are an edge mages get over their melee counterparts. Let them exist. Just because mages are different in a strong way does not mean they should be struck down.
> > You're going to remove these and it'll push people to explore other enchantments. Then they'll make these other enchantments and suddenly you're removing those too in a few months time.
> > Removals removals removals. Most of your balance prs are removals. I'd rather see creativity from you rather than removals. You're removing options in a game that benefits from having a huge list of options.
> 
> This is a preferable solution to "do nothing and hope that someone eventually fixes it".
> 
> If you haven't seen Phoenix Guard in a while, good! I envy you - it means your rounds have been better than some of the ones experienced by myself and others. The problem is not necessarily phoenix guard itself - some damage and firestacks once a minute is certainly an effective enchantment, but far from overpowered. The issue is that you can enchant phoenix guard onto 5 different items, even multiple times on each item, and then when you get hit someone is immediately set to maximum firestacks.
> 
> It's an oversight in the enchanting system that has been brought up repeatedly for months and has not been fixed. I have waited for the long-term solution, it is not forthcoming. This is the stopgap. Hopefully it encourages someone to fix the problem, and if it doesn't, then at least we can save people the agony of dealing with it.

You've stated the problem yourself, with the solution is right in front of your eyes.
it can be put on several items? Make it only be able to be put on one. You can limit the type of item that enchantments can be put onto (see featherstep).

With that information, I suggest locking phoenix/frostguard to either ring, neck, or shield (choose one!) enchantments only.

This change will make it so that you can only ever have 2 stacks of these enchantments on you at all times.

Better yet, you can go a step further with that change and make it so you can't stack two of the same enchantment on the same item. No double storage, no double featherstep, no double unbreaking. Instead you can do featherstep and longstriding, as an example.

The solution to most problems is sometimes written between the lines when you write out the problem. I think the suggestion I've made is a much more healthy, and not difficult, way to balance out phoenix/frostguard: by locking it to a single slot instead.

CameronWoof:
> Better yet, you can go a step further with that change and make it so you can't stack two of the same enchantment on the same item. No double storage, no double featherstep, no double unbreaking. Instead you can do featherstep and longstriding, as an example.
> 
> The solution to most problems is sometimes written between the lines when you write out the problem. I think the suggestion I've made is a much more healthy, and not difficult, way to balance out phoenix/frostguard: by locking it to a single slot instead.

If you feel like designing a solution, by all means! This would be preferable to disabling them entirely.

However, in several months of discussion, no one has stepped up to do that, and I'm tired of waiting and watching as people abuse the absolute shit out of the magic and enchantment systems to the detriment of other players.

GrayRachnid:
> > Better yet, you can go a step further with that change and make it so you can't stack two of the same enchantment on the same item. No double storage, no double featherstep, no double unbreaking. Instead you can do featherstep and longstriding, as an example.
> > The solution to most problems is sometimes written between the lines when you write out the problem. I think the suggestion I've made is a much more healthy, and not difficult, way to balance out phoenix/frostguard: by locking it to a single slot instead.
> 
> If you feel like designing a solution, by all means! This would be preferable to disabling them entirely.
> 
> However, in several months of discussion, no one has stepped up to do that, and I'm tired of waiting and watching as people abuse the absolute shit out of the magic and enchantment systems to the detriment of other players.

The solution I've given you takes the same amount of effort as removing them. Less, as it's a one line change. I am not going to develop for this codebase, but I am still playing the game it's running. If you can't accept my feedback because I refuse to code, then that's on you.

CameronWoof:
> > > Better yet, you can go a step further with that change and make it so you can't stack two of the same enchantment on the same item. No double storage, no double featherstep, no double unbreaking. Instead you can do featherstep and longstriding, as an example.
> > > The solution to most problems is sometimes written between the lines when you write out the problem. I think the suggestion I've made is a much more healthy, and not difficult, way to balance out phoenix/frostguard: by locking it to a single slot instead.
> > 
> > 
> > If you feel like designing a solution, by all means! This would be preferable to disabling them entirely.
> > However, in several months of discussion, no one has stepped up to do that, and I'm tired of waiting and watching as people abuse the absolute shit out of the magic and enchantment systems to the detriment of other players.
> 
> The solution I've given you takes the same amount of effort as removing them. Less, as it's a one line change. I am not going to develop for this codebase, but I am still playing the game it's running. If you can't accept my feedback because I refuse to code, then that's on you.

The solution of making it slot-specific would *help* but it would also not be a good solution. It's an impactful enchantment, I'd want it to be on chest armor, but that would still allow 4 stacks of it. 

The real solution is a combination of multiple things - you need to make sure you can't put multiple of the same enchantment on any one piece of gear **and** you need to make sure that even if you have multiple copies of Phoenix Guard, only one of them actually procs on hit (but all of them are put on cooldown). You could also disable enchanting worn equipment and then do a check to make sure no other piece of equipment has a matching enchantment with the unique tag on it before letting you equip it. 

I've thought about the problem, we know what the solutions are, what we lack are coders with the skill to implement them and the drive to actually do it. 

Until those people make themselves known, this is the next best thing.

GrayRachnid:
And I've offered you an easier solution that can happen with a one line change. You want a short term fix, I am giving you one that does not involving removing the enchantment. Lock it to rings. Lock it to shields. Lock it to bracers. Any of these three are soulful and make sense while also limiting it to just two enchantments.

Kick the bucket down the road, sure. Just don't remove a feature while doing so. Restrict it instead.

chemistrymain3:
Objective improvement. Uncounterable powergame utility enchantment that makes you invincible.

GrayRachnid:
> Objective improvement. Uncounterable powergame utility enchantment that makes you invincible.

Everything in this statement is factually incorrect other than the word enchantment. Amazing.

envoyin:
I remember when we removed Physician "spells" in 1.0 in promise of a complete alchemy rework to counter balance it. We never got the alchemy rework, until we entered 2.0 in which we rebased to AP. I hope this doesn't happen again.

RadiantFlash7:
Right uh. I'd rather not remove these. If anything, I'd rather just make a check in the signal calls so the effect only ever applies once, so it so players don't endlessly stack phoeniz or frostveil enchantments.

I'm not gonna close this PR, so I can have it up as a reminder to actually- DO it, but I'm putting my soon to be intended change out there

Xander3359:
I agree with GrayRachnid on this.
And if Radiant is up to make a proper fix, then he can close this once it's up.
In the meantime, I don't think these are in need of urgent removal to this extreme.

coffeecoffeecoffee:
> Right uh. I'd rather not remove these. If anything, I'd rather just make a check in the signal calls so the effect only ever applies once, so it so players don't endlessly stack phoeniz or frostveil enchantments.

This would be the perfect solution IMO.

maker-of-nightmares:
You could fix your key issue with it by changing it so it triggers _per entity_ once a minute or by making it so an entity gets a flag that says "Recently Guarded" so they don't get another stack. You could also make them only applicable to the chest slot or tabards or clothes(which some enchantments already make usage of to avoid stacking). I personally suggest having the guard restricted to either a Tabard or clothing slot items only the same way Featherstep and Darksight are Boot and Head slots only

PR #2584, opened 2026-09-03
Bulk Artificer Recipes + New Recipes

ophaq:
## About The Pull Request
- Adds bulk wooden cog, bulk bronze cog, golem exhibitor, and bodkin arrow recipes to artificer table's recipes.

## Testing Evidence
ArtificerRecipes

## Why It's Good For The Game
Less menu navigation and adding things that were missing is good. You need a TON of cogs for certain projects so the bulk recipes help shorten the time to make them. Also, we were missing bodkin arrows despite having blacksteel already in there. Did you know that the golem exhibitor was only on the anvil too for some reason? It still is but you can actually make it on the artificer table which makes more sense.

## Changelog
:cl:
add: Adds bulk wooden cog, bulk bronze cog, golem exhibitor, and bodkin arrow recipes to artificer table's recipes.
/:cl:

### DIFF
--- a/code/modules/roguetown/roguejobs/artificer/artificer_recipes/artificer_recipes.dm
+++ b/code/modules/roguetown/roguejobs/artificer/artificer_recipes/artificer_recipes.dm
@@ -86,6 +86,13 @@
 	hammers_per_item = 10
 	skill_level = 3
 
+/datum/artificer_recipe/wood/bulk
+	name = "Bulk Wooden Cogs 3x (+2 Planks)"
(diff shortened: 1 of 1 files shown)


annikaRU:
The bulk cogs should really be set to four instead of three for bronze ones, it'd cost less to make six than it would to make six in bulk

ophaq:
> The bulk cogs should really be set to four instead of three for bronze ones, it'd cost less to make six than it would to make six in bulk

It makes it x3 as in x3 the recipe. So it ends up as 6 cogs.

PR #2592, opened 2026-09-04
Decorative Mentor Hat/Mentor Robe/Flowery Robe

IronDragoon:
## About The Pull Request

Adds loadout and craftable decorative versions of these three items. They are identical other than that they have zero armor values, and as such can be used in the head/chest decoration slots. 

Also fixes the pontifex robe, a PR a month ago broke them and made them unable to be dyed so this just fixes that (it looks good)

## Testing Evidence

Tested, works just as expected

## Why It's Good For The Game

Lots of clothing has versions like this (especially for the loadout) that let players customise their look. Kazen stuff meanwhile only has 'armored' variants (but incredibly weak as loadout) so I was requested to add this and I agree. Variety is great

Obligatory image for the inevitable kazen haters:

kazengunapology

## Changelog

:cl:
add: loadout and craftable decorative versions of mentor hat, robe and flowery robe
/:cl:

### DIFF
--- a/code/modules/clothing/rogueclothes/armor/unique.dm
+++ b/code/modules/clothing/rogueclothes/armor/unique.dm
@@ -61,6 +61,12 @@
 	armor = ARMOR_LEATHER_STUDDED 
 	max_integrity = ARMOR_INT_CHEST_LIGHT_MEDIUM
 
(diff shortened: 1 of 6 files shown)


P-roxii:
Queen

PR #2597, opened 2026-09-05
the BUFF pierce weapons have been yearning for: changes penetrating integrity damage calculations

zeroisthebiggay:
## About The Pull Request

This PR buffs piercing weapons. It maybe even makes them usable compared to using a blunt or even just a sword. Moreover, it removes that odd feeling where, when your weapon force was high enough, you might've felt like you were breaking people's armor SLOWER than if it was lower.

Why is this? This is because ANY damage that penetrated ANY amount of armor drastically had its integrity damage reduced. The calculation goes like this. You add your damage with your AP. If you have more total effective 'damage' than your opponent does armor, it takes the excess and goes through. Let me give you an example.

Your opponent has forty cut armor on their torso. You have a sword that has thirty cutting damage and fifteen AP. This is effectively forty five damage. Forty five beats forty, obviously - so you penetrate. Five damage goes through to their body. Five damage is dealt to their armor.

In order to deal five damage to their body - and give them a TINY, microscopic bleed that'd take hours to die by, and has its bleed rate clamped because armor just Does this while it's not broken... you exchanged dealing what could've been forty five damage to their integrity, to instead deal five integrity damage and five damage to their body.

As expected, this is REALLY FUCKING BAD. Penetration weapons SUCK. ASS. FOR THIS SHITASS REASON!

Here's the new math. I can't be fucked to explain it again, so read these comments I put in the code to explain how it works. I wanted to actually die after spending two hours drooling at math.

```
Penetrative damage splits its total damage between armor and body, whatever damage armor protected from still lands on the armor and damages it.
AP weapons will break armor slower on average, but not "five intdamage per stab" slower as before.
		
I'm going to try and explain this to the best of my ability after banging my head against it.
AP effectively strips away armor. Forty five AP hitting fifty armor turns that to effectively five armor.
The remaining armor will reduce the damage of any incoming attack, so fifty damage is dampened to forty five.
Damage that got dampened is what gets dealt to the armor as intdamage.
A fifty damage attack is dampened to forty five, with five being dealt as intdamage.
That's really low, so we generate a number that's forty percent of your total force.

This number serves as an intdamage floor, we use it instead if your intdamage would've otherwise been lower than it.
If you get COMPLETELY penetrated, half of the damage will be dealt to the armor as well. Your gear shouldn't stay pristine after being totally cleaved.
 ```
 
 Besides that, a very minor change - higher AP means the bleeding clamp is higher through armor. If you have super high AP, you'll deal worse bleeds than someone who's barely able to poke past.

## Testing Evidence

I can't really do at-length testing of this. Maintainers will have to testmerge it and players will have to share how they feel about it, whether it's too strong or too weak. Here's a picture of me breaking a gambeson with a rapier.

image

## Why It's Good For The Game

Read the above paragraphs and you tell me this math is good. Stabbing weapons suck ass right now, and if you deal too much damage and start accidentally penning someone, your ability to kill them quickly by destroying their armor falls off a fucking CLIFF. I've had fights personally go on for way longer than they should've because my sword was TOO STRONG and couldn't destroy their gambeson because I was dealing single digit integrity damage with each slash.

## Changelog

:cl:
balance: redid the way integrity damage is calculated when piercing armor. the short of it is that you won't be dealing single digit integrity damage to armor for the crime of doing too much damage and penetrating it. spear and rapier users rejoice
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/human_defense.dm
+++ b/code/modules/mob/living/carbon/human/human_defense.dm
@@ -36,10 +36,15 @@
 			playsound(loc, get_armor_sound(used.blocksound, blade_dulling), 100)
 		var/intdamage = damage
 		var/consume_debuff = TRUE
-		// Penetrative damage deals significantly less to the armor. Tentative.
+		// Penetrative damage splits its total damage between armor and body,
+		// whatever damage armor protected from still lands on the armor and damages it.
+		// AP weapons will break armor slower on average, but not "five intdamage per stab" slower as before.
 		if((damage + armor_penetration) > protection && d_type != "blunt")
 			consume_debuff = FALSE
 			intdamage = (damage + armor_penetration) - protection
+			intdamage = protection - armor_penetration
+			if(armor_penetration >= protection) // If you get COMPLETELY penetrated, half of the damage will be dealt to the armor as well. Your gear shouldn't stay pristine after being totally cleaved.
+				intdamage = damage / 2
 		if(intdamfactor != 1)
 			intdamage *= intdamfactor
 		if(d_type == "blunt")
--- a/code/modules/mob/living/carbon/human/human_defense.dm
+++ b/code/modules/mob/living/carbon/human/human_defense.dm
@@ -41,8 +41,14 @@
 		// AP weapons will break armor slower on average, but not "five intdamage per stab" slower as before.
 		if((damage + armor_penetration) > protection && d_type != "blunt")
 			consume_debuff = FALSE
-			intdamage = (damage + armor_penetration) - protection
-			intdamage = protection - armor_penetration
+			// WE TAUGHT THIS MONKEY ARMOR PIERCING CODE AND IT KILLED ITSELF IMMEDIATELY.
+			// I'm going to try and explain this to the best of my ability after banging my head against it.
+			// AP effectively strips away armor. Forty five AP hitting fifty armor turns that to effectively five armor.
+			// The remaining armor will reduce the damage of any incoming attack, so fifty damage is dampened to forty five.
+			// Damage that got dampened is what gets dealt to the armor as intdamage.
+			// A fifty damage attack is dampened to forty five, with five being dealt as intdamage.
+			// That's really low, so we generate a number that's forty percent of your total force.
+			intdamage = max(protection - armor_penetration, damage * 0.4) // This number serves as an intdamage floor, we use it instead if your intdamage would've otherwise been lower than it.
 			if(armor_penetration >= protection) // If you get COMPLETELY penetrated, half of the damage will be dealt to the armor as well. Your gear shouldn't stay pristine after being totally cleaved.
 				intdamage = damage / 2
 		if(intdamfactor != 1)
--- a/code/modules/mob/living/carbon/human/species.dm
+++ b/code/modules/mob/living/carbon/human/species.dm
@@ -1943,7 +1943,8 @@ GLOBAL_LIST_EMPTY(roundstart_races)
 					H.apply_damage(Iforce * user.used_intent.blunt_chip_strength, BRUTE, def_zone, blunt_chip_block)//, spread_damage = TRUE)
 					H.next_attack_msg += " <span class='warning'>and yet the force punches through!</span>"//But sometimes it lies!
 		if(!nodmg)
-			var/datum/wound/crit_wound = affecting.bodypart_attacked_by(user.used_intent.blade_class, (Iforce * weakness) * ((100-(armor_block+armor))/100), user, selzone, crit_message = TRUE, weapon = I, armor_penetration = pen)			if(should_embed_weapon(crit_wound, I))
+			var/datum/wound/crit_wound = affecting.bodypart_attacked_by(user.used_intent.blade_class, (Iforce * weakness) * ((100-(armor_block+armor))/100), user, selzone, crit_message = TRUE, weapon = I, armor_penetration = pen)
+			if(should_embed_weapon(crit_wound, I))
 				var/can_impale = TRUE
 				if(!affecting)
 					can_impale = FALSE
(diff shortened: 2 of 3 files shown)


zeroisthebiggay:
oops

RadiantFlash7:
Piercing alone isn't actually bad presently. You can't artery crit someone through armor, it's true, but it's VERY easy to bleed someone out in a fight ever since we nerfed CON to matter less for bleed rate.

That said, theres no reason to punish players for dealing maximum pierce, or for partial pierce not properly damaging armor. The armor change you've pressed here is probably how 90% of the server thought it was already functioning.

zeroisthebiggay:
> Piercing alone isn't actually bad presently. You can't artery crit someone through armor, it's true, but it's VERY easy to bleed someone out in a fight ever since we nerfed CON to matter less for bleed rate.

It'll be slower than if you had just a regular weapon, definitely. God forbid they drink red, since you can't inflict any real wounds, either.

coffeecoffeecoffee:
> Piercing alone isn't actually bad presently. You can't artery crit someone through armor, it's true, but it's VERY easy to bleed someone out in a fight ever since we nerfed CON to matter less for bleed rate.
> 
> That said, theres no reason to punish players for dealing maximum pierce, or for partial pierce not properly damaging armor. The armor change you've pressed here is probably how 90% of the server thought it was already functioning.

It is actually REALLY REALLY TERRIBLE. Nobody takes the eagle's beak or similar polearms for this reason in my experience as MAA and Merc. Everybody is using the greatswords and other weapons for a reason!

Tekkomeka:
If I got it right, I think new math is terrible, change it please.

Make it so that only your damage goes to integrity, not damage + ap. Because, for example, look how damage will work for high STR gnolls then (and for any other high str high pen intents weapon builds):

Before this PR:
100 absoprtion plate.
60 armor pen.
45 damage.
---
5 integrity damage, 5 body damage.

With this PR:
100 absoprtion plate.
60 armor pen.
45 damage.
---
100 integrity damage, 5 body damage.

What I want it to be:
100 absoprtion plate.
60 armor pen.
45 damage.
---
45 integrity damage, 5 body damage.


Otherwise armor pen weapons will be superior and we will be again motivated to start minmaxing damage. That's not good.


UPD:

Nevermind, your explanation is a bit wrong. I've just tested it and no matter what I kept doing same amount of damage to integrity, which is 
image

100 (plate absorpiton) - 40 (great axe pen) = 60 integrity damage.

> That previous example now means that five damage goes through, and forty damage is dealt as integrity damage. Math tentative. Might need reduction.

That is simply not true. It would be 40-15 = 25 integrity damage, not 40.

Why not just make integrity damage equal to your force damage when you pierce armor?

zeroisthebiggay:
> If I got it right, I think new math is terrible, change it please.
> 
> Make it so that only your damage goes to integrity, not damage + ap. Because, for example, look how damage will work for high STR gnolls then (and for any other high str high pen intents weapon builds):
> 
> ## Before this PR:
> 100 absoprtion plate.
> 60 armor pen.
> 45 damage.
> 5 integrity damage, 5 body damage.
> 
> ## With this PR:
> 100 absoprtion plate.
> 60 armor pen.
> 45 damage.
> 100 integrity damage, 5 body damage.
> 
> ## What I want it to be:
> 100 absoprtion plate.
> 60 armor pen.
> 45 damage.
> 45 integrity damage, 5 body damage.
> 
> Otherwise armor pen weapons will be superior and we will be again motivated to start minmaxing damage. That's not good.
> 
> UPD:
> 
> Nevermind, your explanation is a bit wrong. I've just tested it and no matter what I kept doing same amount of damage to integrity, which is image
> 
> 100 (plate absorpiton) - 40 (great axe pen) = 60 integrity damage.
> 
> > That previous example now means that five damage goes through, and forty damage is dealt as integrity damage. Math tentative. Might need reduction.
> 
> That is simply not true. It would be 40-15 = 25 integrity damage, not 40.
> 

I'm Really bad with explaining math and it was super late at night. I'll have to take another look at my explanations so they're more coherent.

> Why not just make integrity damage equal to your force damage when you pierce armor?

I considered it, but I was afraid if you just dealt your force whenever you pierced armor, it'd make penetration weapons too good, since you're dealing full integrity damage in addition to piercing them.

CameronWoof:
The scheme I think makes the most intuitive sense:

Force + AP determines if your attack penetrates an armor's protection value.

If you penetrate, you deal damage to flesh equal to the difference between your Force + AP and the protection value.

If you penetrate, you deal damage to integrity equal to 70% of the initial attack's force OR the the initial attack's force minus the amount of force passed to flesh, whichever is higher. (The goal here is to ensure that while penetrating armor, you are still making headway towards breaking armor and opening your opponent up to criticals BUT you aren't dealing essentially double damage by penetrating to flesh and dealing the same tamage to armor integrity, too.)

If you don't penetrate, you deal damage to integrity equal to the initial attack's force.

zeroisthebiggay:
> The scheme I think makes the most intuitive sense:
> 
> Force + AP determines if your attack penetrates an armor's protection value.
> 
> If you penetrate, you deal damage to flesh equal to the difference between your Force + AP and the protection value.
> 
> If you penetrate, you deal damage to integrity equal to 70% of the initial attack's force OR the the initial attack's force minus the amount of force passed to flesh, whichever is higher. (The goal here is to ensure that while penetrating armor, you are still making headway towards breaking armor and opening your opponent up to criticals BUT you aren't dealing essentially double damage by penetrating to flesh and dealing the same tamage to armor integrity, too.)
> 
> If you don't penetrate, you deal damage to integrity equal to the initial attack's force.

I can consider adding a clamp, but I don't think there's many weapons that deal the majority of their damage through fully piercing armor. I don't know because spears are so janky to use compared to every single other weapon that I'm not experienced in their actual performance.

chemistrymain3:
> > The scheme I think makes the most intuitive sense:
> > Force + AP determines if your attack penetrates an armor's protection value.
> > If you penetrate, you deal damage to flesh equal to the difference between your Force + AP and the protection value.
> > If you penetrate, you deal damage to integrity equal to 70% of the initial attack's force OR the the initial attack's force minus the amount of force passed to flesh, whichever is higher. (The goal here is to ensure that while penetrating armor, you are still making headway towards breaking armor and opening your opponent up to criticals BUT you aren't dealing essentially double damage by penetrating to flesh and dealing the same tamage to armor integrity, too.)
> > If you don't penetrate, you deal damage to integrity equal to the initial attack's force.
> 
> I can consider adding a clamp, but I don't think there's many weapons that deal the majority of their damage through fully piercing armor. I don't know because spears are so janky to use compared to every single other weapon that I'm not experienced in their actual performance.

There are none.

PR #1553, opened 2026-05-05
Sneakmageddon 2

Vizoplez:
## About The Pull Request
PARTIAL PORT OF:
https://github.com/Rotwood-Vale/Ratwood-Keep/pull/1228
DESIGNED TO MAKE SNEAKING LESS ABUSABLE ON BOTH SIDES.

Also fixes a few miscellaneous bugs, such as; Amulets on feet causing runtimes
Objects on feet causing runtimes

## Testing Evidence
https://github.com/user-attachments/assets/a7aedc9a-8af1-4892-889e-9204d2891024

https://github.com/user-attachments/assets/cae3f310-9ac8-4634-a6a3-4725e796e2a9

You can see pretty clearly that the system works, it entirely hides players from other clients unless they have a high enough perception, or if they bump into the sneaked player directly - and even then, it barely lasts 1.5 seconds until sneaking is re-established.

This is not a significant nerf to sneaking, rather an alternate system introduced to make the current one less painful for staff, players, garrison and retinue, AND antags.

It does however nerf some specific aspects of sneaking that made no sense from a roleplay and gameplay standard, like bumping into them not de-sneaking them.
>

## Why It's Good For The Game
100% consequence free stealth moment (its no longer such)
It nerfs being able to sit in perfect darkness as a zizoite wretch and continually snuff lights to attack people without consequence or care.
It fixes a few other things, such as being able to point at sneaking people to find their location, being able to examine them, etc.
It fixes being unable to counter them (thrown rocks/etc now de-stealth them too, same as hitting them!)
Technical gnoll buff to prevent more stupid shit like people pointing at them to find where they are.

## Changelog

:cl:
fix: Amulets causing a ton of runtimes in foot slot
fix: Being able to point at, examine, and see invisible sneak users via alt-rightclicking
balance: Gnoll system moved to use invisibility atop of alpha=0
balance: PER now allows for passively spotting sneaked players. Ratio may be changed
add: Invisi-Sneaking (No more metagaming via pointing, guardseman!)
balance: You are now de-sneaked for exactly 4 seconds upon attacking any object, or player. This does NOT affect grappling, to note, so kidnapping is still viable!
balance: You are now de-sneaked for exactly 0.5 seconds upon someone bumping into you.

/:cl:

PR #2598, opened 2026-09-05
Undoes an unintended balance change related to gnoll stealth discoverability

CameronWoof:
## About The Pull Request

Alright so #2540 added journeyman Stealth to all gnolls, and suddenly it became impossible to detect gnolls in Stalk. What happened? 

Well, months back, in #1553 a bonus was added to Stalk's *invisibility level*, which consisted of a flat bonus and a multiplier based on their stealth skill (which at the time was zero). This meant that in order to see gnolls, you had to have at least Expert Tracking, win a probability check, and ALSO have a minimum of 14 PER. 

Since gnolls now have Stealth skill, that requirement got bumped up to **18** PER, which vanishingly few people have. This PR restores the previous balance before the recent addition of stealth skills to gnolls by removing the bonus from their stealth skill.

## Why It's Good For The Game

An avalanche of reports that gnoll stealth was "bugged" turned out to just be an oversight and a weird interaction between two PRs with unintended consequences. This returns the game to the previous state.

## Changelog

:cl: CameronWoof
fix: Gnolls should now be similarly difficult to reveal from stealth as before.
/:cl:


zoktiik:
OUUUUGGGGHHH

PR #2603, opened 2026-09-05
Update antagonist.dm

Kroto-Ker:
linters.


Xander3359:
Is... the file not in LF?

PR #2607, opened 2026-09-05
Vamps don't ash on death

chemistrymain3:
## About The Pull Request

See title.

## Testing Evidence

I didn't bother to test it because some guy said that testing things for bugs isn't possible and I'm salty. It works, though. 
## Why It's Good For The Game

Vampires turning to dust on death might be thematic but unfortunately it kind of sucks for the vampire and also sucks that there's a conversion antagonist that functionally RR's you for clicking 'yes' to being converted and thus guarantees most people will click no. Since we seem to be moving away from RR as a consequence for dying to an antagonist, I don't particularly see why we shouldn't move in the direction of not being RR'd just for dying in general. Hopefully it encourages more people to click 'yes' to being converted.

## Changelog

:cl:
balance: Vampires no longer turn to ash on dying.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/death.dm
+++ b/code/modules/mob/living/carbon/human/death.dm
@@ -56,11 +56,6 @@
 		SSdroning.kill_rain(client)
 
 	if(mind)
-		if(!gibbed)
-			var/datum/antagonist/vampire/VD = mind.has_antag_datum(/datum/antagonist/vampire)
-			if(VD)
-				dust(just_ash=TRUE,drop_items=TRUE)
-				return
 
 		var/datum/antagonist/lich/L = mind.has_antag_datum(/datum/antagonist/lich)
 		if (L && !L.out_of_lives)


Gug1234:
The better solution to this is something like https://github.com/Azure-Peak/Azure-Peak/pull/7431, where the sunder crits aren't a one-shot but put vampires into a state of potential submission.

Given how powerful vampires are and how they can trivially repair literally every wound save decapitations and sunders, the dust is necessary as their balancing lever.

Sometimes people need to die; this is something we shouldn't be allergic to, otherwise you have weird ludonarrative dissonance where the OOC expectation to prevent anyone from being RR'd trumps the IC motivations to make sure Lord Dracula, scourge of the living, stays dead and buried.

chemistrymain3:
It's very clearly not necessary. Also, the AP solution removes ashing on death from everyone except the VL anyway.

https://github.com/Azure-Peak/Azure-Peak/pull/4600

Nobody but the VL on AP ashes on death, for very similar reasons.

Frankly, the excuse to me that vampire in general needs to ash on death as its 'balancing lever' does not hold any water. They're powerful if methuseleh or vampire lord, otherwise they can't access most of their strong abilities. They can heal off all wounds that aren't cutting a limb off or killing them, but frankly, most fights end in killing them OR skullcracking and chaining them, which they cannot escape from.

And frankly, I see no reason to lean towards 'We need more death everything is too hugboxed' but only for antags. Only the antag dies but good guys never die is stupid and unhealthy for the game. Choosing to be converted actively PUNISHING you for being converted by making you unrecoverable is also bad.

CameronWoof:
What actually happens when you kill a vampire with this change? Do they get lyfe-ended and die like other players?

Xander3359:
Citing https://github.com/Rotwood-Vale/Ratwood-2.0/pull/1225

The reason this remains is because we don't yet have stakes that you can use to dust a vamp.
\+ You have to make it where a dead vamp forcibly drops their disguise

UntoldTactics:
I concur with this PR but I would also say azures form of sunderance is better, and wish we had that. That doesn't stop regular kills dusting current vampires ofc, but yeah. It may also be worth looking into fixing/better implementing the deconversions? I'm pretty sure absolver and the silverblood poultice still don't work to cure vampires fully, especially the formers ritual should be working if we want to prioritize deconversions as a possibility.

IIRC, being DNR on AP also makes you dustable since you're DNR anyway. Otherwise, I don't see this PR being all that bad. I do concur with the disguise drop stuff.

I also wish sunder wasn't such a mess but it's probably beyond us in regards to making sunder not transfer 5 billion stacks into mortals and making them bonfires. (Sunder being one of the big things that fucks with vampires but also ends up sabotaging the people fighting them.)

chemistrymain3:
> Citing #1225
> 
> The reason this remains is because we don't yet have stakes that you can use to dust a vamp. + You have to make it where a dead vamp forcibly drops their disguise

I'll make it so that a dead vampire drops their disguise. I don't see the reason we'd add dusting, though? If you want to RR someone, just cut their head off or bury them.

Also, I don't see the logic of literally like anything anyone was saying on the previous PR. It's not true. Vampires aren't super strong unless they're the VL or Methuseleh, AKA the spawn roundstart antag vamps. Their sirelings are much weaker. They don't get infinite energy, they get infinite blue.

And the idea that they just like, 'can come back'- Yeah, that's the point. If you revive them. As with any antagonist. Or any person.

UntoldTactics:
> > Citing #1225
> > The reason this remains is because we don't yet have stakes that you can use to dust a vamp. + You have to make it where a dead vamp forcibly drops their disguise
> 
> I'll make it so that a dead vampire drops their disguise. I don't see the reason we'd add dusting, though? If you want to RR someone, just cut their head off or bury them.
> 
> Also, I don't see the logic of literally like anything anyone was saying on the previous PR. It's not true. Vampires aren't super strong unless they're the VL or Methuseleh, AKA the spawn roundstart antag vamps. Their sirelings are much weaker. They don't get infinite energy, they get infinite blue.
> 
> And the idea that they just like, 'can come back'- Yeah, that's the point. If you revive them. As with any antagonist. Or any person.

Note that tennite revivals just straightup gib undead. You'd need to do a lux revival (assuming those work on vampires), the funny chair or an actual ascendant revival. TLDR; It'd be pretty hard to mistakenly revive a vampire.

RadiantFlash7:
I'm going to close this for now. Per Xander, we want to make it so ashing vampires happens when a vampire is staked. Removing ashing all together, not so much.

PR #2622, opened 2026-09-06
Precise tievery

wikimody:
## About The Pull Request
Stealing from somone with left arm selected will steal from left back slot, right from right slot
Stealing from the chest now steals from cloak instead

## Testing Evidence
Zrzut ekranu 2026-09-06 180740

## Why It's Good For The Game
Having to roll 50 50 betwen instantly giving yourself away by stealing lets say shield (very common back item) or stealing from the bag isnt particularly plesant. Arms were also unused zones for tieving and legs gave this precise version alredy for belt stealing, hence me coppying over the logic. 
Cloaks all have storage, coudl be resonably stolen from and the chest target was free, so i added them as posible target 

## Changelog

:cl:
add: Pickpocketing while targeting an arm now pickpockets coresponding back zone (no longer will you roll a dice to see if you pick a shield instead)
add: Pickpocketing chest now pickpockets cloak instead
/:cl:

### DIFF
--- a/code/game/objects/items/rogueweapons/mmb/steal.dm
+++ b/code/game/objects/items/rogueweapons/mmb/steal.dm
@@ -180,7 +180,7 @@
 	if(!isnum(range_add))
 		range_add = 0
 	var/steal_radius = 1 + range_add
-	var/list/stealablezones = list("chest", "neck", "groin", "r_hand", "l_hand", "r_leg", "l_leg")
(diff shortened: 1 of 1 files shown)


ophaq:
Everything else is cool but I would like it if cloaks were untouched.

wikimody:
> Everything else is cool but I would like it if cloaks were untouched.

I generally hold to belief that things shouldn't be immune to mechanics unless there is good reason for them to be. Why should cloaks be immune to pickpocketing?

yanlucca:
Good one, boss.

PR #2624, opened 2026-09-06
Teeny Tiny Vanguard Adjustment

SirSwagmeyer:
## About The Pull Request

Vanguard Footman
- Expert Wrestling -> Jman
- +2 STR -> +1 STR, +2 WIL, +2 PER

Vanguard Archer
- Expert Athletics -> Jman
- Jman Climbing -> Expert

## Testing Evidence

Literally just line changes. It compiles. 

## Why It's Good For The Game

Much-requested changes that nobody has bothered to implement. 

It's difficult to understand the justifications behind giving Vanguard the ability to hit the Strength softcap, or to give them expert wrestling _on top of that._ 

Vanguard archers deserve a bit of a treat in their climbing department, which was something people were lamenting about. I've knocked down their athletics skill a level to compensate for the fact they can climb just about any wall. 

## Changelog

:cl:

balance: tweaked vanguard stats and skills just slightly. 

/:cl:

### DIFF
--- a/code/modules/jobs/job_types/roguetown/garrison/vanguard.dm
+++ b/code/modules/jobs/job_types/roguetown/garrison/vanguard.dm
@@ -76,8 +76,8 @@
 		/datum/skill/combat/knives = 3,
 		/datum/skill/combat/axes = 2,
 		/datum/skill/combat/swords = 2,
-		/datum/skill/misc/athletics = 4,
-		/datum/skill/misc/climbing = 3,
+		/datum/skill/misc/athletics = 3,
+		/datum/skill/misc/climbing = 4,
 		/datum/skill/misc/sneaking = 4,
 		/datum/skill/misc/swimming = 2,
 		/datum/skill/misc/medicine = 1,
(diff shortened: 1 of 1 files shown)


Tzi23:
I personally think it'd be better to keep the +2 STR instead of change it to a +6 in weighted stats myself. Dumb brute strength for a levy makes sense to me, but the PER buff for Footman isn't necessary. 

I agree on the rest. Removing expert wrestling would be great, and I like the Archer changes too.

SirSwagmeyer:
> I personally think it'd be better to keep the +2 STR instead of change it to a +6 in weighted stats myself. Dumb brute strength for a levy makes sense to me, but the PER buff for Footman isn't necessary.
> 
> I agree on the rest. Removing expert wrestling would be great, and I like the Archer changes too.

My vision is that a Levy does not rely purely on brute force - that feels more like a city watchman's way of fighting.

The levy is a scrappy, conscripted or volunteered soldier. They're taught the basics and should _ideally_ rely on a degree of dirty tricks and cunning to make it out alive. 

And, in all honesty, we _really_ don't need another role that's able to hit the STR softcap so readily. Even the Master Warden doesn't get +2 strength.

chemistrymain3:
To be honest, with journeyman skills, 2 STR isn't very significant. Even with struggler or muscular, it just doesn't matter as much.

STR only matters if you hit. Jman means you usually cannot. It mattered for wrestle vanguards because they could actually land the grapples. Now you can break out.

Ambivalent either way.

HalfAwake691:
Hitler dead

SirSwagmeyer:
> To be honest, with journeyman skills, 2 STR isn't very significant. Even with struggler or muscular, it just doesn't matter as much.
> 
> STR only matters if you hit. Jman means you usually cannot. It mattered for wrestle vanguards because they could actually land the grapples. Now you can break out.
> 
> Ambivalent either way.

My concerns with +2 STR is more regarding the fact they _have jman skills in maces, spears, swords, and axes_ - it isn't a rare occurrence to see a Vanguard take blacksmith apprentice and run around with a double-headed steel greataxe, a blacksteel flamberge, or a halberd by the first hour of the round. I don't know if that's something I want to see vanguards doing on such a regular basis.

PR #2618, opened 2026-09-06
THAT'S A Knife - New Ascendant Daggers for Heretics, Additional Daggers Overhauled

SirSwagmeyer:
## About The Pull Request

image

New Ascendant daggers with their own special little intents:
- Zizo's dagger inflicts shallower wounds, but cuts twice as quickly. Don't let the wounds build up.
- Graggar's dagger has a rend intent with additional damage. 
- Matthios' dagger can be thrown like a tossblade; it can't be picked up and thrown back at you, unless they're also a Matthiosan.
- Baotha's dagger thrusts quicker.

Adds Rondel Daggers, Kampfmessers, Kukris, and Kris Daggers into the game. Each have their own unique functionality to them, from stabbing/cutting quicker at the exchange of a lack of one intent or the other, or the ability to viciously tear through flesh.

Desert Riders spawn with kukris. Desert Town garrison does as well.

Also changed up the helmets for heretics - Graggar gets a volf helm, Noc gets a hounskull.

## Testing Evidence

image
image
image
image

## Why It's Good For The Game

Current collection of knives are pretty bland. Felt they deserved some love. 

## Changelog

:cl:
add: Added a variety of new daggers into the game, including the kampfmesser, the rondel, kris, and kukri.

/:cl:

PR #2627, opened 2026-09-06
minor aimed/strong/swift refactor

zeroisthebiggay:
## About The Pull Request

Deletes a lot of duplicated code relating to how these intents do click CD. Wires a lot of it through the same procs instead of using duplicated math.

Fixes a few bugs, namely:
- Swift can now be used to swift objects, with appropriate stamina drain.
- Strong now correctly drains stamina when used against objects.
- Apparently strong correctly did increased damage against structures already. I just assumed it didn't because it never used any extra stamina.
- If a weapon somehow gets lower click CD than six, the swift intent floor of six won't increase your click CD, it just won't change. No weapons currently exist with a click CD lower than six, but futureproofing is good.

## Testing Evidence

I can't really show this without taking a video and I don't wanna, but just know I was swifting the fuck out of windows in my test world and it drained my stamina a lot and everything.

## Why It's Good For The Game

Bugs and futureproofing.

## Changelog

:cl:
fix: you can swift intent objects
fix: strong intenting objects now drains stamina
/:cl:

### DIFF
--- a/code/_onclick/click.dm
+++ b/code/_onclick/click.dm
@@ -25,6 +25,21 @@
 /mob/proc/changeNext_move(num, hand, override = FALSE)
 	next_move = world.time + ((num+next_move_adjust)*next_move_modifier)
 
+/mob/proc/get_rmb_clickcd(base_clickcd)
+	var/adf = base_clickcd
+	if(istype(rmb_intent, /datum/rmb_intent/aimed))
+		adf = round(adf * CLICK_CD_MOD_AIMED)
+	else if(istype(rmb_intent, /datum/rmb_intent/swift))
+		adf = min(adf, max(round(adf * CLICK_CD_MOD_SWIFT), CLICK_CD_INTENTCAP))
+	return adf
+
+/mob/proc/_get_swifstrong_stam_penalty()
+	if(istype(rmb_intent, /datum/rmb_intent/strong))
+		return EXTRA_STAMDRAIN_SWIFSTRONG
+	if(istype(rmb_intent, /datum/rmb_intent/swift) && used_intent.clickcd > CLICK_CD_INTENTCAP)
+		return EXTRA_STAMDRAIN_SWIFSTRONG
+	return 0
+
 /mob/living/changeNext_move(num, hand, override = FALSE)
 	var/mod = next_move_modifier
 	var/adj = next_move_adjust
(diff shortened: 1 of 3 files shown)


Venithefox:
you should lower the swift intent floor slightly to make room for the rondel dagger and baothan dagger in https://github.com/Rotwood-Vale/Ratwood-2.0/pull/2618 to work properly on swift intent as currently they attack too fast for swift intent to function correctly

zeroisthebiggay:
> you should lower the swift intent floor slightly to make room for the rondel dagger and baothan dagger in #2618 to work properly on swift intent as currently they attack too fast for swift intent to function correctly

That's initially why this PR was made, to fix the bug where swift intent would actually make the dagger slower because it had five CD on its intent. I don't think a dagger that has an intent that's already faster than swift intent needs to work with swift; if a weapon is too fast for swift, swift intent just does nothing.

Venithefox:
> > you should lower the swift intent floor slightly to make room for the rondel dagger and baothan dagger in #2618 to work properly on swift intent as currently they attack too fast for swift intent to function correctly
> 
> That's initially why this PR was made, to fix the bug where swift intent would actually make the dagger slower because it had five CD on its intent. I don't think a dagger that has an intent that's already faster than swift intent needs to work with swift; if a weapon is too fast for swift, swift intent just does nothing.

it's more so just to make always running strong intent with the daggers less appealing. even if it was only a little bit faster from swift, it's better to have a useless gimmick with variety (as you will stam tf out if you use this on swift) than only limit it to strong intent (because your ass is NOT using defend for its passive on a dagger)

zeroisthebiggay:
> it's more so just to make always running strong intent with the daggers less appealing

lol. lmao.

Out of scope regardless. This is a refactor, not a balance PR.

code/_onclick/click.dm
```diff
@@ -25,6 +25,21 @@
...
 
+/mob/proc/get_rmb_clickcd(base_clickcd)
+	var/adf = base_clickcd
+	if(istype(rmb_intent, /datum/rmb_intent/aimed))
+		adf = round(adf * CLICK_CD_MOD_AIMED)
+	else if(istype(rmb_intent, /datum/rmb_intent/swift))
+		adf = min(adf, max(round(adf * CLICK_CD_MOD_SWIFT), CLICK_CD_INTENTCAP))
+	return adf
+
+/mob/proc/_get_swifstrong_stam_penalty()
```
Xander3359: why is this a private proc?

zeroisthebiggay: Private? Do you mean the underscore preceding it? That's my bad. I'll remove it but if I'm misinterpreting let me know.

PR #2631, opened 2026-09-07
Minor changes to second wind (abysor T0)

wikimody:
## About The Pull Request
Second wind incantation has been changed from shout to emote
Knockdown is now clensed
Immbolise and of balance are clensed instead of blocking the cast
Paralisis now blocks the cast (stun still blocks it to)

## Testing Evidence
Zrzut ekranu 2026-09-07 021749

Zrzut ekranu 2026-09-07 021822

Zrzut ekranu 2026-09-07 021915

## Why It's Good For The Game
Speaking incantation stoped the spell from being used while downed in watter. At least for me it feels like gods of the sea stand up spell should be usable in sea
Knockdown not being removed means you coudl have stood up and then instantly fallen down because of the status. Paralisis not stoping the spell is an issue for the same reson.
It feels like immbolise and of balance are the kind of things this spell shoudl help you against instead of stoping you from casting. Unlike garagorite stuns and paralise arent cleansed, so it is meaningfully worse then the ascendand version

## Changelog

:cl:
add: Second wind (abyssors T0) has now emote incantation (so usable when under water)
balance: some mobility reducing  effects are now clensed instead of stoping the cast of second wind
fix: You should no longer have to worry about instantly falling down after using the second wind
/:cl:

### DIFF
--- a/code/modules/spells/roguetown/acolyte/abyssor.dm
+++ b/code/modules/spells/roguetown/acolyte/abyssor.dm
@@ -87,8 +87,8 @@
 	sound = 'sound/magic/abyssor_splash.ogg'
 	associated_skill = /datum/skill/magic/holy
 	antimagic_allowed = FALSE
-	invocations = list("What is drowned shall rise anew!")
-	invocation_type = "shout"
+	invocations = list("is lifted up by spectral waves!")
+	invocation_type = "emote"
 	recharge_time = 120 SECONDS
 	devotion_cost = 30
 	miracle = TRUE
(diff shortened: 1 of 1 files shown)


Xander3359:
Changelog should be formatted in a way that players know exactly what changed when it's visible to the players.

`You should no longer have to worry about instantly falling down after using the miracle`

What miracle? What does this mean? Make sure it explains what it's changing.

wikimody:
I assumed the reference to second wind in the first line will make it obvious for the next two lines, but fair enough I will change it

Changed

PR #2638, opened 2026-09-07
Unsettling Vice (fixed)

MrShibaa:
## About The Pull Request

Re-adds Unsettling beauty in the form of a different vice UNSETTLING, which always applies a negative mood effect towards people who examine you.

## Testing Evidence

image

image

image

image

image

## Why It's Good For The Game

Allows people to express how uncanny or unnerving their characters are.. And is just soulful flavour altogether while always applying a negative status effect on those examining you..

## Changelog
:cl:
add: Unsettling Beauty back as "Unsettling" Vice
change: Makes it so it always gives the examiner stress..
change: Ugly now conflicts with unsettling


zeroisthebiggay:
Should it be mutually exclusive with ugly? Both to prevent mood dunking people who examine you and because someone looking unappealing is usually either in an uncanny way or an ugly way.

MrShibaa:
> Should it be mutually exclusive with ugly? Both to prevent mood dunking people who examine you and because someone looking unappealing is usually either in an uncanny way or an ugly way.

To address this, I've made it so ugly now conflicts with unsettling flaw, so now you can only have either the ugly or the unsettling flaw.. Not both.

zeroisthebiggay:
> > Should it be mutually exclusive with ugly? Both to prevent mood dunking people who examine you and because someone looking unappealing is usually either in an uncanny way or an ugly way.
> 
> To address this, I've made it so ugly now conflicts with unsettling flaw, so now you can only have either the ugly or the unsettling flaw.. Not both.

Any the ugly virtue?

PR #2646, opened 2026-09-08
[Admin] AFK timer and admin notification, ghost setting

IronDragoon:
## About The Pull Request

Time since being disconnected is now tracked and visible, upon reaching fifteen minutes it will trigger an admin notification, and a prompt to allow admins to quickly far travel in these situations (if required).

Also lets admins set a custom ghost sprite which is saved in prefs now, to prevent it needing to be done on every instance of ghosting

## Testing Evidence

image

image

## Why It's Good For The Game

Requested admin tools

## Changelog

:cl:
add: Added examine text to show how long someone was disconnected, and admin warning at certain times
admin: lets admins permanently set ghost sprite
/:cl:

### DIFF
--- a/code/modules/mob/living/ssd_indicator.dm
+++ b/code/modules/mob/living/ssd_indicator.dm
@@ -1,6 +1,9 @@
 GLOBAL_DATUM_INIT(ssd_indicator, /mutable_appearance, mutable_appearance('icons/mob/ssd_indicator.dmi', "default0", FLY_LAYER))
+GLOBAL_LIST_INIT(disconnected_admin_alert_role_times, list(
+	// "Grand Duke" = 10 MINUTES,
+))
 
-#define DISCONNECTED_ADMIN_ALERT_TIME 15 MINUTES
+#define DEFAULT_DISCONNECTED_ADMIN_ALERT_TIME 15 MINUTES
 
 /mob/living/proc/set_ssd_indicator(state)
 	if(state && stat != DEAD)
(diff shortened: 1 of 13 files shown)


code/modules/mob/living/ssd_indicator.dm
```diff
@@ -1,4 +1,9 @@
 GLOBAL_DATUM_INIT(ssd_indicator, /mutable_appearance, mutable_appearance('icons/mob/ssd_indicator.dmi', "default0", FLY_LAYER))
+GLOBAL_LIST_INIT(disconnected_admin_alert_role_times, list(
+	// "Grand Duke" = 10 MINUTES,
+))
```
Xander3359, line 4: ??

code/modules/mob/living/ssd_indicator.dm
```diff
@@ -7,3 +12,51 @@ GLOBAL_DATUM_INIT(ssd_indicator, /mutable_appearance, mutable_appearance('icons/
...
+	if(!ishuman(src))
+		return
+	disconnected_admin_alert_sent = TRUE
+	var/fartravel_link = "(<a href='?_src_=holder;[HrefToken(TRUE)];ssd_sendbacktolobby=[REF(src)]'>Fartravel</a>)"
+	message_admins(span_adminnotice("[ADMIN_LOOKUPFLW(src)] ([get_disconnected_admin_alert_job_name()]) has been in a deep slumber for [DisplayTimeText(world.time - last_logout_time, 1)]. [fartravel_link]"))
+
+/mob/living/proc/get_disconnected_admin_alert_time()
+	var/role_name = get_disconnected_admin_alert_role_name()
+	if(role_name in GLOB.disconnected_admin_alert_role_times)
+		return GLOB.disconnected_admin_alert_role_times[role_name]
```
Xander3359, line 45: This won't ever return anything because it's just an empty list

Xander3359 (changes requested):

code/modules/mob/living/ssd_indicator.dm
IronDragoon: I put this in so that in future we could set it to only alert for specific roles and with specific times if we wanted, but does nothing currently. I can remove if preferred

code/modules/mob/living/ssd_indicator.dm
IronDragoon: See other comment

PR #2668, opened 2026-09-10
Part One: Timer Subsystem Modernization [TG Port]

hugmeorelse:
## About The Pull Request

A few problems have arisen due to our antiquated timer and sound subsystems. To address them, it felt more than necessary for us to bring in what TG has, so we have a stronger foundation to build off of. 

Originally, I was investigating sounds, as we had reasons to believe, based on reports, that hard deletes from Tracy were contributing to some of the lag spikes we were facing. Overall, it seemed best that we upgrade from the ground up, especially considering TG has made some good changes that should prove beneficial for us in the long run.

For instance, if we want to improve the sound subsystem, we need to actually bring in TG's timer subsystem, as the two are tied together through the `SSsound_loops` timer subsystem. Furthermore, we need to adapt it for our own usage. So, this PR ports that and adapts it for our use case.

For players, timers are the invisible clock managing delayed effects, repeated actions, staged interactions, temporary states, and sounds. They should almost never be noticed, firing only once at the intended time and stopping when the object responsible ceases to exist. They only become noticeable when things break, which isn't good. We've seen the effects of that prior, especially during that big event. 

Part Two is right around the corner with a major Sound Subsystem Overhaul, combining a port of TG’s sound subsystem with our own substantial rework built around it that expands its capabilities for our use case. **It's essentially done and is waiting for this to get verified and tested.** Both parts were developed together. However, splitting them allows us to validate timers before placing the larger sound overhaul on it. After all, that second PR is significantly larger in size, adding over 5,000 lines across 125 files, not including the analysis and testing scripts built to support its optimization. It spares both me and the maintainers a lot of trouble.

Overall, if all goes well, this should give us a good foundation to work off of to improve the codebase further. Plus, I learned a lot, and I still have so much to learn. So, that was a joy. I don't really poke around at code where I have to worry about timers typically, so it's an experience.

**Note:** @Xander3359, tried to follow auto-doc commenting standards as per your recommendation.

## What It Adds

- Ports TG's more modern timer implementation
- Brings TG’s support for multiple timer subsystem instances, allowing specialized workloads to use independent queues and subsystem budgets.
- Adds per-call routing through addtimer(), deltimer(), timeleft(), and updatetimedelay()
- Adds TG’s TIMER_DELETE_ME flag
- Adds TG’s updatetimedelay()
- Adds TG’s dedicated SSsound_loops timer subsystem for the follow-up sound overhaul, isolating looping audio in its own queue and subsystem budget. Its priority is adapted to Ratwood’s subsystem scale.
- Ports TG’s timer diagnostics with Ratwood extensions for client-time timers and SSsound_loops.
- Adds our own timer unit test covering core behavior and Ratwood’s subsystem integration to catch future regressions.

## What It Changes

- Converts compatible spawn() blocks into named callbacks, reducing their use from 153 to 24. These blocks previously acted as anonymous deferred work or "shitty anon async".
- Pending owner-bound callbacks are cancelled when their owner is destroyed. Global callbacks
  remain available where delayed work deliberately needs to outlive its source.
- Preserves delays that were implicit in bodyless `spawn(delay)` statements rather than silently
  making those effects immediate.

These were specifically added from our end. So, these are unproven. Where else, TG's changes are proven on their end. Keep an eye out:

- Fixes `REALTIMEOFDAY` rollover tracking so client-time timers remain valid across midnight.
- Preserves client-time timers through timer-subsystem recovery, verified through a forced Master
  Controller recovery.

- **Fixes four pre-existing bugs that have nothing to do with timers.** The ported `addtimer`
  guard caught one, and the rest were just always there. It was pissing me off:
  - `dust()` calls `death()` first, and some mobs delete themselves in there
  - The Eoran pomegranate aura read `head` on mobs that don't have one.
  - A winch only unlinked one of the gates pointing at it when destroyed.
  - A fox could leak a reference on delete and hard delete, because simple animals only cleared
    themselves from one of the four AI status lists and only one copy of themselves from it.

There's a bunch of other bug fixes, but shrug.

AP Economy Merge Conflict Resolved:
- Merged with upstream after the economy port landed: ATM (meister) and vaultbank's spawn call was replaced.
- There's one slight behavioral change from this: if a machine gets destroyed mid-drill, the next payout no longer fires. Arguably more correct behavior anyways.

## Testing Evidence

- I poked at it with a stick. Specifically:
  - Basically, tested my sound rework throughout it.
  - Did my due diligence in checking a few basic things.
  - Made sure timers worked generally for certain things like making sure the water overlayed over me when I got into water, forging an ingot, and making sure it noticed changes in state while smithing.
  - Tipping cows, putting down torches to extinguish them, checked the smoke state for the arquebus and wolkenmaw, with the intended behavior. You know basic shit.
  - Live timer testing using verbs to trigger queues and such.
- It compiles. Yay.
- Passed the full suite of unit tests 
- Forced Master Controller recovery preserved ordinary and client-time timers, which fired on
  schedule afterward. Verified in private testing, but we have to see how it works in real life. So, mark it as "theoretical".
- Integration runs, 14 of 14 tests with 0 runtimes and `clean_run.lk` success on:
  - roguetest,
  - Dun World
  - Desert Town
- Technically, Rockhill passes but also "fails", but it's a map-loading problem. The stack merges fuck up the rotation gizmos in a stack and has nothing to do with the PR. 

## Why It's Good For The Game

- Gives delayed and repeating gameplay one maintained timer implementation with explicit
  ownership and cancellation.
- Can track time growth easier.
- Provides the isolated scheduling foundation needed for the sound rework without combining both
  low-level changes in one PR.

**Also, it made me fix a couple of bugs that the guards in TG's timer subsystem exposed that was hidden prior.** 

## Changelog

:cl:
code: Ported TG's modern timer subsystem and converted compatible delayed work to timer callbacks.
fix: Client-time timers now remain valid across midnight and timer subsystem recovery.
admin: Added Check Timer Sources for identifying which call sites created live timers.
fix: Fixed several lifecycle errors exposed by the stricter timer and integration tests.
/:cl:


hugmeorelse:
As expected, CI fails to go through. One runtime and two hard deletes are known pre-existing issues, and they trigger on the main branch from what I tested.

The other two runtimes solely exist due to TG timer's capabilities, i.e. they now have a guard to call out whenever it tries to apply timers to already deleted objects. I found this out early while testing out the port, and I found that we had forges that would instantly get deleted whenever it was initialized, because the lava would destroy them, triggering this runtime.

So, for those two runtimes (separate from the lava), it's being counted as a failure for detecting what it used to be silent on prior, even though these are all pre-existing issues that existed before this PR. Rather, it was just never detected until now.

hugmeorelse:
It was annoying me. Fixed. It's not triggering in unit tests, so that's good enough for me.

hugmeorelse:
**AP Economy Merge Conflict Resolved:**
- Merged with upstream after the economy port landed. 
- There's one slight behavioral change from replacing spawn with the callback: if a machine gets destroyed mid-drill, the next payout no longer fires. Arguably more correct behavior anyways.

hugmeorelse:
No screaming about something horribly wrong and on fire so far. So, I'm happy. :)

PR #2672, opened 2026-09-11
Power game reduction for enchantments

RadiantFlash7:
## About The Pull Request
Enchantments now function by providing a cooldown debuff to the wielder/wearer of an enchantment. The length of the cooldown depends based on the enchantment that triggers. Functionally this means only one enchantment will ever trigger even if you have multiple enchantments on armor- only one of them will ever activate before the cooldown comes into effect.

## Testing Evidence
image
image

## Why It's Good For The Game
Butchers a common power gaming method into oblivion.

## Changelog

:cl:
balance: Only a single enchantment can trigger before a cooldown applies to all other active enchantments that use cooldowns.
/:cl:

### DIFF
--- a/code/datums/magic_items/greater_items/greater.dm
+++ b/code/datums/magic_items/greater_items/greater.dm
@@ -1,16 +1,18 @@
 #define PHOENIX_GUARD_COOLDOWN 60 SECONDS
-
+#define FROSTVEIL_COOLDOWN 20 SECONDS
+#define LIGHTNING_COOLDOWN 40 SECONDS
+#define LIFESTEAL_COOLDOWN 10 SECONDS
+#define VOIDTOUCHED_COOLDOWN 10 SECONDS
 ///T3 Enchantmentsdatum
 /datum/magic_item/greater/lifesteal
 	name = "life steal"
 	description = "It seems bloodthirsty."
-	var/last_used
-	var/flat_heal = 10
+	var/flat_heal = 25
 	var/static/list/damage_heal_order = list(BRUTE, BURN, OXY)
 	var/warned = FALSE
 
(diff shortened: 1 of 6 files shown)


coffeecoffeecoffee:
byeah. Killing the method without killing the content is kino.

maker-of-nightmares:
Why does life steal have a cooldown? You can't do 5 of those at once it's only on a weapon and last I knew not a single person is using 3 swords to cooldown max it lmao

Also I don't see anything in the code that says it is discriminatory on different enchantments when one triggers though there is some that state it gives different CDs based on the enchantment(which is fine). Correct me if I am wrong of course, but I sincerely don't believe it should disable all active defense/offense enchantments and just whatever effect triggered for its cooldown.

RadiantFlash7:
> Why does life steal have a cooldown? You can't do 5 of those at once it's only on a weapon and last I knew not a single person is using 3 swords to cooldown max it lmao
> 
> Also I don't see anything in the code that says it is discriminatory on different enchantments when one triggers though there is some that state it gives different CDs based on the enchantment(which is fine). Correct me if I am wrong of course, but I sincerely don't believe it should disable all active defense/offense enchantments and just whatever effect triggered for its cooldown.

Intended function, in truth. If players want to enchant up with an active enchantment on their weapon AND on one their armor, I'm squinting at them heavily for it. In theory, it should reduce the gamer behavior of enchantments.

maker-of-nightmares:
that doesn't  make a lick of consistency nor sense. It makes it unnessassarily confusing and while I understand this is a nightmare on a lux armor asshole sprinting around at ludicrous speed healing and setting you on fire for daring to strike at him, it also makes it confusing to understand _why_ their gear suddenly doesn't work and seems to randomly come on and off cooldown with no reasoning behind the different timers

PR #2687, opened 2026-09-12
Removes ooze noblood

chemistrymain3:
Does title.

## Testing Evidence

One line change.
## Why It's Good For The Game

Oozes are immune to chest crits, bleeding out, and as they can wear armor, easydismember functionally does not apply in 90% of cases. They are a powergamer's favorite race accordingly, because even if critted in the chest, they suffer neither fracture nor will they bleed out and die, so the only way to reliably, consistently kill them is to hit the head, which has been made significantly harder lately. This is further made worse by crit resist.

Don't bother leaving feedback on the PR. I will not read it.

## Changelog

:cl:
del: Removed ooze nobleed
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/roguetown/other/ooze.dm
+++ b/code/modules/mob/living/carbon/human/species_types/roguetown/other/ooze.dm
@@ -9,7 +9,7 @@
 	When an ooze consumes a corpse fresh, mummified, or skeletal, they retain the memories and skills of the fallen adventurers or commoners that litter the ruins. \
 	Taking on new life, they adopt the mannerisms of this formative meal even if they can realize and mentally separate from their template, although many think of themselves as these dead peoples, insisting this is but a second chance at life..<br>\
 	(+1 CON | +1 WIL | -1 INT) <br>\
-	Easy Dismember | Limb Regrowth | No Bones | No Blood</b></span><br><br>"
+	Easy Dismember | Limb Regrowth | No Bones </b></span><br><br>"
 
 
 	default_color = "79F299"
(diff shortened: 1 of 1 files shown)


coffeecoffeecoffee:
can we consider doing something about the simplemob form as well since this seemingly already went through https://github.com/Rotwood-Vale/Ratwood-2.0/pull/2688

an ooze can get up from certain death into a perfectly capable simplemob ooze that can start attacking you and running you out of stamina as soon as they fall in their 'primary' form, and doesn't seem to be a valid target of intents (it certainly can't be Defended against at least). they can dodge in this form, and i don't think it even has stamina, so it can just force you to defend/dodge to run you out of yours or gleefully munch through your armor while you're recovering it... _after_ you should have beaten them. attacking them doesn't even necessarily knock them out of this form - i don't know how much damage it's supposed to take, but it could take a fair little bit in my experience with it.

in my opinion a good solution would be to take away its ability to do things like dodge completely, make it valid for intents like Defend, but give it the ability to move under/through people so it can be the slippery bastard form it was probably intended to be instead of a direbear transform in a smaller sprite

chemistrymain3:
Stop leaving comments. I’m not reading or doing them.

petwir1:
Beyond needed.

rintezukathearmlessartist-beep:
I'd prefer another way to make them killable, honestly, but they're currently hitler.

So remove it and let someone else figure it out later because it's just being abused now.

Femboyms:
I am actually loving to see a single race breaking the de-gut meta, ooze are basically mechanoids but less tank, you can only kill them if you attack the head or life end them. in addition to that- makes no sense to make a slime bleed, but a rebalance of how their hunger works would be great- like eating lux, or making them impossible to be important roles, such is roles of nobility. unless their lore matches something Idk. Ik they have easy dismember but so many counters it with critical resistance which leads a internal conflict between "nerf the min max" and "nerf the ooze" as I said- I honestly I am loving to see a single race change the meta as it was before-

maybe that way the devs fix the skullcrack killing people. since thats the only way to stop mechanoids/ooze and I have seem alot of them recently

chemistrymain3:
irrelevant pr, identical version already merged, need to fix the actual problem w/ their wounds-

PR #2700, opened 2026-09-12
remove stam drain per tick from frostbite

Gug1234:
## About The Pull Request

- Removes stam drain per tick from frostbite status to prevent chain stunning/exhausted pant emote spam from said chain stunning 

## Testing Evidence
3-line removal

## Why It's Good For The Game
Frostbite spell on sojourner is a single-target hitscan slowdown that, on top of stat debuffs and removing a flat 25 stamina on initial cast, drains 5 stamina per tick. What often happens is the stamina drain gets someone to zero and then repeatedly keeps them at 0 for what is effectively a far longer, and frankly janky-looking, stun that spams the exhausted emote/text span alerts for the duration of the effect. 

Frost by itself is good stamina damage, stat drain, and slowdown; the drain per tick, on top of the other effects, makes it oppressive. 

Sojourner's prime targets are mages, who rely on and drain stamina to a greater extent than melee fighters. Gameplay-wise, timing a frostbite spell on an enemy mage right after they cast to try and push them over the edge of stamina exhaustion with burst stam damage is far more interesting than just mindlessly putting your on tick drain on someone to watch them crumple into chainstun exhaustion.

## Changelog

:cl: Yuckuza
balance: Frostbite status no longer drains 5 stamina per tick.
/:cl:

### DIFF
--- a/code/modules/spells/spell_types/wizard/spells_status_effects.dm
+++ b/code/modules/spells/spell_types/wizard/spells_status_effects.dm
@@ -42,10 +42,6 @@
 	target.add_movespeed_modifier(MOVESPEED_ID_ADMIN_VAREDIT, update=TRUE, priority=100, multiplicative_slowdown=4, movetypes=GROUND)
 	target.stamina_add(25)
 
-/datum/status_effect/buff/frostbite/tick()
-	var/mob/living/target = owner
-	target.stamina_add(5)
-
 /datum/status_effect/buff/frostbite/on_remove()
 	var/mob/living/target = owner
 	target.update_vision_cone()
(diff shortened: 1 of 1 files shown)


code/modules/spells/spell_types/wizard/spells_status_effects.dm
```diff
@@ -42,10 +42,6 @@
 	target.add_movespeed_modifier(MOVESPEED_ID_ADMIN_VAREDIT, update=TRUE, priority=100, multiplicative_slowdown=4, movetypes=GROUND)
 	target.stamina_add(25)
 
-/datum/status_effect/buff/frostbite/tick()
-	var/mob/living/target = owner
-	target.stamina_add(5)
-
```
RadiantFlash7, line 48: // When stamcrit, removes it to prevent it from chaining too hard
	if(target.stamina >= target.max_stamina)
		target.remove_status_effect(/datum/status_effect/buff/frostbite)

RadiantFlash7 (changes requested):
I'm going to take a page out of AP on this one. Instead of removing the stamina tick, use the following lines to prevent chain stunning.

code/modules/spells/spell_types/wizard/spells_status_effects.dm
Gug1234: oh nice, yeah better idea for sure. Changed in latest commit.

RadiantFlash7 (approved):

PR #2711, opened 2026-09-13
Forbid Marked by Gnolls flaw on a number of "important" roles

zoktiik:
### Ready for merge, PR has been updated in-place to use Quirks instead.

5 minute PR.

The following 11 roles can no longer take Marked by Gnolls:
- Duke, Hand, Steward, Marshal
- ~~Court Physician, Court Magician~~
- Baron, ~~Retainer,~~ Slaver
- Bishop, Martyr, ~~Keeper,~~ Inquisitor
- Merchant, ~~Guildmaster,~~ Innkeeper, ~~Nightmaster~~

(see [here](https://github.com/Rotwood-Vale/Ratwood-2.0/pull/2711#issuecomment-5655848082) for why some of these roles are crossed-out)

<details>
<summary>[Decent candidates for a Marked ban that haven't made the list]</summary>

- **Prince/Princess** - The "bait heirs" phenomenon. A massive liability for the Keep guards, who have to deal with invading gnolls always trying to break into the Prince's bedroom.
- **Councillor** - These being gone would negatively impact the internal-affairs, the delegation capacity, of the Crown. But from a cursory glance, Marked Councillors do seem to be more rare compared to Marked heirs.
- **Acolyte, Templar** - Acolytes being perma-gone would negatively impact the Church's ability to bring people back. Templars being gone leaves the church without much defense.
</details>

Gnolls have the ability to whisk their targets away to the den with Abduct, and it's practically impossible to get those people back in a timely manner, if at all. Thus, preventing the *obviously* important roles from being Gnoll-marked seems to be an important safeguard for the integrity of the round.

Note that this PR does *not* affect Marked for Death. Any of these roles can allow for Assassins to strike.

## Testing Evidence
Compiled

## Why It's Good For The Game
Impetus: https://github.com/Rotwood-Vale/Ratwood-2.0/pull/2698#issuecomment-5649627235

TL;DR: see About the Pull Request.

Doesn't fix the entire problem, but it should help alleviate the concerns that some of these roles' subordinates might have if these important roles were to disappear entirely.

Also would imply that the Gnolls' priorities aren't quite in-line with those of, say, Matthiosans.

## Changelog

:cl:
del: A number of "important" roles have been forbidden from taking Marked by Gnolls, due to gnolls' high potential for disappearing targets from the round.
/:cl:

PR #2698, opened 2026-09-12
Aggressively un-nerf Gnoll jobslots, rework midround Gnoll injection

zoktiik:
## About The Pull Request
Gnolls got a semi-quiet nerf to their jobslots with the [RoundMods](https://github.com/Rotwood-Vale/Ratwood-2.0/pull/2175) PR.
It removed the old gnoll scaling in favor of RoundMods, but also implemented a nerf to gnoll spawnrates in the same commit. "Too many gnolls" resulted in a hard-nerf, without the community specifically being made aware of it beforehand, and able to discuss it independently of RoundMods in its own PR.

This PR is a reaction in the opposite direction, Now we have [some](https://github.com/Rotwood-Vale/Ratwood-2.0/pull/2548) [refinements](https://github.com/Rotwood-Vale/Ratwood-2.0/pull/2540) to the gnoll gameplay loop, changes which better assist them in hunting down individual players. There is a *positive incentive* for gnolls to hunt instead of frag.

It's possible for us to make a change that's less aggressive, but to what extent needs to be decided based on how things actually play out. The only way to *truly* test how well this buff will work is to see it live.

### The Actual Changes
- **Base Gnoll**
  - Gains 2 base job slots, up from 0. 
    - Note: slots pre-nerf used to be adjusted via [SSgnoll_scaling](https://github.com/Rotwood-Vale/Ratwood-2.0/blob/c483193096af498fd7b92b5fbee1c1ff0423176c/code/controllers/subsystem/rogue/gnolls_subsystem.dm) (removed by RoundMods). It attempted to do similar population scaling to what RoundMods does now.

- **RoundMods**
  - **Low Gnolls:** *REMOVED*
    - Was giving 2 job slots, but RoundMods had a chance to not roll gnolls at all. Removal reduces bloat in the mods pool.
  - **Medium Gnolls:** weight 14, up from the base 10, Adds 2 job slots (4 total), down from 4.
  - **High Gnolls:** weight 8, down from the base 10. Adds 4 job slots (6 total), down from 6. Raises cap on total gnoll slots to 8 (achievable with migration).

- **Gnoll Migration** (midround event)
  - Weight unchanged (it's 5, the base weight is 10).
  - Amount of gnolls added conditionally changed.
    - WAS 2 gnoll slots every time.
    - NOW 2 gnoll slots every time ONLY if Medium Gnolls or High Gnolls do not roll.
    - NOW has a 50/50 chance to add 1/2 gnolls if a Gnolls round mod rolled.
[...]

PR #2714, opened 2026-09-13
Forbid Marked by Gnolls on Prince/Princess and Councillor

zoktiik:
### Ready for merge, PR has been updated in-place to use Quirks instead.

Anti-gnoll hostile architecture. Modular, "optional" companion PR to #2711. 

Guard mains complain all the time about gnolls constantly breaking into the keep, as a round-to-round thing. Observably, it happens. It becomes a huge problem to have to deal with this as a regular thing, when gnolls are designed to be difficult to catch.

Keep roles who rely upon the safety of the Keep walls should not be an active OOC liability to players guarding the Keep.
The cost of letting you bait is simply far too high. It's too overplayed a trope.

Heirs are needed for obvious reasons. They should be *preparing to succeed* the Duke or Duchess, they should be roleplaying as actual heirs to the throne. They should not, in one week, decide that they want to be whisked away forever after a lyfetime of preparation. Every time the guards get called to "Heir's bedroom, 2nd floor", a rose wilts.

Councillors are also a logical ban choice, as they similarly enjoy the guards' protection. They are perhaps an even more important ban for Marked, because of how they maintain the Keep's daily operations. They are *supposed* to be there for when the Crown needs to delegate something. Keep roleplay needs them.

The other PR was made to address the *direct round impact* of Gnolls successfully abducting targets that are at least semi-important to the round. This PR additionally addresses the *social impact*.

## Testing Evidence
Compiled

## Why It's Good For The Game
See above.

## Changelog

:cl:
del: Prince/Princess and Councillor are additionally forbidden from taking the Marked by Gnolls flaw.
/:cl:


coffeecoffeecoffee:
_as_ a major guard player what bothers me personally is much moreso when an extremely central town support role like merchant or guildmaster turns out to be abductbait rather than the abductbait roles with nothing better to do being abductbait, myself

zoktiik:
Yeah but you complained about gnoll players constantly invading the Keep, which is obviously caused by stuff like this.

I really don't care what bothers you personally. I care about addressing root causes.

Solve player issues in the code.

P-roxii:
Yeah so I don't really see why these non-important roles should be barred from such a vice. So what if they break into the keep? Just have someone make a few buckets and repair the doors, it isn't that hard to deal with. What other threats are there in the keep that occupy the garrison's time? Thieves? The keep is probably the easiest place to catch a gnoll other than the shops around town, it confined area where there are only so many places the gnoll can go once seen, even if they go invisible.

As both a garrison player, and on occasion keep councilor/heir player, I do not like this idea.

zoktiik:
I already wrote my reply

See also https://github.com/Rotwood-Vale/Ratwood-2.0/pull/2698#issuecomment-5649627235 for part of impetus

Trilaterall:
>I already wrote my reply
> 
> See also [#2698 (comment)](https://github.com/Rotwood-Vale/Ratwood-2.0/pull/2698#issuecomment-5649627235) for part of impetus

That just makes it seem like this is a spite pr, and that you believe that you know best and everyone else’s opinion is irrelevant. There is no “player issue” here. Councillors and Heirs are unimportant to the round. There really isn’t a reason why they should be barred from the vice.

Heirs who want to get caught by Gnolls don’t stay in the keep, they are going to go somewhere not surrounded by guards 24/7. Those in the keep probably are not baiting a Gnoll to come and get them.

zoktiik:
It's not rly a spite PR when the goal is pragmatically responding to the issue of bait heirs/courtiers fucking things up for everyone else

lol

>Heirs who want to get caught by Gnolls don’t stay in the keep

r u sure about that

Mogelix:
> Heirs are unimportant to the round

Do RW-ers really?

RadiantFlash7:
I don't really think we need to be concerned about heirs or councillors getting gnolled honestly. it's a role with multiple slots, and their responsibilities are minimal, beyond when the duke/duchess dies.

PR #2721, opened 2026-09-13
Kills Sojourner and replaces it with PONTIFEX

AlucardHellsingChapel:
## About The Pull Request

The PR finally removes the Sojourner, one of the worst classes to fight against. It is replaced by the Pontifex, which has access to four paths and its own skill set. The Mercenary Pontifex was also removed from the mercenary list.

Arcyne Momentum: Melee strikes grant 1 momentum every 2 seconds to fuel arcyne power. At 7 or more, it overcharges and must be spent. If it decays unspent, all momentum discharges at once, leaving you strained, slowed, and vulnerable.

Fist of Psydon: A 3x3 area attack dealing double damage at 3 or more momentum, consuming the momentum.

Grasp of Psydon: An ensnare and fetch effect that also deals damage at 3 or more momentum.

Storm of Psydon: A minimum of 7 momentum is required for the weaker version. You leap to the target, delivering a series of punches and a final kick to knock them back. At 10 momentum, you execute 10 strikes, being 9 punches and a kick, with the same knockback effects.

## Testing Evidence

https://github.com/user-attachments/assets/c78b33fe-35fe-440b-8d3e-fa6ca8ed2307

image
image
image

## Why It's Good For The Game

It finally kills the most annoying class to fight against and gives it a proper replacement.
image
PSYDON'S WEAKEST SOLDIER

## Changelog

:cl:
add: Added three unique skills
del: Removed old Sojourner
del: Removed Mercenary Pontifex
balance: added some extra durability to pontifex hood(helmet)
/:cl:


AlucardHellsingChapel:
i maay have forgotten to change where it spawns oops?

AllCorruptingWyrm:
THANK FUCK

AlucardHellsingChapel:
will require a bit more work

wikimody:
I dislike removing the mercenary pointifex

AlucardHellsingChapel:
@wikimody i may add a non miraclist version later on but currently my priority was removing the Sojourner

wikimody:
> @wikimody i may add a non miraclist version later on but currently my priority was removing the Sojourner

I'm not against removing soldurner though not because of his stamina drain abilities, but the absolute 0 build potential. He can only ever be a basic ice mage with anti magic shield. He even has T1 arcane to make it absolutely impossible for him to go  outside the nish given to him

AlucardHellsingChapel:
@wikimody don't worry there will be merc pontifex just not right now

AlucardHellsingChapel:
it's something, I have planned it will be a non miraclist version and perhaps with one other path to replace the counterspell (control) path

wikimody:
Sidenote, why remove pointifex at all from mercenaries. You could just give a bit different name to the inqusition one, and it would also let you get by the lore implications this change has

AlucardHellsingChapel:
@wikimody it's quite simple sleep depravation makes me do the worst choices of my entire LYFE i could mostlikely readd it quite easily

ophaq:
Please don't remove mercenary pontifex. It's a fun class and being forced to play in inquisition for it kinda lame. Also, you say it is one of the worst classes to fight against but that is literally any mage that takes a similar spell set up. Are you a skeleton main? They are supposed to counter undead easily. Maybe change up their spell kit instead of just removing them? 

Speaking of which, the spells you added sound pretty cool. You could use some of those and replace the more annoying spells instead. But a lot of these 'spells' are melee focused which sojourner isn't supposed to be thematically. They're a spellcaster not Kenshiro from Hokuto no Ken. Same with the hierophant. They're pretty much Naledi's epitome of arcyne mastery.

5467543:
sojourner is not 'just a mage with ice spells' they have a class specific frost bolt that ONLY sojourner gets and it will exhaust you in 2 hits and instantly win the fight, it's unbelievably toxic

son goku is better

RadiantFlash7:
Sojourner was very particularly designed to be the bane of mages. I don't personally think the inquisition needs another melee class when we have disciple. I'll leave this up for a discussion- but at the moment I'm leaning towards a rejection for this PR.

ceeetchell:
i'd rather if you just reworked sojourner so the inq can have some sort of mage/alchemist option, and then just add this cool thing as it's own ortho subclass

AlucardHellsingChapel:
> i'd rather if you just reworked sojourner so the inq can have some sort of mage/alchemist option, and then just add this cool thing as it's own ortho subclass

Well.... the path of survival more or less gives them able to make potions they may not be the best but they still can...

RadiantFlash7:
Going to call it on this PR for now. If you want to update the mercenary pontifex role with a similar setup, I can review that there- but as it is... We're keeping sojournor for the time being.

PR #2724, opened 2026-09-13
Cure rot patrial rework

wikimody:
## About The Pull Request
Changed the cost of cure rot from 200 devotion to 100 devotion
Added new mechanics for it: It fully purges all toxin damage and removes ALL chems from person affected (the prelate version lacks this)
The mass flashbang only happens when the prelate version of the spell is used
It now deals 75 burn if the prelate version is used

## Testing Evidence
Zrzut ekranu 2026-09-13 215357

## Why It's Good For The Game
Curently cure rot is a spell with exists exclusivly to deal with deadlites. It doesnt even do that particularly well because it takes: a cross nearby, long charge up time, 1/5 th of devotion for people with full 1000 max devotion. Its only advantage over an extremly simple surgery (with anyone who uses the miracle can perform because pestrian) is that if the deadlite wasnt somhow killed before it got to you, it isnt killed by miracle when undedliting, meaning in very specyfic circumpstancess this can serve as a revive. Its quite rare though as people tend to instantly kill any deadlites they see
This first makes it not extremly expensive for doing a job of 4 second surgery, and gives it another use case outside of deadlite curing, that is dealing with toxins and other chem issues. It is fiting for pestra the goddes of medicine and poisons to have a way to deal with both of those things, and it makes the miracle usfull outside of its deadlite bashing. It still has high coldown, high cost and needs a standing cross, so it shoudnt be opresive on any poisoners. (it also lets you cure deadly poisoned bloodless psydonites with is as hard to heal as it sounds)
Also some minor changes to the prelate version of the spell (they get one innatly, its astarta themed instead of pestrian and costs 30 devotion with 2 minute coldown instead). It doesnt get the tox healing as thats not what astarta is about, only it causes the mass flashbang because pestra has no relation to light, and it deals 75 burn damage when succesfuly removing rot as again astarta is not particularly gentle either. Mostly flavour changes

## Changelog

:cl:
add: Pestrian cure rot miracle now also heal tox damage and purgess all chems from the target (its aslo bit cheaper to cast)
balance: Bishop version doesnt do that and now deal minor burn damage on succesfull removal (Astarta is not a plesant goddes)
qol: Pestra denies all conotetions to light and leaves mass flashbanging people to Astarta 
/:cl:


wikimody:
I found some weird bug with inheretence of apply damage on undead animals, so fixed that

PR #2730, opened 2026-09-14
Ranged Black Oak Pariah

ceeetchell:
## About The Pull Request

this pr does two things: introduces a ranged option to the black oak pariah wretch, and changes the max slots from 1 to 2.

### ranger

- gives you the elven recurve bow from the mercenary class and quiver of arrows
- restricts your sidearm selection to only saber and dagger
- lowers sword skill to journeyman
- gives ~~expert~~ master bow skill
- replaces summon and enchant weapon with mending and longstrider ( summon and enchant are 2 spell points, i figured those 2 spell point options would be more useful to a ranged kit. )
- gives +1 per

i also changed up the tutorial text a bit to be more appropriate for a role that would have (possibly) two people instead of one

## Testing Evidence
pretty simple inventory change

oaWCVEsI5y

## Why It's Good For The Game

why not?

the old tutorial text mentions bows, yet there isn't a bow option present. the mercenary has a bow option, so it's a bit weird for the wretch verison to not have one as well.

as for adding an extra slot, playing games is more fun with friends, and the addition of a ranged option means that both players can (if they want to) have a unique kit. it also adds more flavor to the vale on an antagonistic force you may see that isn't just zizo undeath murderball.

## Changelog

:cl: mushy
add: Black Oak Pariah now has a ranged variant.
balance: Black Oak Pariah now has two slots.
/:cl:


DogExtrordinaire:
the trees are speaking elvish

UntoldTactics:
You'll probably want the ranged option to trade some stats for additional per, since they really don't get much for a ranged class. The poacher would be better than them in near every way, even accounting for magic.

Honestly elves especially r kinda screwed since they only get +1 speed (whereas dark elves get +1 int and +1 per or whatever), etcera.

zBxnny:
Should change the stats if you're giving them the ranged option. They don't get considering per to trade their melee capabilities for ranged, basically meaning they have no damage. And also expert bows is kinda dog for a specific ranged class, consider giving them master.

ceeetchell:
i wanted to avoid making them just a better poacher or really broken, since spells + woodwalker + bow seems already really powerful, but sure i'll give them a bit more

RadiantFlash7 (changes requested):
Turning low memory mode on while this is testmerged is not good. I'll need you to revert that commit, or I'll probably have to untestmerge the next time we update TGS.

ceeetchell:
> Turning low memory mode on while this is testmerged is not good. I'll need you to revert that commit, or I'll probably have to untestmerge the next time we update TGS.

omfg i'm an idiot sorry

RadiantFlash7 (approved):
Better, Thank you.

PR #2733, opened 2026-09-14
Remove stamina addition in frostbite effect

chemistrymain3:
Removed stamina addition from frostbite status effect.

## About The Pull Request

Does what the title says

## Testing Evidence

Works
## Why It's Good For The Game

Frost spells already have several functionally undodgeable means of application and also frostbite slows you down by a factor of 4 (twice that of Lightning Bolt) for 20 seconds. It did not need to utterly nuke your stamina by 125 over 20 seconds as well, which is part of what made frost so oppressive. 

There was a fix to stop it chainstunning you, but frankly it did not function. You could still get chain-exhausted, just not while your exhaust was already at 0. 

They are still excellent at their intended role- slowing someone down to the point they can't get away from you. If they're 'too weak' (they will not be) I can give them something else in exchange. At the moment, they're widely considered Fucking Busted on frost mages, frost enchantments, dreamwalker weapons, etc.
## Changelog

:cl:
balance: Frost spells no longer drain your stamina directly.
/:cl:


wikimody:
I belive it makes them quite pointless, but the stamina hell wasnt good either.

HalfAwake691:
Could just make the debuff last longer to compensate but yes stamlock hell is really ech.

RECALLPROLOGUE:
the op died to a mage

PR #2752, opened 2026-09-15
Fixes a minor bug with stamina / fixes a minor description error

chemistrymain3:
## About The Pull Request

You now correctly spawn with full stamina on spawn.

Athletics skill now correctly tells you what it does, instead of citing mechanics that no longer exist.

Rather than setting your energy cost to 0 as a skeleton through energy = max_energy, skeletons now just get the infinite energy trait, which does the same thing but doesn't make you scratch your head going 'wait where's the trait.'

## Testing Evidence

I did it on my phone in a webedit individually one by one. There's probably an indentation error. I'll go actually look at it when i'm on my computer. Ignore the commit spam. It's just ugly. Don't code on your phone.

## Why It's Good For The Game

It's just a bugfix.
## Changelog

:cl:
fix: You now spawn with max energy on spawn as intended.
fix: Athletics skill now correctly tells you what the Athletics skill does.
/:cl:


Xander3359:
Pretty sure there's a good reason why we removed infinite energy from skeletons

chemistrymain3:
> Pretty sure there's a good reason why we removed infinite energy from skeletons

You didn't. They HAVE infinite energy.

Their energy is always set to max.

PR #2753, opened 2026-09-15
swap thespian crit resistance for bleed resistance, buff bronze heavy plate integ, fixed azure place names

Gug1234:
## About The Pull Request

- Thespian critical resistance swapped for bleed resistance. Now only the gladiator loadout gets crit resistance. 
- Bleed proc trait logic tweaked; bleed resistance is checked before crit resistance, meaning the 0.5x bleed rate modifier comes before crit resistance's 0.8x (this effectively makes prenerf crit resistance when both traits are stacked).
- Bronze panoply adjusted to 500 integ and 450, on par with steel. Better fits the format established by other bronze, like the helmets, where, in exchange for worse protection, they have greater integ. 

## Testing Evidence
test evidence3

## Why It's Good For The Game
Closer to Azure implementation. Plate training + crit resistance is probably a bad idea : )

## Changelog

:cl: Yuckuza
balance: Nerfed thespian crit resistance, replacing it with bleed resistance. Gladitor loadout for thespian is now the only loadout that gets crit resistance.
balance: Bronze heavy armor integ buffed to be on par with steel. 
code: bleed rate math adjusted so that bleed resistance 0.5x reduction takes priority over crit resistance 0.8x.
/:cl:


Gug1234:
> One thing that fucks with me about thespian, is that their "natural armor" (not pitfighter chest) doesnt get head coverage. Why wouldn't you just roll NA and get the whole thing?

The natural armor they get has better protection and Regen than virtue nat armor.

PR #2755, opened 2026-09-16
Malodorous Rework (and Plap hearts)

IronDragoon:
## About The Pull Request

With Malodorous now a quirk, I have made it more indepth.  Also I have RENAMED it to Redolent (which just means strong scent). You can choose between positive/neutral/gross effects and use either a default or custom described scent. Neutral has the standard 'stink lines' effect, gross has this AND occasional flies buzzing around, positive has an occasional pink sparkle. These will be fully transfered to a receiver during ERP, making them inherit the stink for 15 mins or until cleaned off.

Also adds the ERP hearts/plap effect which is gated to a pref. It only appears to participants to minimise performance issues.

I renamed this later on, wherever the text or images say malodorous, just imagine it says redolent, trust. 

## Testing Evidence

image
image
image
image

Neutral Scent
image
image

Gross Scent
image
image

Positive Scent
image
image

Scent 'sharing'
image

ERP effects
image

## Why It's Good For The Game

Lots of people have asked me for this. I think lots of people want more control over this and how it is perceived and I think this adds a lot of options

## Changelog

:cl:
add: Malorodous is now much more configurable, can be gross/neutral or positive and is renamed to Redolent
/:cl:


fisherfridrich:
I do suggest trying to add a toggle, as some ppl are really put off by musk/scent play being out in the open for everyone. So adding a toggle, just like what was planned with chastity content, would be nice so there is no unnecessary shit slinging against this PR.

IronDragoon:
> I do suggest trying to add a toggle, as some ppl are really put off by musk/scent play being out in the open for everyone. So adding a toggle, just like what was planned with chastity content, would be nice so there is no unnecessary shit slinging against this PR.

I mean I suppose so, I only didn't given it was already in the game before this point but I will give it a look

Llizny:
Can you remove the 1 point cost? It had zero before this which made it nice to fit on without triumph points.

EDIT: Adding/or separating the cost per smell seems like a much better option. Positive smell gives a small stress down while neutral and negative will inflict stress to anyone that smells it. It feels counter-intuitive to make this negative part cost a point.

sippykot:
the particles are triggering too often, the pink sparkle spam is pretty obnoxious

coffeecoffeecoffee:
treating the pink sparkle stench as the most disgusting odor ever conceived wafting of all the princesses in pink clouds and using the vomit emote when the sparkles come off them the first time they enter a room is pretty funny, but i don't think that's the intended effect and it's very much not very good for any other kind of roleplay that I've seen thus far

"good" smells blasting off you in clouds of particulate matter is amusing in a twilight vampire gag sort of way, but the default for having a "nice" smell should probably be to not play an animation at all and have it as an option, because a good smell is usually one that's not blasting you in the face, or indeed, even immediately noticeable i'd imagine

EmilyFontaine:
Could someone weave the code for malodorous custom descriptions into natural armor, I'm so tired of writing it manually every time...

sippykot:
> treating the pink sparkle stench as the most disgusting odor ever conceived wafting of all the princesses in pink clouds and using the vomit emote when the sparkles come off them the first time they enter a room is pretty funny, but i don't think that's the intended effect and it's very much not very good for any other kind of roleplay that I've seen thus far
> 
> "good" smells blasting off you in clouds of particulate matter is amusing in a twilight vampire gag sort of way, but the default for having a "nice" smell should probably be to not play an animation at all and have it as an option, because a good smell is usually one that's not blasting you in the face, or indeed, even immediately noticeable i'd imagine

i mean, the whole quirk is about a very strong smell so it makes sense?
i just think the particles show up too quickly, should be like 2-3 times as long between particles imo

RadiantFlash7 (changes requested):
After some feedback from players experiencing it in round, and discussion amongst maintainers, The pink particles are a bit much at the moment. I'd reccomend either changing the effect to something less jarring, or cutting an effect entirely for positive smells.

zeroisthebiggay:
It's been said to death but I'm going to throw my hat in the ring and say, yes, the sparkles are too much and I think they're more disruptive than someone having green stink lines.

Kroto-Ker:
Un-testmerged until the obnoxious pink sparkles are removed or made much less intrusive.

IronDragoon:
> Un-testmerged until the obnoxious pink sparkles are removed or made much less intrusive.

this has been made far less intrusive. Visuals toned down and occur only once every 30 seconds, text messages every 60

PR #2775, opened 2026-09-17
Clingy rework

Xander3359:
## About The Pull Request
Clingy now lets you mark a target.
Time spent away from that target makes you gain an increasingly strong mood debuff.
If the target ever despawns (like far travel or deletion) you will become able to select a new target.

The person being clung to now has a way to deny someone clinging to them.
This is effectively opt-out on an individual basis, because clingy people still should be able to give it a try (Though the ideal is that they RP some buildup before just ripping it)

They have:
1) A chat message they can click on
2) A verb in the IC tab that lets them drop anyone clinging to them

## Testing Evidence
image

image
image
image

## Why It's Good For The Game
Vices should not be free. They should be impactful and affect your gameplay.
There are other "free" vices as well, those will be addressed as my motivation to work on them arises.

## Changelog

:cl:
balance: Changed how clingy works, it now requires you to actually act clingy
/:cl:


Eve393:
i’m all for making vices more interesting than ‘stand adjacent to anyone for a bit’

but i have pretty big ooc concerns with how this is implemented

stalking/clingy behavior is a bannable offense within the boundaries of making a player uncomfortable, and personally having my character obsessed over without a way to opt-out is genuinely uncomfortable to me

i have had to ticket similar behavior multiple times and can tell you that the sheer concept of trying to be with someone’s character without giving them privacy will probably end up with a lot of people upset

i know that we spoke on discord, you have said “then either tell them to go away or kill them” but i don’t think these should be the solution of an interaction forced by mechanics

i know you have it in your interest to have a semblance of IC creepiness, but IC creepiness needs a lot of finnesse to pull off without weirding someone out if they haven’t signed up for it

any other interaction where you are the ‘target’ requires you to opt in, and i know you said that this too should have a similar mechanic where you can at least avoid it

But in its current state i don’t think potentially making someone uncomfortable behind the screen via stalking should be a game mechanic

5467543:
This vice punishes the person being obsessed over more than the person obsessing, and currently with no way to opt-out of it

zeroisthebiggay:
I like the concept of making clingy an actual vice, but encouraging obsessive behavior isn't it. I'd rather it be reworked so it's kind of how it is now, but more harsh. Needing you to be adjacent to other players for an extended period, maybe even pixelshifted so you are really close. Not working in combat mode, with harsher debuffs since it's a relatively simple vice to fulfill.

Xander3359:
> This vice punishes the person being obsessed over more than the person obsessing, and currently with no way to opt-out of it

It's a valid concern. 

I'll add a way for the target to have a way to opt out, both initially and whenever they so choose. 

And if they opt out it will prevent the person from ever selecting them in the round. 

As well I'll likely add a mood buff to the person being followed, something like "%CLINGY% is always following me like a dog, how cute!" So that there's a (minor) benefit to allowing someone to cling to you.

Xander3359:
New update: The person being clung to can now reject someone clinging to them

NotGoldenglow:
I'd give it a look-over, sometimes the debuff doesn't clear or even recognizes that you are indeed clinging to someone.

Xander3359:
> I'd give it a look-over, sometimes the debuff doesn't clear or even recognizes that you are indeed clinging to someone.

fixing in next round (Just updated the PR)

NotGoldenglow:
... maaaay also want to implement some kind of safeguard in case the Clinged goes SSD. There's been plenty of tickets because 'I can't find my target!' and they've gone and SSD in a corner without far travelling.

PR #2778, opened 2026-09-17
QOL: Automatic Parry/Dodge selection

JustMeTheIInd:
## About The Pull Request

This is an extremely simple (slop, sorry!) PR. Whenever you spawn, it looks at your traits, and if it sees you as having either TRAIT_MEDIUMARMOR or TRAIT_HEAVYARMOR, it auto-sets your stance to 'Parry'. Otherwise it remains in 'Dodge' 

## Testing Evidence

Went on MAA: Parry
Went on Forester Warden: Parry
Went on Ranger Warden: Dodge
Went on Soilson: Dodge
Went on Adjudicator: Parry
Went on Adventurer ->Battlemaster: Parry

## Why It's Good For The Game

Less pointless clicking... good?

## Changelog

:cl:
qol: Parry/Dodge now sets automatically, based on whether or not you have armor traits.


zeroisthebiggay:
This shouldn't affect NPCs, right? It if it does there'd be problems since they're manually set.

JustMeTheIInd:
> This shouldn't affect NPCs, right? It if it does there'd be problems since they're manually set.

I do not think so, it uses job.dm and subclass.dm, and I believe NPCs don't use 'jobs' as much as they use their own special code.

JustMeTheIInd:
I checked and apparently it did affect NPCs, made it a client check now so it should hopefully only look on players

PR #2779, opened 2026-09-17
automatic free use toggle

zeroisthebiggay:
## About The Pull Request

Adds a toggle in OOC preferences that automatically turns on free use.

## Testing Evidence

image

## Why It's Good For The Game

Saves clicks.

## Changelog

:cl:
add: automatic free use toggle
/:cl:


Xander3359:
Just make it a pref then???????????????????????

zeroisthebiggay:
> Just make it a pref then???????????????????????

You might be onto something

zoktiik:
Having both the pref toggle and the temporary setting is good here. You can situationally toggle it on or off based on what's happening in the round, and then the "permanent" pref setting is just the default.

A lot of people would like to have the default be one or the other. That's what the pref here is for.

Xander3359:
> Having both the pref toggle and the temporary setting is good here. You can situationally toggle it on or off based on what's happening in the round, and then the "permanent" pref setting is just the default.
> 
> A lot of people would like to have the default be one or the other. That's what the pref here is for.

You can also situationally toggle the pref. You just type "Toggles" in the command bar and scroll to the ERP options then toggle it. It's the exact same as when you turn off ERP panel so nobody can touch you.

zoktiik:
Kinda, but I already have the button right there in the relevant menu. Maybe that could trigger a prefs save, but I think it'd be a little janky under-the-hood with the two structures having to communicate.

The-Sun-In-Splendour:
this is actually not testmerged depite it being tagged that it is

PR #2784, opened 2026-09-18
Belt whip

Randumb-Phantomb:
## About The Pull Request

You can now whip people with your belt, it does like ten damage.
Also chains too.

## Testing Evidence

image

I tested it, it works.

Whatever errors aren't my fault

## Why It's Good For The Game

Sometimes it's a hassle to go get a whip to have fun times, so now you can use your belt. Also you can use chains too. They function as like, extremely bad backup weapons, like, terrible. They do 10 damage, have no bleed, and don't pierce any armor. You can also punish your children for being fuckups as duke, without like, murdering them.

## Changelog

:cl:

add: You can now whip people with your belt, it does like ten damage.
Also chains too.

/:cl:


MrShibaa:
As a South African who has experienced multiple ass whoppings via the belt from my parents, I do not want this.. The mental pain and ptsd it causes is way too much..

zoktiik:
Does remind me that "blunt" whipping options (that cause no bleeding on WEAK) are very scarce

PR #2786, opened 2026-09-18
Being buckled to structures does not block spellcasting, with exceptions

zoktiik:
## About The Pull Request
I find it very annoying that when I buckle myself to a bed, I cannot cast spells. Your hands being tied already blocks most spellcasting, and that would include being buckled to a bed, to hanging chains, etc.

If there's something I'm missing that *specifically* should be blocking spellcasting, please comment.

Specific exceptions: pillory, noose, gallows, crosses, wooden horse, meathook, x-pillory, hanging chains (both types), torture table (the rack), guillotine, sacrificial altar, pyre.
Basically, all "bondage equipment."

**post-analysis edit.**
Technically speaking, this PR also means that:
- Being "buckled" to a trap like maneaters, quicksand, or spike pits **do not** block spellcasting anymore.
  - (They used to block because they're not simple mobs.)
- Piggyback riding would now allow spellcasting - and actually doesn't have the minor constraints that mounted spellcasting has.
- You can cast spells while "buckled" to a gloryhole object.
- You can cast spells from a dinghy.
- Being buckled to a cross or a pyre redundantly blocks magic casting, since you need to be "handcuffed" (restrained) in order to remain buckled.

Completely unrelated to this PR (atm):
- You can cast spells, whether on-self or "invoked", from inside a handcart, minecart, closet, or chest.
  - (I'm pretty sure, at least.)

Currently I'm asking maintainers for guidance on how to proceed.

## Testing Evidence
Compiled, idk if I will test this. Just throwing it out there.

## Why It's Good For The Game
QoL mainly

## Changelog

:cl:
qol: Being buckled to structures does not block spellcasting unless it's specifically some kind of bondage equipment. Having your hands restrained still counts, so bed-buckling still works. 
/:cl:


maker-of-nightmares:
Did you test to ensure this doesn't stop handless casts from activating properly like Message on the restricted structures?

zoktiik:
I just double-checked and Message is blocked on live (upstream Ratwood-2.0), you cannot cast even Message when you are buckled. The current logic bypasses all of that stuff.

I could make handless spells castable while buckled, but that would technically be yet another gameplay change. Not opposed to it, it makes sense, but it's a little extra hassle and its own little can of worms.

maker-of-nightmares:
You used to be able to cast message while bound it was intended as a means of calling for help and is a rather potent RP tool(I've used it to make like minor illusions for demonstration before among other things). You can also cast the 2 conjured armors when just chained up so it would make sense to ensure those are still castable on pillories and such

zoktiik:
Sure, but that's not this PR's fault. I'm going to fix it anyways. New PR coming very soon, since the scope and mission have changed a fair bit.

zoktiik:
#2795
@maker-of-nightmares

PR #2794, opened 2026-09-19
Obliterates subtle scomring and crownstone use (because I died to it once)

AllCorruptingWyrm:
## About The Pull Request

I didn't actually die to it, I just wanted a clickbait title.

## Testing Evidence

when I get to it

## Why It's Good For The Game

Because going "Sec to surgery," is bad. Radios are bad.

## Changelog

:cl:
tweak: scom rings have a delay
tweak: scom rings say aloud, not whisper
Tweak; Crownstone rings have a delay
tweak: Crownstone rings say aloud, not whisper.
/:cl:


zeroisthebiggay:
image

Be deadass a second.

I like the speaking aloud part. The delay is unintuitive, superfluous, and makes the whole spoken aloud part pointless if you're assigning it a kneejerk value of fifteen seconds. At least write these values in the PR body.

GrayRachnid:
I'm okay with this change only because then I won't feel bad about calling for help and the other side not noticing.

Add a delay to listeners too.

AllCorruptingWyrm:
> I'm okay with this change only because then I won't feel bad about calling for help and the other side not noticing.
> 
> Add a delay to listeners too.

Oh my god.

AllCorruptingWyrm:
> image
> Be deadass a second.
> 
> I like the speaking aloud part. The delay is unintuitive, superfluous, and makes the whole spoken aloud part pointless if you're assigning it a kneejerk value of fifteen seconds. At least write these values in the PR body.

How about five seconds? and sure I can do that.

zeroisthebiggay:
> How about five seconds?

I don't think it needs a delay. Being forced to say it out loud (and ideally it having a loud sound cue so you know that is what they're doing) pretty much has the same effect on escalation.
If you're dying on the delay hill, then a two or three second one with a tochat message saying `John Sergeant raises his crownstone up to his face...` gets the point across perfectly fine.

deadmomo69:
Just have it put a message they put the scomstone up to their lips when they use it. It was like that before, and it was considered "escalation", I dunno why that changed.

RadiantFlash7:
My instinctive thought is to keep the whisper function of speaking into the scomstone, and instead, make an overt message when someone uses it.  So you can firmly tell 'yes, this person just whispered into their stone'.
That prevents needless chatter filling the chatbox being added in areas when someones barking orders into an scom stone, while allowing people who see this happen to be well aware of it happening.

Randumb-Phantomb:
Make it so a physical rat comes out of the ring and runs to the nearest scom, and then the message is sent.

loleo14:
You know there use to be a little thing on SR whenever someone used their rings it said that they pressed it against their mouths or something, maybe we should use that-- wait it was on RW too for awhile but then it was removed?

sippykot:
🧂

5467543:
Currently when trying to antagonize anyone with a scomstone, you just have to assume backup is running to your position ASAP because you have no real way of telling whether they've spoken into the ring or not - paradoxically making escalation forced and rushed for both parties

good change

the delay is good too but people aren't ready for that truthnuke yet, and the 'speaks into the ring' notice when someone begins to start using it is an even better change because it allows agency on the part of the escalator to respond BEFORE 50 knights have been dispatched to their location and means you don't need to add an annoying delay for the user

currently mage message has this exact mechanic so there's even precedent for it

AlucardHellsingChapel:
Tbh very good pr that should be merged as fast as possible

Ducky-Momo:
seeing antags doing a bit or otherwise RPing just for some fuck with a scomstone to go "HELP LOWTOWN", knowing a garrison deathball is on the way is the epitome of :/

very much in support of this

PR #2795, opened 2026-09-19
Spells: Fix riding/buckled casting logic, + MORE

zoktiik:
## About The Pull Request
This PR comes with three main changes. **None of this is set in stone.**
But in its current state, we're fixing shitty code and allowing for there to be a sane initial state to rebalance off of.
- You can now cast while buckled to "structures"  that aren't restraints or traps. Bondage equipment, maneaters, quicksand, spike pits, etc. prevent casting spells that require you to be unrestrained.
  - The behavior used to be broken. Sitting down in a chair or lying down in a bed while unbound would still block your casting. This is because being buckled to *anything* would block spellcasting if it wasn't a steed.
  - New variable buckle_blocks_spells controls what structures explicitly block the somatic (gestural) component of spells.
- The following "structures" now explicitly block gestural components for spellcasting when you're buckled to them. They do not block spells with `gesture_required = FALSE`.
  - "Bondage equipment": X-pillories, Torture tables, Pillories, Wooden horses, Hanging chains, Meathooks, Nooses, Gallows, Guillotines, Crosses*, Pyres, Operating Tables*, Cart Rail Breaks*. (* = redundant w/ `buckle_requires_restraints`)
  - Traps: Spike pits, Maneaters, Tanglers, Wooden spike barricades, Quicksand (in Desert Town)
- Mounted casting is now standardized. You can cast while "riding" anything, including steeds, human mobs (carrying & piggyback), and dinghys, but the thing you're "riding" still needs to stop for 2 seconds after any movement.
  - There is a new rejection message to indicate as much.
  - Under-the-hood, we check for you being buckled to anything that uses /datum/component/riding.

### This PR does not
- Affect mounted casting from a steed in any way. That already has its 2 second delay post-move. In all aspects but code, I have *not* touched that behavior.

## Testing Evidence
I'll be TMing this on my private test server to make sure the basic stuff all works fine. Slopbot review couldn't find nuttin', and after reading over the code, I think the logic's pretty rock-solid. *I will still check to make sure things work.*

**Update 1:** Somatic spellcasting restrictions by structures appear to be working quite well.  Still need to test riding, but so far, so good.

https://github.com/user-attachments/assets/7e6bae24-16ee-4d45-8444-21d7e33612f9

https://github.com/user-attachments/assets/7e745606-9025-4133-b08c-c9ef72fd8454

**Equestrian demo, showing that EXISTING MOUNTED BEHAVIOR REMAINS UNCHANGED.**

https://github.com/user-attachments/assets/5df23858-bfd1-4c7e-a0fc-38cff7794b4d

## Why It's Good For The Game
To put it simply, the logic for spellcasting got fucked when they added the check for being buckled to steeds. It completely ignores the verbal/somatic logic that spells normally have to follow, affecting ALL FORMS OF BUCKLING to boil the issue down to whether or not you're specifically buckled *to a steed*. If not, everything else gets ignored. **This is bad.**

So bad, in fact, that being buckled to anything will block even the Message spell, which is intended to be more-or-less unblockable.

This implementation of the mounted casting change is purely for the sake of consistency. I don't mind making changes.

## Changelog

:cl:
fix: Being "buckled" to structures (sitting down in a chair, lying on a bed) no longer blocks spellcasting on its own. It also now respects spells that don't require a gesture component to cast. Now, you have to specifically be restrained, or buckled to a structure that restrains you (i.e. the rack, pillory, etc), or caught in a trap that you need to resist out of, for any of that to block gestural casting.
add: Mounted casting has been opened up to anything you can "ride" (carried/piggyback on a human mob, riding a dinghy). However, these also require that you (and your "mount") stay still for 2 seconds before you can cast.
/:cl:


DorfGuy1:
can you also take a look at the mounted lance charge mechanics? As of now it seems like you are forcefully dismounted each time you do it

zoktiik:
> can you also take a look at the mounted lance charge mechanics?

Not in-scope. This is really only touching spell code. Maybe I'll do it, but it would be an entirely separate thing, 

I say maybe because I don't really like triaging reports.

edit: Also nevermind I'm not doing it because people are making unrelated touches to that code lol.

maker-of-nightmares:
Absolute giga chad! I can't really see any issues with this now unless it has some one line in it that says "If Desert Town Explode server" lmao

zoktiik:
@RadiantFlash7 can this get a fullmerge so I can make an alternative variant to #2846 hooking into this code I wrote

PR #2796, opened 2026-09-19
Naledi complex no longer optional, Vizier divergence tweak

RadiantFlash7:
## About The Pull Request
Per the request/feedback of players,
Naledi complex is no longer optional for Naledi Adventurers.
Vizier Divergence no longer has a 2 seconds immobilize. Instead it has a 0.5 second immobilize, with a slowdown effect that lasts 3 seconds, in line with other spells.

## Why It's Good For The Game
Player feedback.
Vizier is very good- easily the best at it's role out of all the naledi roles. That said, divergence was designed around a one on one confrontation. It was not designed for lopsided fights and it was being used to gank people for a free 2 second immobilize. This should help, I think?
Naledi complex being forced is more soulful then being optional.

## Changelog

:cl:
balance: Naledi Complex is no longer optional
balance: Divergence no longer has a 2 second immobilize. It now has a .5 second immobilize and a brief slowdown, like other spells.
/:cl:


code/modules/spells/spell_types/naledi.dm
```diff
@@ -573,8 +576,17 @@
 
 	spawn_fragments(center)
 
+/datum/status_effect/debuff/divergence/proc/remove_slow()
+	slow_timer = null
+	if(owner)
+		owner.remove_movespeed_modifier(MOVESPEED_ID_TIMELINEDIVERGE, TRUE)
```
Xander3359:
```suggestion
	owner?.remove_movespeed_modifier(MOVESPEED_ID_TIMELINEDIVERGE, TRUE)
```

Xander3359 (changes requested):

PR #2797, opened 2026-09-19
Fix Xylix Abscond keeping you buckled after teleport

zoktiik:
## About The Pull Request
Found this issue while testing #2795.

The Abscond miracle is specifically designed to let you teleport anywhere while restrained, even if you're buckled to something. `gesture_required` is explicitly set to `FALSE`, meaning you can cast it while restrained. With the new changes in #2795, this means that the Xylixian will teleport out of the structure, moving away from it while the game still considers the person "buckled". **This breaks things.**

I personally think Abscond is stupid broken, but it is not really my place to nerf this spell. I just want to fix it so we don't get weird state conflicts like this.

## Testing Evidence
Testmerged on my server alongside #2795, it works.

https://github.com/user-attachments/assets/a94f5728-4f95-4485-a93b-c46b05753c11

## Why It's Good For The Game
Necessary since we're unfucking the conditions for casting spells

## Changelog

No CL since upstream blocks this from happening at all currently, Just merge alongside the other PR.


Xander3359:
Is this why the spell seems so shit?
Is it supposed to break you out of cuffs too or are you just fucked if they chain you.

zoktiik:
It does not break you out of any cuffs. It just teleports you. 

Since, on upstream ANY spell is uncastable while buckled to something that isn't a steed, **you don't see what happened on the video on the live server right now.** Clicking Abscond would fail because you're buckled to something. Even though it's probably the only "invoked" spell which isn't supposed to be affected by restraints at all.
I'm fixing a bug and then also fixing the unintended behavior that gets exposed *after the bugfix*.

Yes Abscond is basically useless right now, not even because that's intentional. It's just because of broken shit in spell code that affects all spells.

ophaq:
This is a good bugfix.

Also as someone who mains Xylixian let me tell you abscond isn't useless at all. It's basically a mage's blink that takes several seconds to charge up and you can leap up Z-levels with it like the leap spell. So it's great at getting to roofs and getting out of crowded doors / groups of people. It has a somewhat long cooldown though.

zoktiik:
Useless for getting unbuckled

ophaq:
> Useless for getting unbuckled

Yeah the point of it was not to be unbuckled. It's purely for mobility. Dunno why it had that unbuckle feature...

zoktiik:
Oh wait so it SHOULDN'T let you escape something you're buckled to
(it was letting me teleport off but kept buckled state)

Alright well I'll fix dat shit.
Gestural component will remain FALSE but I'll just check the buckled object

Is it supposed to help you escape while bound AND grabbed as well? @ophaq 
Or even just teleporting while bound?
Cuz if the second one's not true, that's stupid easy to toggle off with just making the gestural component not a requirement

ophaq:
> Oh wait so it SHOULDN'T let you escape something you're buckled to (it was letting me teleport off but kept buckled state)
> 
> Alright well I'll fix dat shit. Gestural component will remain FALSE but I'll just check the buckled object
> 
> Is it supposed to help you escape while bound AND grabbed as well? @ophaq Or even just teleporting while bound? Cuz if the second one's not true, that's stupid easy to toggle off with just making the gestural component not a requirement

It's supposed to let you escape. But if it counts as being buckled despite not being buckled and causing issues like you said then that should be fixed. It's like mage's blink, which lets you escape grabs or move past people. If you're already chained and it's somehow getting you out of that then that should be fixed. But I am pretty sure being chained prevents most spells from being used.

zoktiik:
You set `gesture_required = FALSE` which means it ignores all restraints when cast-checking. In this case it will ignore being caught by traps, or having your arms bound, or being placed on restraining furniture.

Am I setting it to `TRUE`?

zoktiik:
I'm just gonna make it `gesture_required = TRUE` if I don't get a response lol, cause the whole conversation points to it not being supposed to be usable while you're restrained at all

ophaq:
> I'm just gonna make it `gesture_required = TRUE` if I don't get a response lol, cause the whole conversation points to it not being supposed to be usable while you're restrained at all

Test it in game to see if that fixes it. I don't know off hand but if the spell works as intended but fixes what you want it to in testing then great.

zoktiik:
I am asking *you* what specifically *is* intended, and what specifically *is not* intended, because you made or ported the spell. You put it in, and with these fixes I am technically adding new behavior over live, because the spells fix PR makes buckling more nuanced again after it was broken for some time. Restraint structures are now specifically considered "a thing" with the main spells PR.

So with Abscond, are any of the following not intended:
- Teleporting while your arms are tied
- Teleporting while your arms are tied and you are in a grab
- Teleporting while you are put in a restraint structure
- Teleporting while your arms are tied and you are put in a restraint structure

If *none* of these are "intended" then I will make it "`gesture_required`".
If one or more of these are "intended" then we can code specific behavior to accommodate for that, i.e. I will implement that.

ophaq:
> I am asking _you_ what specifically _is_ intended, and what specifically _is not_ intended, because you made or ported the spell. You put it in, and with these fixes I am technically adding new behavior over live, because the spells fix PR makes buckling more nuanced again after it was broken for some time. Restraint structures are now specifically considered "a thing" with the main spells PR.
> 
> So with Abscond, are any of the following not intended:
> 
>     * Teleporting while your arms are tied
> 
>     * Teleporting while your arms are tied and you are in a grab
> 
>     * Teleporting while you are put in a restraint structure
> 
>     * Teleporting while your arms are tied and you are put in a restraint structure
> 
> 
> If _none_ of these are "intended" then I will make it "`gesture_required`". If one or more of these are "intended" then we can code specific behavior to accommodate for that, i.e. I will implement that.

Yeah none of that is intended. If you're bound up, you shouldn't be teleporting out. If you're just normally grabbed and not tied that is fine, as you can already do that with blink.

zoktiik:
Right, I'll keep this fix for teleporting from non-restraining objects.

The only other edge case I can think of that gesture_required doesn't cover is "armless cast", in other words being able to cast with disabled or missing arms that are otherwise unrestrained. I think that *would* perhaps be the only other case that we can't directly cover here, and tbh that can just be a variable for spells.

Should be all I need now, thank you!

